PR #2, opened 2016-05-20
First commit.

vitillo:
This PR has to be rebased before merging it but I figured the full log might be handy for reviewing purposes. @mreid-moz, could you take a look please?


README.md
```diff
@@ -1 +1,60 @@
...
+```
+
+and pushed to Docker hub with
+```bash
+docker push vitillo/telemetry-airflow
+```
+
+### Testing
+
+A single task, e.g. `spark`, of an Airflow dag, e.g. `example`, can be run with an execution date, e.g. `2015-06-02`, in the `test` environment with:
```
mreid-moz: Example date doesn't match the run command below.

ansible/deploy_aws.yml
```diff
@@ -0,0 +1,24 @@
...
+      s3: bucket={{airflow_bucket}} region={{region}} mode=create
+
+    - name: copy EMR bootstrap script
+      s3: bucket={{airflow_bucket}} region={{region}} object=steps/airflow.sh src=ansible/files/spark/airflow.sh mode=put
+
+    - name: create task definition
+      shell: "{{ item }}"
+      with_items:
+        - ecs-cli configure --cluster {{ ecs_cluster_name }}
+        - ecs-cli compose --project-name airflow-test --file ansible/files/docker-compose.yml up
```
mreid-moz: What does the `project-name` argument do? Can we use something more descriptive than `airflow-test`?

ansible/files/airflow/airflow.cfg
```diff
@@ -0,0 +1,195 @@
...
+
+# The folder where your airflow pipelines live, most likely a
+# subfolder in a code repository
+dags_folder = /usr/local/airflow/dags
+
+# The folder where airflow should store its log files. This location
+base_log_folder = /usr/local/airflow/logs
+# An S3 location can be provided for log backups
+# For S3, use the full URL to the base folder (starting with "s3://...")
+s3_log_folder = None
```
mreid-moz: This option has been deprecated, see the [latest example config](https://github.com/apache/incubator-airflow/blob/master/airflow/configuration.py#L174-L183)

ansible/files/airflow/airflow.cfg
```diff
@@ -0,0 +1,195 @@
...
+# Whether to load the examples that ship with Airflow. It's good to
+# get started, but you probably want to set this to False in a production
+# environment
+load_examples = False
+
+# Where your Airflow plugins are stored
+plugins_folder = /usr/local/airflow/plugins
+
+# Secret key to save connection passwords in the db
+fernet_key = $FERNET_KEY
```
mreid-moz: Can the config file read environment variables like this? Or will it just use the literal string `$FERNET_KEY`?

ansible/files/airflow/airflow.cfg
```diff
@@ -0,0 +1,195 @@
...
+base_url = http://localhost:8080
+
+# The ip specified when starting the web server
+web_server_host = 0.0.0.0
+
+# The port on which to run the web server
+web_server_port = 8080
+
+# Secret key used to run your flask app
+secret_key = temporary_key
```
mreid-moz: Do we need a real key here?

ansible/files/airflow/airflow.cfg
```diff
@@ -0,0 +1,195 @@
...
+# If you want airflow to send emails on retries, failure, and you want to
+# the airflow.utils.send_email function, you have to configure an smtp
+# server here
+smtp_host = localhost
+smtp_starttls = True
+smtp_ssl = False
+smtp_user = airflow
+smtp_port = 25
+smtp_password = airflow
+smtp_mail_from = airflow@airflow.local
```
mreid-moz: `telemetry-alerts@mozilla.com`

ansible/files/airflow/airflow.cfg
mreid-moz: Never mind, I see it's set below. Looks like it should be changed to `{FERNET_KEY}` to be properly replaced with `sed`

ansible/files/spark/airflow.sh
```diff
@@ -0,0 +1,112 @@
...
+
+if [ -z "$job_name" ] || [ -z "$user" ] || [ -z "$uri" ] || [ -z "$data_bucket" ]; then
+    error_msg "missing argument(s)"
+    exit -1
+fi
+
+s3_base="s3://$data_bucket/data/$user/$job_name"
+
+# Wait for Parquet datasets to be loaded
+while ps aux | grep hive_config.sh | grep -v grep > /dev/null; do sleep 1; done
```
mreid-moz: `while pgrep -f hive_config.sh  > /dev/null...`

ansible/deploy_aws.yml
vitillo: See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose.html.

ansible/files/spark/airflow.sh
```diff
@@ -0,0 +1,112 @@
...
+    esac
+    shift
+done
+
+if [ -z "$job_name" ] || [ -z "$user" ] || [ -z "$uri" ] || [ -z "$data_bucket" ]; then
+    error_msg "missing argument(s)"
+    exit -1
+fi
+
+s3_base="s3://$data_bucket/data/$user/$job_name"
```
mreid-moz: Including the owner name in the S3 uri will change the data path if a job's owner changes. Do we want this?

ansible/files/airflow/airflow.cfg
vitillo: I don't think so.

ansible/files/airflow/airflow.cfg
vitillo: Are you planning to configure smtp on that machine?

mreid-moz: I would rather uses SES if we can

ansible/files/spark/airflow.sh
vitillo: Yes, I don't want jobs with the same names to clash between users. I would also like to enforce ownership.

dags/operators/emr_spark_operator.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+    @apply_defaults
+    def __init__(self, job_name, owner, uri, instance_count, env={}, arguments="", *args, **kwargs):
+        super(EMRSparkOperator, self).__init__(*args, **kwargs)
+        self.job_name = job_name
+        self.owner = owner
+        self.uri = uri
+        self.arguments = arguments
+        self.environment = " ".join(["{}={}".format(k, v) for k, v in env.iteritems()])
```
mreid-moz: We should probably validate that the items in `env` are not going to cause problems (by containing spaces or quotes etc) and throw an error

dags/operators/emr_spark_operator.py
```diff
@@ -0,0 +1,138 @@
...
+
+        client = boto3.client('emr', region_name=EMRSparkOperator.region)
+        response = client.run_job_flow(
+            Name = self.job_name,
+            ReleaseLabel = EMRSparkOperator.release_label,
+            JobFlowRole = EMRSparkOperator.flow_role,
+            ServiceRole = EMRSparkOperator.service_role,
+            Applications = [{'Name': 'Spark'}, {'Name': 'Hive'}],
+            Configurations = requests.get("https://s3-{}.amazonaws.com/{}/configuration/configuration.json".format(EMRSparkOperator.region, EMRSparkOperator.spark_bucket)).json(),
+            LogUri = "s3://{}/logs/{}/{}/".format(EMRSparkOperator.airflow_bucket, self.owner, self.job_name),
```
mreid-moz: +1 for adding job logging.

ansible/deploy_aws.yml
mreid-moz: Let's change to something like `telemetry-airflow`?

ansible/files/spark/airflow.sh
mreid-moz: I'm OK with this. Currently the 'run a job' and 'publish data' tasks are one and the same. It would be a good idea to separate them and take a more deliberate approach to publishing data for use by dashboards etc.

dags/operators/emr_spark_operator.py
vitillo: The script runner does also its own set of substitutions (like " -> \")...

mreid-moz:
Looks good, r+

PR #3, opened 2016-06-19
Various improvements.

vitillo:


ansible/deploy_aws.yml
```diff
@@ -8,11 +8,38 @@
...
 
+    - ec2_group_facts:
+        region: "{{ region }}"
+        filters:
+          group-name: "{{ ecs_sg_name }}"
+      register: sg
+
+    - name: create load balancer
+      ec2_elb_lb:
+        name: telemetry-ecs
```
mreid-moz: I think we should make the elb name a var in the environment.

ansible/deploy_aws.yml
```diff
@@ -8,11 +8,38 @@
...
+          interval: 60 # seconds
+          unhealthy_threshold: 2
+          healthy_threshold: 2
+
     - name: create task definition
       shell: "{{ item }}"
       with_items:
         - ecs-cli configure --cluster {{ ecs_cluster_name }}
-        - ecs-cli compose --project-name telemetry-airflow --file ansible/files/docker-compose.yml up
+        - ecs-cli compose --project-name telemetry-airflow --file ansible/files/docker-compose.yml create
```
mreid-moz: nit: trailing space

ansible/library/ec2_group_facts.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/python
```
mreid-moz: I did not carefully review this file - I assume it's copied in from somewhere else?  If that's the case, can we add something to the commit message saying so (ideally with a link to where it came from)?

ansible/files/airflow/airflow.cfg
```diff
@@ -89,6 +89,10 @@ authenticate = False
 # Filter the list of dags by owner name (requires authentication to be enabled)
 filter_by_owner = False
 
+# Authentication
+authenticate = {AIRFLOW_ENABLE_AUTH}
+auth_backend = airflow.contrib.auth.backends.password_auth
```
mreid-moz: This will require a manual step of creating a user account. We should add that to the `README`.

mreid-moz: Eventually, we're going to move to something like LDAP, right?

vitillo: Yes.

mreid-moz:
Looks good, r+

PR #17, opened 2016-07-04
Increase timeout to 10 hours (temporarily).

mreid-moz:
The job is expected to take 3-4 hours, but was intermittently
timing out at 6 hours. Increase to 10 hours while profiling and
debugging the job.


vitillo:
r+

PR #18, opened 2016-07-04
Move data output to the telemetry-parquet bucket.

mreid-moz:
This affects all jobs via the location of the airflow.sh script, but
in particular, it also moves the output location for the
`main_summary.py` and `crash_aggregates.py` jobs.

New jobs should be tested by explicitly writing to the test bucket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ docker push mozdata/telemetry-airflow
 
 ### Testing
 
+When developing and testing a new DAG, it should be configured to write data to the `telemetry-test-bucket` bucket, and **not** the `airflow_bucket` specified in the ansible environment. Only when the new code and data have been validated should it begin writing to the `airflow_bucket`.
+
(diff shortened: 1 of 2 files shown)


maurodoglio:
@mreid-moz will the jobs on airflow start to spit data to the prod bucket when you deploy this?
can I change the output bucket from the airflow admin?

mreid-moz:
@maurodoglio That's right, both jobs will start writing to `telemetry-parquet` with this change.

I don't think the Airflow Web UI lets you change it, but you can easily override the output bucket by explicitly setting `env = {"date": "{{ ds_nodash }}", "bucket": "telemetry-test-bucket"},` for example.

I can make that change for now, and we can migrate `crash_aggregates.py` separately if you prefer.

maurodoglio:
No that's fine, I think the job has been on airflow for more than a week now. I'll delete the scheduled job from atmo when this lands. Thanks!

maurodoglio:
👍

vitillo:
r+

PR #19, opened 2016-07-07
Add job to generate longitudinal view.

vitillo:


dags/longitudinal.py
```diff
@@ -0,0 +1,24 @@
...
+    'retry_delay': timedelta(minutes=30),
+}
+
+dag = DAG('longitudinal', default_args=default_args, schedule_interval='@weekly')
+
+t0 = EMRSparkOperator(task_id="longitudinal",
+                      job_name="Longitudinal View",
+                      execution_timeout=timedelta(hours=10),
+                      instance_count=30,
+                      env = {"date": "{{ ds_nodash }}", "bucket": "{{ task.__class__.airflow_bucket }}"},
```
mreid-moz: Has this already been tested? Should we configure it to output in `telemetry-test-bucket` for it's trial run?

vitillo: I tested it locally.

mreid-moz: Ok, since it doesn't overwrite or append, it should be safe anyways. Looks good!

PR #20, opened 2016-07-08
Add client count view job.

vitillo:


jobs/client_count_view.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [[ -z $bucket || -z $date ]]; then
```
mreid-moz: Needs quote around the vars you're checking, like:
`if [[ -z "$bucket" || -z "$date" ]]; then`

Otherwise if they are both empty you will get a bash error `-bash: unexpected argument `||' to conditional unary operator`.

Please also update the check in `longitudinal` and `crash_aggregate` view jobs.

vitillo:
@mreid-moz r?

PR #29, opened 2016-07-22
Add documentation for re-running a task

mreid-moz:
r? @maurodoglio

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/29)


maurodoglio:
Sorry I totally forgot about this pull request. This is one of those cases where I feel like we should use bz review flags to assign reviews (instead of the ephemeral github notifications)

The change looks great, thanks for writing it down. 🚢 it!

PR #37, opened 2016-09-06
Bug 1275889 - Backfill main_summary v3

mreid-moz:
Set the start date for this dataset back to the beginning of time.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/37)


vitillo:
@mreid-moz Aren't 6 months back enough?

mreid-moz:
6 months should be enough, but since backfill appears to be manual anyways, I figured I might as well set it back to the time the first data is available. Then we won't need to change the code if more history is needed.

PR #48, opened 2016-09-29
Bug 1306225 - Decrease default max dag runs to 5

mreid-moz:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/48)

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -14,7 +14,7 @@ default_args = {
     'retry_delay': timedelta(minutes=30),
 }
 
-dag = DAG('main_summary', default_args=default_args, schedule_interval='@daily', max_active_runs=10)
+dag = DAG('main_summary', default_args=default_args, schedule_interval='@daily')
 
 # Make sure all the data for the given day has arrived before running.
 t0 = BashOperator(task_id="delayed_start",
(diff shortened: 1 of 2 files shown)


dags/addons.py
```diff
@@ -13,7 +13,7 @@
     'retry_delay': timedelta(minutes=30),
 }
 
-dag = DAG('addons', default_args=default_args, schedule_interval='@daily')
+dag = DAG('addons', default_args=default_args, schedule_interval='@daily', max_active_runs=5)
```
maurodoglio: The value of this parameter is the same as the default set in the file above, do we really need to enforce it here as well?

dags/main_summary.py
```diff
@@ -14,7 +14,7 @@
     'retry_delay': timedelta(minutes=30),
 }
 
-dag = DAG('main_summary', default_args=default_args, schedule_interval='@daily', max_active_runs=10)
+dag = DAG('main_summary', default_args=default_args, schedule_interval='@daily', max_active_runs=5)
```
maurodoglio: same as above

dags/addons.py
mreid-moz: Nope, don't need it in both places. Just being overly wordy. I'll trim it.

PR #52, opened 2016-10-04
Bug 1304662 - Increase main_summary delay to 1hr

mreid-moz:
Handle the increased timeout to flush data to S3, ensuring that all
data is available for the previous day when we process main pings.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/52)


vitillo:
r+

PR #55, opened 2016-10-13
Bug 1286249: Add timeouts to jobs without them

sunahsuh:
Not sure how I should be testing these, but the dags are kicking off fine locally so it looks like the timeouts are being parsed correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/55)

### DIFF
--- a/dags/bugzilla_dataset.py
+++ b/dags/bugzilla_dataset.py
@@ -29,6 +29,7 @@ env = {
 t0 = EMRSparkOperator(
     task_id="update_bugs",
     job_name="Bugzilla Dataset Update",
+    execution_timeout=timedelta(hours=5),
     instance_count=1,
     env=env,
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/bugzilla_dataset.sh",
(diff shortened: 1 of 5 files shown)


autolander:

vitillo:
@mreid-moz r?

mreid-moz:
How did you decide upon the timeout values?

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/55#-:-KU2XLjbelAFMJoJjiAT:b2mk7wq)*

sunahsuh:
I looked at the last 100 runs for the jobs and eyeballed the average runtime -- if that was under ~2hrs, i set the timeout for 4hrs (which seemed like the precedent with other jobs with timeouts). The Telemetry aggregator job is the only one that takes a bit more time, and the times have a rather high variance -- there's 4 runs that broke 7 hours, but otherwise they're between 5-7 hours, so I set that timeout to 10 hours. I have a feeling if we have a daily job that breaks 10 hours that might be something we might be concerned about.

mreid-moz:
:lgtm: Great, that seems like the right approach.

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/55#-:-KU2ng852JXwoLfLYiss:bbxab4o)*

PR #56, opened 2016-10-13
Bug 1290146 - Change date param to end of the week in longitudinal tasks

sunahsuh:
This hasn't been tested on an actual run, but the rendered output on each of the tasks looks correct:

Note that this means if we trigger a manual run of the DAG, the date passed will be current_date + 7 days since there's no way to override execution date in that command.

This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/56)

### DIFF
--- a/dags/longitudinal.py
+++ b/dags/longitudinal.py
@@ -1,7 +1,7 @@
 from airflow import DAG
 from datetime import datetime, timedelta
 from operators.emr_spark_operator import EMRSparkOperator
-from utils.constants import DS_PLUS_7_NO_DASH
+from utils.constants import DS_WEEKLY
 
 default_args = {
     'owner': 'rvitillo@mozilla.com',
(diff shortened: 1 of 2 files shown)


autolander:

vitillo:
> Note that this means if we trigger a manual run of the DAG, the date passed will be `current_date + 7 days` since there's no way to override execution date in that command.

That is not going to work. For example one might trigger a dag today and generate a longitudinal dataset with a date in the future which does not contain any signal for future days. We have jobs that make certain assumptions based on the dataset generation date.

vitillo:
@mreid-moz r?

mreid-moz:
Can we modify the code so the job uses `min(yesterday, current_date + 7 days)`? That way we'd ensure that the generated dataset doesn't contain any days that are not yet finished.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/56#-:-KU2o_98EEWJOpGzkWlz:b7o3cqb)*

sunahsuh:
I went with Roberto's suggestion re: switching to the normal date if the DAG is triggered manually

sunahsuh:
I was able to test it locally by manually triggering the dag and looking at the rendered params in the web UI -- scheduled runs use execution date + 7, manual runs use current date

mreid-moz:
Hmm, now the variable name for the new macro doesn't really reflect what it is... maybe "DS_END_OF_WEEK"?

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/56#-:-KU3OKZbx_fo4uj5xLxV:b-18siv9)*

mreid-moz:
:lgtm:

---

Reviewed 1 of 3 files at r1, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/56#-:-KU3TzZZ9lwZkqIAAaSb:bnfp4nl)*

PR #57, opened 2016-10-17
Bug 1310486 - Default dags_are_paused_at_creation to True in airflow config

sunahsuh:
This fixes the issue of dev deployments accidentally launching clusters for elapsed jobs -- all new DAGs will now start off by default and must be explicitly turned on when ready.

screen shot 2016-10-17 at 5 17 31 pm

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/57)


vitillo:
Nice!

PR #60, opened 2016-10-18
Bug 1309352 - Update all local paths in ansible playbooks for ansible 2.1.2.0

sunahsuh:
Local paths in playbooks are now relative to the playbook location, not where the playbook is run from in ansible 2.1.2.0

After we merge these changes, everyone will need to update their ansible versions locally.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/60)

### DIFF
--- a/ansible/deploy_aws.yml
+++ b/ansible/deploy_aws.yml
@@ -9,7 +9,7 @@
       s3: bucket={{airflow_bucket}} region={{region}} mode=create
 
     - name: copy EMR bootstrap script
-      s3: bucket={{airflow_bucket}} region={{region}} object=steps/airflow.sh src=ansible/files/spark/airflow.sh mode=put
+      s3: bucket={{airflow_bucket}} region={{region}} object=steps/airflow.sh src={{ playbook_dir }}/files/spark/airflow.sh mode=put
 
     - name: create load balancer
       ec2_elb_lb:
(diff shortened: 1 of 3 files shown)


vitillo:
@sunahsuh Does the README need to be updated as well?

vitorbaptista:
I had a similar problem running this repository on my local machine. However, I solved it by using ansible's `{{ playbook_dir }}` variable, like:

```yml
compose_conf:
  - "{{ playbook_dir }}/files/docker-compose.yml"
  - "{{ playbook_dir }}/files/docker-compose-local.yml"
```

Not sure if that solves the issue @sunahsuh was having but, if so, it feels like a more stable way of fixing this.

sunahsuh:
@vitillo no, it works the same from the end user's perspective
@vitorbaptista thanks for the tip! That'd let us keep backwards compatibility with older versions =)

sunahsuh:
Switched to the playbook_dir variable -- works fine for local deployment, but it's too close to the job run times for me to feel comfortable doing a prod deployment

mreid-moz:
Looks good! With the change to `playbook_dir`, devs no longer need to upgrade ansible per the original description.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/telemetry-airflow/60#-:-KUc0f0LCQ4v_grMLmn-:b-z0g708)*

PR #66, opened 2016-10-31
Bug 1305423 - Add mobile jobs

maurodoglio:
These jobs have been running for months on atmo v1, it's time to move
them to airflow.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/66)

### DIFF
--- a/dags/android_addons.py
+++ b/dags/android_addons.py
@@ -1,0 +1,27 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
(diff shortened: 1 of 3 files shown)


maurodoglio:
@sunahsuh could you please r?

sunahsuh:
Looks good to me -- my only suggestion would be to consider grouping these tasks under a single DAG without any dependencies between them, since they seem to be a logical grouping.

maurodoglio:
That's a good idea, but I prefer to leave it eventually for a separate PR because there's a bit of pressure to deploy this sooner rather than later. Thanks for the review 😃

PR #68, opened 2016-11-07
Reduce the number of instances for Addons

mreid-moz:
Reading from main_summary is a lot more efficient than reading the
raw data. 3 instances should be plenty.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/68)


autolander:

PR #71, opened 2016-11-11
Increase instances for distribution viewer script

robhudson:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/71)

### DIFF
--- a/dags/longitudinal.py
+++ b/dags/longitudinal.py
@@ -77,7 +77,7 @@ t5 = EMRSparkOperator(task_id="distribution_viewer",
                       email=["telemetry-alerts@mozilla.com", "chudson@mozilla.com"],
                       execution_timeout=timedelta(hours=10),
                       release_label="emr-5.0.0",
-                      instance_count=5,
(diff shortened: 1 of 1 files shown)


autolander:

dags/longitudinal.py
```diff
@@ -77,7 +77,7 @@
                       email=["telemetry-alerts@mozilla.com", "chudson@mozilla.com"],
                       execution_timeout=timedelta(hours=10),
                       release_label="emr-5.0.0",
-                      instance_count=5,
+                      instance_count=30,
```
vitillo: I thought this wasn't needed any longer after stopping the SparkContext. Did I misunderstand?

vitillo: Using many instances might cause to hit some AWS limits when multiple jobs are scheduled at the same time; let's try to not do that unless there is a good reason for it.

robhudson: Yeah, this seems to be running fine as-is. 👍

PR #76, opened 2016-12-15
Bug 1309622 - Prepare for dockerflow instead of Ansible+ECS

jezdez:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/76)


jezdez:
@sunahsuh @vitillo This is basically just a POC in case we go with dockerflow (which I strongly suspect given the recent discussion around using a PaaS instead). Let's wait for a discussion with @jasonthomas proper before merging any of this.

Dockerfile
```diff
@@ -1,23 +1,67 @@
...
+ENV AWS_REGION us-west-2
+# ENV AWS_ACCESS_KEY_ID
+# ENV AWS_SECRET_ACCESS_KEY
+
+ENV SPARK_BUCKET telemetry-spark-emr-2
+ENV AIRFLOW_BUCKET telemetry-test-bucket
+ENV PRIVATE_OUTPUT_BUCKET telemetry-test-bucket
+ENV PUBLIC_OUTPUT_BUCKET telemetry-test-bucket
+
+ENV EMR_KEY_NAME mozilla_vitillo
```
mreid-moz: Can we use the `20161025-dataops-dev` key here?

Dockerfile
```diff
@@ -1,23 +1,67 @@
...
+ENV EMR_FLOW_ROLE telemetry-spark-cloudformation-TelemetrySparkInstanceProfile-1SATUBVEXG7E3
+ENV EMR_SERVICE_ROLE EMR_DefaultRole
+ENV EMR_INSTANCE_TYPE c3.4xlarge
+
+ENV AIRFLOW_HOME /app
+ENV AIRFLOW_AUTHENTICATE False
+ENV AIRFLOW_BROKER_URL redis://redis:6379/0
```
sunahsuh: Inquiring minds want to know -- what made you decide to switch to redis?

jezdez: Ah, simple, Redis is operationally a lot less complex locally and both AWS (ElastiCache) and Heroku offer service for it for production.

Celery supports Redis as a 1st level broker and Airflow doesn't use any of its more complex features that would require a more advanced broker such as RabbitMQ. In fact, I believe the "CeleryExecutor" is pretty limited in how it can be configured, so it's running almost with the default Celery configuration.

In short: for simplicity's sake

Dockerfile
jezdez: Sure! Mind explaining the difference? In atmo we use that key as well.

vitillo: See Bug [1321071](https://bugzilla.mozilla.org/show_bug.cgi?id=1321071). By using a shared key I am not the only one that can access a cluster if something goes wrong.

jezdez: Ah, excellent, thanks for the pointer!

jezdez:
@jasonthomas Any chance for you to take a look at this as well? This is related to the things I'm doing for ATMO so will have to update the Dockerfile/docker-compose.yml files with what I'm learning there.

jezdez:
@jasonthomas FTR I've updated this docker setup with the structure that we agreed on for atmo in https://github.com/mozilla/telemetry-analysis-service/pull/86

PR #99, opened 2017-03-24
Bug 1350380 - Use ds_nodash instead of DS_WEEKLY for churn

acmiyaguchi:
This uses `ds_nodash` to set the execution date to the date that the job is scheduled on.


sunahsuh:
> The code should run on Wednesdays, and pass in the actual calendar date on which it runs.
ds_nodash will pass in the execution date, which isn't the date that the task actually runs but the *start* of the period (e.g. last Wednesday's date)

sunahsuh:
...Unless you're using cron syntax like you are here 🕐

PR #102, opened 2017-04-11
Bug 1355153 - Run parquet2hive after longitudinal generation

whd:
Requires https://github.com/mozilla/emr-bootstrap-spark/pull/79.

I verified that steps run via EMRSparkOperator inherit the environment variables added in the bootstrap script.

### DIFF
--- a/jobs/longitudinal_view.sh
+++ b/jobs/longitudinal_view.sh
@@ -5,6 +5,8 @@ if [[ -z "$bucket" || -z "$date" ]]; then
     exit 1
 fi
 
+set -e
+
 git clone https://github.com/mozilla/telemetry-batch-view.git
 cd telemetry-batch-view
 sbt assembly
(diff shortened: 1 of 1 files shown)


jobs/longitudinal_view.sh
```diff
@@ -15,3 +15,5 @@ spark-submit --executor-cores 8 \
              target/scala-2.11/telemetry-batch-view-1.1.jar \
              --to $date \
              --bucket $bucket
+
+parquet2hive -ulv 1 --sql s3://telemetry-parquet/longitudinal | beeline -u jdbc:hive2://${HIVE_SERVER}:10000
```
vitillo: This should run only if the Spark job completes succesfully. Otherwise what could happen is that the job fails,  this command runs succesfully and finally Airflow reports the whole thing as succesful.

whd:
I copied some return code checking from elsewhere in this repo.

jobs/longitudinal_view.sh
```diff
@@ -15,3 +15,9 @@ spark-submit --executor-cores 8 \
              target/scala-2.11/telemetry-batch-view-1.1.jar \
              --to $date \
              --bucket $bucket
+
+rc=$?
```
vitillo: What about using `set -e`?

vitillo:
@fbertsch good point, I think @acmiyaguchi's p2h operator will fix that

vitillo:
LGTM, thanks!

PR #104, opened 2017-04-24
Bug 1357877 - [part 1] Add ToplineSummaryView

acmiyaguchi:
This commit adds monthly and weekly topline summary generation to airflow. The current reporting uses periods compatible with airflow's `@weekly`and `@monthly` shortcuts.

### DIFF
--- a/dags/topline.py
+++ b/dags/topline.py
@@ -1,0 +1,40 @@
+from airflow import DAG
+from datetime import timedelta, datetime
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'amiyaguchi@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 3, 26),
+    'email': ['telemetry-alerts@mozilla.com', 'amiyaguchi@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 2,
+    'retry_delay': timedelta(minutes=30),
+}
+
+
+def topline_dag(dag, mode, instance_count):
+    t0 = EMRSparkOperator(
+        task_id="topline_summary",
+        job_name="Topline Summary View",
+        execution_timeout=timedelta(hours=8),
+        instance_count=instance_count,
+        env={
+            "date": "{{ ds_nodash }}",
+            "bucket": "{{ task.__class__.private_output_bucket }}",
+            "mode": mode
+        },
+        uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/topline_summary_view.sh",
+        dag=dag)
+
+
+dag_weekly = DAG('topline_weekly',
+                 default_args=default_args,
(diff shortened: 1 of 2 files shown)


dags/topline.py
```diff
@@ -0,0 +1,46 @@
...
+        job_name="Topline Summary View",
+        execution_timeout=timedelta(hours=8),
+        instance_count=instance_count,
+        env={
+            "date": "{{ ds_nodash }}",
+            "bucket": "{{ task.__class__.private_output_bucket }}",
+            "mode": mode
+        },
+        uri=(
+            "https://raw.githubusercontent.com/"
```
mreid-moz: I think this URL should go all on a single line, even if it ends up going over the line-length limit.

dags/topline.py
```diff
@@ -0,0 +1,46 @@
...
+            "mozilla/telemetry-airflow/master/"
+            "jobs/topline_summary_view.sh"
+        ),
+        dag=dag)
+
+    # TODO: Topline Dashboard Data
```
mreid-moz: Please remove this `TODO` - if needed, you can file a follow-up issue.

dags/topline.py
```diff
@@ -0,0 +1,46 @@
+from airflow import DAG
+from datetime import timedelta, datetime
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'amiyaguchi@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 4, 23),
```
mreid-moz: Should we set an earlier `start_date` to trigger some backfill?

dags/topline.py
```diff
@@ -0,0 +1,46 @@
...
+            "date": "{{ ds_nodash }}",
+            "bucket": "{{ task.__class__.private_output_bucket }}",
+            "mode": mode
+        },
+        uri=(
+            "https://raw.githubusercontent.com/"
+            "mozilla/telemetry-airflow/master/"
+            "jobs/topline_summary_view.sh"
+        ),
+        dag=dag)
```
mreid-moz: Nit: Consider returning the resulting task from this function. I know it's not used in the code below, but if later we want to add dependent tasks (ie. the Dashboard output tasks), these will need to be used as upstream dependencies.

acmiyaguchi: The next commit will handle the dependencies. That job also needs to be repeated per mode, so there's no use returning the values.

dags/topline.py
acmiyaguchi: Fair enough, this does make it hard to copy and paste into the url bar.

dags/topline.py
acmiyaguchi: I set it to 2017-03-26, which is the last Sunday in March. This should give a single monthly run and a few weekly runs.

dags/topline.py
acmiyaguchi: Done

PR #106, opened 2017-04-25
Add job for pulse testpilot ETL

harterrt:


dags/txp_pulse.py
```diff
@@ -0,0 +1,25 @@
...
+    'retries': 2,
+    'retry_delay': timedelta(minutes=30),
+}
+
+dag = DAG('txp_pulse', default_args=default_args, schedule_interval='@daily')
+
+t0 = EMRSparkOperator(task_id="txp_pulse",
+                      job_name="Pulse Testpilot Experiment ETL",
+                      execution_timeout=timedelta(hours=2),
+                      release_label="emr-5.0.0",
```
vitillo: Please use the default EMR release unless there is a strong reason to hardcode it.

jobs/txp_pulse.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# We use jupyter by default, but here we want to use python
+unset PYSPARK_DRIVER_PYTHON
+
+# Clone, install, and run
+git clone https://github.com/mozilla/python_etl.git
+cd python_etl
+pip install .
+python setup.py bdist_egg
+spark-submit scheduling/pulse.py
```
vitillo: Why are the following commands needed?
```
+pip install .
+python setup.py bdist_egg 
```

dags/txp_pulse.py
harterrt: Fixed

jobs/txp_pulse.sh
harterrt: These lines install the package locally and package the code so it can be [sent to the worker nodes](https://github.com/mozilla/python_mozetl/blob/master/scheduling/pulse.py#L12). We need to do this here until we get this library on pypi and installed during bootstrap. I'm working on getting this repo up on pypi. Jannis provided some [fixes yesterday](https://github.com/mozilla/python_mozetl/pull/12) that put us much closer.

All the same, I'd like to get this running now. The pulse etl job was scheduled by ATMO but is now failing due to [ATMO Issue 326](https://github.com/mozilla/telemetry-analysis-service/issues/326). I've been running the job manually which is interrupting their analysis. It will also be good to identify any issues before scheduling the remaining jobs.

vitillo: Right, but I think the first line might be superflous.

PR #110, opened 2017-04-28
Bug 1309622 - Prepare for dockerflow instead of Ansible+ECS (take 2)

whd:
Supercedes #76. The commits from jezdez + whd have in theory already been reviewed and approved per #76, but it might be good to check them again because I had to resolve a few merge conflicts.

See https://github.com/mozilla-services/cloudops-deployment/pull/649 for the companion cloudops-deployment PR.

Note: the gpg autosigning I have configured is somewhat dubious in value, esp. with the @jezdez commits that I had to resolve rebase conflicts in. I can squash this or unsign the rebased commits if that's preferred.


Dockerfile
```diff
@@ -1,23 +1,71 @@
...
+
+ENV AIRFLOW_HOME=/app \
+    AIRFLOW_AUTHENTICATE=True \
+    AIRFLOW_AUTH_BACKEND=airflow.contrib.auth.backends.google_auth \
+    AIRFLOW_BROKER_URL=redis://redis:6379/0 \
+    AIRFLOW_RESULT_URL=${AIRFLOW_BROKER_URL} \
+    AIRFLOW_FLOWER_PORT="5555" \
+    AIRFLOW_DATABASE_URL=postgres://postgres@db/postgres \
+    AIRFLOW_FERNET_KEY="VDRN7HAYDw36BTFbLPibEgJ7q2Dzn-dVGwtbu8iKwUg" \
+    AIRFLOW_SECRET_KEY="3Pxs_qg7J6TvRuAIIpu4E2EK_8sHlOYsxZbB-o82mcg" \
```
jgmize: Is this ok to be in a public repo?

whd: Yes.

whd: @jezdez it's not obvious that this isn't an actual secret. Should we change it to something more obviously fake?

Dockerfile
```diff
@@ -1,23 +1,71 @@
...
+
+ENV PYTHONUNBUFFERED=1 \
+    AWS_REGION=us-west-2 \
+    # AWS_ACCESS_KEY_ID= \
+    # AWS_SECRET_ACCESS_KEY= \
+    SPARK_BUCKET=telemetry-spark-emr-2 \
+    AIRFLOW_BUCKET=telemetry-test-bucket \
+    PRIVATE_OUTPUT_BUCKET=telemetry-test-bucket \
+    PUBLIC_OUTPUT_BUCKET=telemetry-test-bucket \
+    EMR_KEY_NAME=mozilla_vitillo \
```
mreid-moz: Can we use the shared ops key here?

whd: We actually do in the [production config version](https://github.com/mozilla-services/cloudops-deployment/pull/649/files#diff-26b01155c1789e2c29e617954cee976eR33). I can update it here as well.

jasonthomas: I think we should just update the values here to make them more as placeholder values then actual real values. This is overridden https://github.com/mozilla-services/cloudops-deployment/pull/649/files#diff-26b01155c1789e2c29e617954cee976eR33 here.

whd: Per discussion we're going to leave the README describing the production environment variables there, but take them out of the Dockerfile. The description of the env variables that need to be set will not necessarily be maintained, but there will be a link to the actual configs being used once the PR in cloudops-deployment is merged.

Dockerfile
jezdez: @whd Yep!

jezdez:
Ugh, that comment was about the [mozlog commit](https://github.com/mozilla/telemetry-airflow/commit/1f1ff087e262dfe0b71f3e35916bcdd0f2156448).

jezdez:
@whd Nicely done, thanks for carrying this over the finish line. 🎆

whd:
As it stands the mozlog formatted logs won't be sent anywhere anyway because our Logging 2.0 infrastructure needs to be updated to support the level of parameterization that the "data" application requires. I look forward to removing this patch.

PR #112, opened 2017-05-09
Add tab spinner severity job

fbertsch:
Tested this, ran successfully in ~5.5 hours on a 12 node cluster.

### DIFF
--- a/dags/spinner_severity_generator.py
+++ b/dags/spinner_severity_generator.py
@@ -1,0 +1,26 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'frank@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 5, 10),
+    'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 3,
+    'retry_delay': timedelta(minutes=30),
+}
+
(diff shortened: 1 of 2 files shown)


dags/spinner_severity_generator.py
```diff
@@ -0,0 +1,26 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'frank@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 4, 25),
```
vitillo: Is the backfill needed?

fbertsch:
@vitillo it does not need backfill, updated.

jobs/run-severity-generator.sh
```diff
@@ -0,0 +1,19 @@
...
+pip install .
+python setup.py bdist_egg
+
+rm -rf output
+mkdir output
+
+spark-submit scheduling/tab_spinner.py
+
+# upload files to s3
+aws s3 sync ./output s3://telemetry-public-analysis-2/spinner-severity-generator/data/
```
vitillo: Why are files manually uploaded instead of doing this in the job itself?

fbertsch: I don't have a good reason for this. It just happened this way as I moved the ipynb over.

vitillo: OK, it would be great if you could move that logic into the job itself.

fbertsch:
@vitillo moved file copying to the job. RFAL

PR #114, opened 2017-05-11
Bug 1329844 - Add topline dashboard generator

acmiyaguchi:
[Bug 1357877 - Add `ToplineSummary` and `topline_dashboard` to Airflow](https://bugzilla.mozilla.org/show_bug.cgi?id=1357877)

This adds the mozetl topline_dashboard script as a dependency of the topline summary. The PR for fixing toLocalIterator issues with timeouts should be addressed before this.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/telemetry-airflow/114)

### DIFF
--- a/jobs/topline_dashboard.sh
+++ b/jobs/topline_dashboard.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -x
+
+if [[ -z "$mode" ]]; then
+   echo "Missing arguments!" 1>&2
+   exit 1
+fi
+
+bucket="net-mozaws-prod-metrics-data"
+prefix="firefox-dashboard"
+
+git clone https://github.com/mozilla/python_mozetl.git
+cd python_mozetl
+python setup.py bdist_egg
+
+# Generate the driver script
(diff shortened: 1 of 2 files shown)


jobs/topline_dashboard.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -x
+
+if [[ -z "$mode" ]]; then
+   echo "Missing arguments!" 1>&2
+   exit 1
+fi
+
+bucket="net-mozaws-prod-metrics-data"
```
mreid-moz: I think files uploaded to this bucket need to have some special headers applied in order to propagate to the dashboard as expected. See [this line](https://github.com/mozilla-services/data-pipeline/blob/master/reports/executive_summary/run.sh#L124) for example.

We may need to explicitly copy the CSV file to the right location as part of the job to ensure it works.

acmiyaguchi: Yep, this should be handled in the corresponding code.

https://github.com/mozilla/python_mozetl/blob/master/mozetl/topline/topline_dashboard.py#L140

PR #119, opened 2017-05-25
Run parquet2hive after Main Summary

robotblake:
This adds a generic parquet2hive job and a call to it for the main_summary dataset.

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -94,6 +94,17 @@ t8 = EMRSparkOperator(task_id="txp_mau_dau",
                       uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/mozetl/testpilot/txp_mau_dau.py",
                       dag=dag)
 
+t9 = EMRSparkOperator(task_id="p2h_main_summary",
+                      job_name="P2H for Main Summary",
+                      execution_timeout=timedelta(hours=4),
+                      owner="bimsland@mozilla.com",
+                      email=["telemetry-alerts@mozilla.com", "bimsland@mozilla.com"],
+                      instance_count=1,
+                      env={"bucket": "{{ task.__class__.private_output_bucket }}"
+                           "prefix": "main_summary"},
+                      uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/parquet2hive.sh",
+                      dag=dag)
+
 t2.set_upstream(t1)
 t3.set_upstream(t1)
 t4.set_upstream(t1)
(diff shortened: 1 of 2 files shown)


whd:
I think we should be setting your p2h job to be the upstream for the jobs that have main_summary as their current upstream (assuming they use the centralized hive metastore to access main_summary). I'm guessing we already have the problem with main summary that [we had with longitudinal](https://bugzilla.mozilla.org/show_bug.cgi?id=1355153). This dag structure appears to make explicit that problem, and if the table is unavailable while the msck repair is running, running it in parallel with existing downstream jobs could cause more failures. In the fix for longitudinal we just tacked the p2h call at the end of the longitudinal run script (https://github.com/mozilla/telemetry-airflow/pull/102), with the understanding that eventually we will use a proper p2h operator as described [here]( https://bugzilla.mozilla.org/show_bug.cgi?id=1356703).

fbertsch:
I don't know if we should really be setting p2h upstream from the other jobs. They are all reading from the actual s3 files right now, and adding an extra 5+ hours delay to when they run may be detrimental. This should *eventually* be done, but not until we make metastore indexing much faster.

whd:
@fbertsch his is why I predicated my comment with "assuming they use the centralized hive metastore to access main_summary", and did not explicitly r- this. @robotblake explained to me that downstream jobs do not use hive, and was going to comment to that effect, but it looks like you beat him to it.

dags/main_summary.py
```diff
@@ -94,10 +94,22 @@
                       uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/mozetl/testpilot/txp_mau_dau.py",
                       dag=dag)
 
+t9 = EMRSparkOperator(task_id="p2h_main_summary",
+                      job_name="P2H for Main Summary",
+                      execution_timeout=timedelta(hours=4),
```
whd: Will this timeout be sufficient? @fbertsch has just mentioned 5+ hours as a possibility.

fbertsch:
@whd I misunderstood (or misread). Thanks for clarifying.

Another point: We are going to be doing a big backfill next week (6 months of data). We don't want to call p2h after every single run. How should we handle this? One option is only run p2h if the $batchdate is $YESTERDAY. All historical runs will not reindex the metastore.

dags/main_summary.py
robotblake: Ugh, yeah, should probably put it longer.

robotblake:
@fbertsch Good point, I'll take a look at adding that.

mreid-moz:
> We don't want to call p2h after every single run. How should we handle this?

Is it possible to check if hive already contains a given partition?

fbertsch:
@robotblake, can we close this or fix it?

PR #122, opened 2017-05-30
Add main_summary backfill job

fbertsch:


jobs/main_summary_backfill.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+if [[ -z "$bucket" || -z "$date" ]]; then
+   echo "Missing arguments!" 1>&2
+   exit 1
+fi
+
+todate=$(($date+6))
```
mreid-moz: I think `todate` should just be using the passed-in environment var (and ideally checking for non-empty with `bucket` and `date` above). This will give you weird values if you pass in something like `20170430` for date.

PR #124, opened 2017-06-02
Add nodes and increase timeout for main_summary_backfill

fbertsch:


dags/main_summary_backfill.py
```diff
@@ -18,8 +18,8 @@
 
 t0 = EMRSparkOperator(task_id="main_summary_backfill",
                       job_name="Main Summary Backfill",
-                      execution_timeout=timedelta(hours=10),
-                      instance_count=20,
+                      execution_timeout=timedelta(hours=24),
```
sunahsuh: Is this enough for 7 days? Seems like we may want to up this to 48 (or even 60?) until we know what the actual runtime is

fbertsch: I ran one day and it took a bit under 3 hours. I could bump it up higher for safety. Consider that the current main_summary only uses 10 nodes

fbertsch:
@sunahsuh went ahead and bumped to 48 hours.

PR #130, opened 2017-06-09
Add experiments aggregates view

fbertsch:


fbertsch:
Tested this on data from 20170610, seems to work as expected. Currently this is way too many nodes, but will make it largely future proof. If we want to downsize we can.

PR #135, opened 2017-06-20
Add experiments viewer import

robhudson:


dags/main_summary.py
```diff
@@ -105,15 +105,24 @@
...
+                       uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/experiment_aggregates_view.sh",
+                       dag=dag)
+
+t11 = EMRSparkOperator(task_id="experiments_aggregates_import",
+                       job_name="Experiments Aggregates Import",
+                       execution_timeout=timedelta(hours=2),
+                       instance_count=1,
+                       owner="chudson@mozilla.com",
+                       email=["telemetry-alerts@mozilla.com", "chudson@mozilla.com"],
+                       env={"date": "{{ ds_nodash }}"},
```
fbertsch: We should input the bucket here as well. If we change the output bucket, the `experiments_aggregates` task will write to a new location, which this job should read from.

robhudson: Updated to also include the bucket. I'll also update my import job to use it.

PR #137, opened 2017-06-27
Rename the search_rollups script

mreid-moz:
Update the name of the search script based on changes in
https://github.com/mozilla/python_mozetl/pull/52

Note that we need to co-ordinate the deploy of this change with merging the above PR.


mreid-moz:
@acmiyaguchi RFAL

PR #139, opened 2017-06-28
Bug 1370522 Migrate Engagement Ratio ETL to python_mozetl

SamPenrose:
The mozetl code has been tested locally but testing this diff locally has been intractable so far.


SamPenrose:
I executed engagement_ratio.sh on a 10-node cluster after backing up engagement_ratio.csv and it reproduced perfectly. Took 2 hours.

PR #149, opened 2017-07-26
Add hardware report job

comloo:


dags/longitudinal.py
```diff
@@ -40,11 +40,11 @@
                       job_name="Firefox Hardware Report",
                       execution_timeout=timedelta(hours=10),
```
fbertsch: Can you confirm that this timeout is still appropriate?

jobs/hardware_report.sh
```diff
@@ -0,0 +1,22 @@
...
+    exit 1
+fi
+
+# Create the package for distribution across the cluster
+git clone https://github.com/mozilla/python_mozetl.git
+cd python_mozetl
+python setup.py bdist_egg
+
+# Generate the driver script
+echo 'from mozetl.hardware_report import hardware_dashboard; hardware_dashboard.main()' > run.py
```
fbertsch: This is different than the way we run the rest of the jobs. Is this the only way to run the hardware report? Doesn't this mean the input variables are ignored? (`bucket` and `date`).

dags/longitudinal.py
comloo: The job takes about 3-6 hours to run, so I think 10 hour timeout is appropriate.

jobs/hardware_report.sh
whd: This appears to be using a new(ish) idiom for python ETL jobs that is also used by churn, engagement ratio, search, and the topline reports.  That said, we're indeed missing an `env` section in the operator definition to propagate the environment variables as arguments.

dags/longitudinal.py
```diff
@@ -40,11 +40,11 @@
...
                       instance_count=15,
```
whd: Add a line like `env={"date": DS_WEEKLY, "bucket": "{{ task.__class__.public_output_bucket }}"},` here (see below).

PR #150, opened 2017-08-01
Install mozetl before running hardware report

harterrt:
This ensures the dependencies are also installed.

I'll be able to test this once https://github.com/mozilla/python_mozetl/pull/85 is submitted.


harterrt:
> It is also my understanding that we're going to replace this idiom with a single top-level entry point anyway (mozilla/python_mozetl#81).

This is correct. I don't think it's worth modifying other jobs until we have the unified endpoint.

PR #151, opened 2017-08-01
Bug 1386306 - Add staging environment for experiments jobs

robhudson:


sunahsuh:
Hey Rob, I was thinking we'd create a brand new dag so we can kick it off as needed, as opposed to adding this to main summary dag. Additionally, I think we can reuse the same scripts for the jobs if we add a `branch` environment variable and add a line to check out that branch

robhudson:
Updated to create a new DAG and pass a branch env variable.

dags/experiments_staging.py
```diff
@@ -0,0 +1,91 @@
...
+        'telemetry-alerts@mozilla.com',
+        'chudson@mozilla.com'
+    ],
+    env={
+        'date': '{{ ds_nodash }}',
+        'db': 'experiments-viewer-staging-db',
+        'bucket': '{{ task.__class__.private_output_bucket }}_staging'
+    },
+    uri=('https://raw.githubusercontent.com/mozilla/experiments-viewer'
+         '/master/notebooks/import.py'),
```
sunahsuh: I suppose if you were interested in also using the staging DAG for your import job you could just change the branch here from master

jobs/experiment_main_summary_view.sh
```diff
@@ -1,17 +1,24 @@
...
-             --from $date \
-             --to $date
+git checkout $branch
+if [ $? -eq 0 ]; then
+   sbt assembly
+   spark-submit --master yarn \
+                --deploy-mode client \
+                --class com.mozilla.telemetry.views.ExperimentSummaryView \
+                target/scala-2.11/telemetry-batch-view-1.1.jar \
+                --bucket $bucket \
```
sunahsuh: You'd want to add the `inbucket` parameter here

(Here's where that option is defined: https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/ExperimentSummaryView.scala#L32)

jobs/focus_event_longitudinal_view.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
sunahsuh: Is this a linter thing? I'm wary of having unrelated changes in this PR (same with the perms changes below and the weird double shebang.) Could you submit these in a separate PR?

robhudson: 👍 I just happened to notice those things.

dags/experiments_staging.py
robhudson: the `env['db']` is my output and the `env['bucket']` is my input. But true... also changing where the script is located would allow me to test changes on a staging branch before they land on master and get run in production.

dags/operators/emr_spark_operator.py
```diff
@@ -41,6 +41,7 @@ class EMRSparkOperator(BaseOperator):
     airflow_bucket = environ['AIRFLOW_BUCKET']
     private_output_bucket = environ['PRIVATE_OUTPUT_BUCKET']
     public_output_bucket = environ['PUBLIC_OUTPUT_BUCKET']
+    staging_private_output_bucket = environ['STAGING_PRIVATE_OUTPUT_BUCKET']
```
sunahsuh: Please also add this to https://github.com/mozilla/telemetry-airflow/blob/master/Dockerfile and https://github.com/mozilla/telemetry-airflow/blob/master/Dockerfile.dev

sunahsuh:
Okay, that last comment is really the only thing left. We'll still need to work with @whd re: adding that variable to prod, so please don't merge before then :) (If that happens while you're out I can merge this when we're good to go)

robhudson:
Good catch. I also added the new env variable to the README.

README.md
```diff
@@ -75,6 +75,8 @@ variables:
   stored in, e.g. `telemetry-public-analysis-2`
 - `PRIVATE_OUTPUT_BUCKET` -- The AWS S3 bucket where private job results are
   stored in, e.g. `telemetry-parquet`
+- `STAGING_PRIVATE_OUTPUT_BUCKET` -- The AWS S3 bucket where private job
+  results for staging environments are stored in, e.g. `telemetry-parquet-staging`
```
whd: The example value here doesn't match the one in Dockerfile.dev.

whd:
Using production in a staging capacity (as opposed to stage) is probably fine, but it might also be possible to use staging for this (by running these specific tasks without the main_summary dependency). Stage is configured to have all dags off by default and output all data to telemetry-test-bucket.

If we want to go the route of running stage jobs from prod, I don't know if we need a new bucket specifically for staging data. Maybe exposing telemetry-test-bucket to production via an additional environment variable is sufficient. Otherwise we need to make the requisite change to IAM roles etc. for adding another bucket.

If we are adding environment variables, we might also want to add `telemetry-backfill` as a potential bucket of interest.

Aside from these thoughts I can coordinate merging and deploying this for some time this week. This will require a full redeploy, and recently we've had trouble doing those due to all the broken/retrying jobs. Things look somewhat cleaner this week though.

PR #156, opened 2017-08-11
Fix dev celery result backend url

sunahsuh:
I don't think you can template the backend url in this particular file like this.


sunahsuh:
Okay, tagging in @acmiyaguchi since he did some work dockerizing the python mozetl tests

acmiyaguchi:
Context for future me:

This is the error that should throw when running a task without this patch: `AttributeError: 'DisabledBackend' object has no attribute '_get_task_meta_for' airflow`.

sunahsuh:
Wouldn't that be expected behavior? I looked into the Dockerfile parser code a bit and that big ENV statement is [processed as a single line](https://github.com/docker/docker-ce/blob/master/components/engine/builder/dockerfile/parser/parser.go#L280) since we use one ENV directive + line continuation for that whole block and env variable interpolation is [done once](https://github.com/docker/docker-ce/blob/ef3fca8458fdc0aefaa68fb5502af5109a424b6f/components/engine/builder/dockerfile/evaluator.go#L160) per line before execution

acmiyaguchi:
I'm convinced, it seems like my case is not the norm. I wonder how I've managed to evade the celery issues so far.

PR #157, opened 2017-08-18
Add a job for filling the addon recommender HBase view

Dexterp37:
The data will flow into the instance that was setup [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1383739).

@jasonthomas, is the `aws` part in the .sh file the correct way to get the host? I cargo culted this from the other hbase operator we have.


Dexterp37:
Changing the reviewer to Mauro since he has more context and can review this :)

PR #161, opened 2017-08-22
Bug 1365580 - Create separate dag for daily and monthly search rollup

acmiyaguchi:
This adds daily and monthly search rollups as seperate dags that depend on `main_summary` using the `ExternalTaskSensor`. This lets us match up dags running at different frequencies.


acmiyaguchi:
Changes are rebased, and hopefully ready for a monthly run tomorrow.

PR #163, opened 2017-08-28
Make the HBase Addon Recommender job use EMR 5.8.0

Dexterp37:
This is needed otherwise, due to bug [SPARK-12837] in Spark < 2.2.0,
broadcasting during a join will result with an OOM error like: "Total
size of serialized results of XXX is bigger than
spark.driver.maxResultSize (YYY GB)"


Dexterp37:
The context from this change is in mozilla/telemetry-airflow#291 - the job is failing and I found that EMR 5.8.0 ships with Spark 2.2.0

PR #165, opened 2017-08-30
Add heavy_users job

fbertsch:


dags/main_summary.py
```diff
@@ -159,6 +159,17 @@
                        output_visibility="private",
                        dag=dag)
 
+t14 = EMRSparkOperator(task_id="heavy_users_view",
+                      job_name="Heavy Users View",
+                      owner="frank@mozilla.com",
+                      email=["telemetry-alerts@mozilla.com", "frank@mozilla.com"],
```
sunahsuh: Operational note -- would be nice to get one more person as backup when you're not around, in our continuing effort to remove single points of failure :)

(I suppose that should be me since I did the review on the job. Dammit.)

fbertsch:
@sunahsuh thank you for volunteering :)

PR #168, opened 2017-09-06
Comment out heavy_users to try and fix main_summary

fbertsch:


fbertsch:
@sunahsuh argh argh, you're right, changed :)

PR #170, opened 2017-09-07
Schedule the TAAR locale job

Dexterp37:
Hey Anthony, this PR is for scheduling the job you reviewed. This should run weekly, right after the longitudinal job, with the output being stored on a private bucket.

### DIFF
--- a/dags/longitudinal.py
+++ b/dags/longitudinal.py
@@ -2,6 +2,7 @@ from airflow import DAG
 from datetime import datetime, timedelta
 from operators.emr_spark_operator import EMRSparkOperator
 from utils.constants import DS_WEEKLY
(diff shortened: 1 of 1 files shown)


Dexterp37:
@acmiyaguchi , I bumped the instance count from 1 to 3, otherwise the job takes much more than 4 hours. Is this still good?

acmiyaguchi:
Yes, it still looks good. Roughly, how long does it take?

Dexterp37:
It takes about 3 hours to complete with 3 nodes. I've also updated the PR to use EMR 5.8.0 (SPark 2.2.0) since you mentioned that in the TAAR job fix PR due to the bug in [ApproxQuantile](https://issues.apache.org/jira/browse/SPARK-21550).

PR #171, opened 2017-09-08
Bug 1397048 - Embed search_rollups directly into dags

acmiyaguchi:
This removes the `ExternalTaskSensor` and makes the `search_rollup` job depend
on concrete instances of a job. The daily job will depend on the
`main_summary` dag, while the monthly job will run on the 2nd of the
month.

### DIFF
--- a/dags/search_rollup.py
+++ b/dags/search_rollup.py
@@ -24,25 +24,13 @@ default_args = {
 }
 
 
-dag_daily = DAG('search_rollup_daily',
-                 default_args=default_args,
-                 schedule_interval='@daily')
-
-
 dag_monthly = DAG('search_rollup_monthly',
                   default_args=default_args,
-                  schedule_interval='@monthly')
+                  schedule_interval='0 0 2 * *')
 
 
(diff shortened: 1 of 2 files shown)


dags/search_rollup.py
```diff
@@ -24,25 +24,13 @@
...
-dag_daily = DAG('search_rollup_daily',
-                 default_args=default_args,
-                 schedule_interval='@daily')
-
-
 dag_monthly = DAG('search_rollup_monthly',
                   default_args=default_args,
-                  schedule_interval='@monthly')
-
+                  schedule_interval='0 0 2 * *')
```
harterrt: Why change the interval here?

acmiyaguchi: We have to wait for data on the last day of the month to be processed before running the monthly job. The failure mode when we're missing a few days of data in a monthly rollup is subtle and currently not validated -- it would require asserting that the rows for all days of the current month exist. 

This could be offset from UTC+0 a few hours to account for `main_summary` running.

PR #173, opened 2017-09-15
Increase number of files for focus_event_longitudinal

fbertsch:


fbertsch:
@maurodoglio it is, I agree that it may be a good idea to add there, for the batch aggregate runs.

PR #180, opened 2017-10-02
Increase timeout and bump nodes for longitudinal

fbertsch:


dags/longitudinal.py
```diff
@@ -19,8 +19,8 @@
 
 t0 = EMRSparkOperator(task_id="longitudinal",
                       job_name="Longitudinal View",
-                      execution_timeout=timedelta(hours=10),
-                      instance_count=30,
+                      execution_timeout=timedelta(hours=12),
+                      instance_count=40,
```
mreid-moz: I could've sworn there was a max `instance_count` at one time... But I'm not seeing anything.

PR #187, opened 2017-10-10
Remove hbase main_summary tasks

jasonthomas:
https://bugzilla.mozilla.org/show_bug.cgi?id=1402322


fbertsch:
Warned users and removed code in `python_moztelemetry` in https://github.com/mozilla/python_moztelemetry/commit/e5dbc486855b1e76e7507f043d4e88fa56929774. Let's give it a few days to let the new code take effect, but we should be able to merge this relatively soon.

PR #191, opened 2017-10-18
Bug 1408162 - Schedule churn_v3

acmiyaguchi:
This schedules churn_v3 as the primary churn dataset. This will run concurrently with the old v2 dataset. I have [bug open for decommissioning churn_v2](https://bugzilla.mozilla.org/show_bug.cgi?id=1409584).


acmiyaguchi:
Thanks, I'll make sure to look into tagging.

PR #195, opened 2017-11-08
Bug 1415602 - Make date sanitation correct for future year boundaries

fbertsch:
This moves from a static year sanitation to a programmatically
determined one from the current year, allowing for edge
boundaries of Dec 31 and Jan 1 (where the year can be the
previous or the next).

Tested on Presto.


fbertsch:
@sunahsuh that is being handled by @relud in bug 1414281.

PR #196, opened 2017-11-08
Bug 1412971 - Create incremental client_count_daily data set

relud:
this will create a data set in `client_count_daily/v1/` partitioned by `submission_date`
r? @mreid-moz

### DIFF
--- a/dags/client_count_daily.py
+++ b/dags/client_count_daily.py
@@ -3,10 +3,10 @@ from datetime import datetime, timedelta
 from operators.emr_spark_operator import EMRSparkOperator
 
 default_args = {
-    'owner': 'frank@mozilla.com',
+    'owner': 'relud@mozilla.com',
     'depends_on_past': False,
-    'start_date': datetime(2016, 8, 13),
-    'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
+    'start_date': datetime(2016, 11, 1),
+    'email': ['telemetry-alerts@mozilla.com', 'relud@mozilla.com'],
     'email_on_failure': True,
     'email_on_retry': True,
     'retries': 2,
@@ -19,8 +19,8 @@ client_count_view = EMRSparkOperator(
     task_id="client_count_daily_view",
     job_name="Client Count Daily View",
     execution_timeout=timedelta(hours=10),
-    owner="frank@mozilla.com",
-    instance_count=20,
+    owner="relud@mozilla.com",
+    instance_count=5,
     env = {"date": "{{ ds_nodash }}", "bucket": "{{ task.__class__.private_output_bucket }}"},
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/client_count_daily_view.sh",
     dag=dag)
--- a/dags/client_count_daily.py
+++ b/dags/client_count_daily.py
@@ -13,14 +13,14 @@ default_args = {
     'retry_delay': timedelta(minutes=30),
 }
 
-dag = DAG('client_count', default_args=default_args, schedule_interval='@daily')
+dag = DAG('client_count_daily', default_args=default_args, schedule_interval='@daily')
 
 client_count_view = EMRSparkOperator(
-    task_id="client_count_view",
-    job_name="Client Count View",
+    task_id="client_count_daily_view",
+    job_name="Client Count Daily View",
     execution_timeout=timedelta(hours=10),
     owner="frank@mozilla.com",
     instance_count=20,
     env = {"date": "{{ ds_nodash }}", "bucket": "{{ task.__class__.private_output_bucket }}"},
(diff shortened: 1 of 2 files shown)


relud:
the first commit does a no-change copy of client_count, and the second commit renames it to client_count_daily, and adds three flags to spark that should make it incremental by submission_date

dags/client_count_daily.py
```diff
@@ -0,0 +1,26 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'frank@mozilla.com',
```
mreid-moz: s/frank/relud/ throughout :)

dags/client_count_daily.py
```diff
@@ -0,0 +1,26 @@
...
+    'depends_on_past': False,
+    'start_date': datetime(2016, 8, 13),
```
mreid-moz: Let's set this to a recent start date, then do backfill separately. I think we may want to add a couple more columns before we go back in time, so let's change this to `datetime(2016, 11, 1)`.

dags/client_count_daily.py
```diff
@@ -0,0 +1,26 @@
...
+}
+
+dag = DAG('client_count_daily', default_args=default_args, schedule_interval='@daily')
+
+client_count_view = EMRSparkOperator(
+    task_id="client_count_daily_view",
+    job_name="Client Count Daily View",
+    execution_timeout=timedelta(hours=10),
+    owner="frank@mozilla.com",
+    instance_count=20,
```
mreid-moz: instance_count can be reduced since we're only processing one day's data. Let's start with `5` and see how that performs.

jobs/client_count_daily_view.sh
```diff
@@ -0,0 +1,47 @@
...
+cd telemetry-batch-view
+sbt assembly
+
+base="normalized_channel,"
+base+="country,"
+base+="locale,"
+base+="app_name,"
+base+="app_version,"
+base+="e10s_enabled,"
+base+="e10s_cohort,"
```
mreid-moz: Per [Bug 1413515](https://bugzilla.mozilla.org/show_bug.cgi?id=1413515), let's remove `e10s_cohort`.

jobs/client_count_daily_view.sh
```diff
@@ -0,0 +1,47 @@
...
+base+="os,"
+base+="os_version"
+
+select="substr(subsession_start_date, 0, 10) as activity_date,"
```
mreid-moz: Per [Bug 1414281](https://bugzilla.mozilla.org/show_bug.cgi?id=1414281), let's be more selective about `activity_date`:
`regexp_extract(subsession_start_date, '(20[0-9]{2}-[0-9]{2}-[0-9]{2})', 1) as activity_date`

This implies also adding `AND activity_date IS NOT NULL` to the `--where` clause (which wouldn't work right if this were SQL, but it looks like it should work here -- please test)

jobs/client_count_daily_view.sh
```diff
@@ -0,0 +1,47 @@
...
+             --output-parition "submission_date=$date" \
+             --from $date \
+             --to $date \
+             --files "s3://telemetry-parquet/main_summary/v4/" \
+             --count-column "client_id" \
+             --select "$select" \
+             --grouping-columns "$group" \
+             --where "client_id IS NOT NULL" \
+             --output "$bucket/client_count_daily" \
+             --num-parquet-files 250
```
mreid-moz: With only a single day's data per partition, I think `2` parquet files per day should be good. This should result in files just a bit over 200MB in size if my math is correct.

dags/client_count_daily.py
```diff
@@ -0,0 +1,26 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'relud@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2016, 11, 1),
```
mreid-moz: s/2016/2017/

jobs/client_count_daily_view.sh
relud: tested, this worked for `submission_date=20171101`

relud:
`make run COMMAND="test client_count_daily client_count_daily_view 20171101"` succeeded, with two 500MB file outputs in `s3://telemetry-test-bucket/client_count_daily/v1/submission_date=20171101/` and took 14 minutes to run the EMR step after bootstrapping on 5 nodes.

relud:
some more test runs, for comparison:
2 nodes, 5 parquet files, 20171102 => 20 min, 200MB files
1 node, 5 parquet files, 20171103 => 48 min, 200MB files

mreid-moz:
Nice! Can you update to use 2 nodes, 5 parquet files?

relud:
done

PR #198, opened 2017-11-10
[Bug 1362229] Add PR tests for dag compilation

haroldwoo:
- Modify airflow conf to EFS dag mount
- add circle ci simple test for dag compilation


Dockerfile
```diff
@@ -37,6 +37,10 @@ RUN mkdir /app && \
     groupadd --gid 10001 app && \
     useradd --no-create-home --uid 10001 --gid 10001 --home-dir /app app
 
+# python-slim base image has missing folders required for psql install
```
whd: nit: prefer directory to folder.

PR #199, opened 2017-11-13
Bug 1412971 - Make client count daily depend on main summary

relud:
without this regularly scheduled client_count_daily runs are empty.


dags/main_summary.py
```diff
@@ -199,6 +199,18 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/retention.sh",
     dag=dag)
 
+client_count_daily_view = EMRSparkOperator(
+    task_id="client_count_daily_view",
+    job_name="Client Count Daily View",
+    execution_timeout=timedelta(hours=10),
+    owner="relud@mozilla.com",
+    start_date=datetime(2017, 11, 1),
```
mreid-moz: I don't think a separate `start_date` is needed for sub-tasks.

PR #200, opened 2017-11-14
Increase number of instances for main_summary job

mreid-moz:


mreid-moz:
cc @jasonthomas @harterrt 
Can y'all think of any reasons this might be a bad idea?

dags/main_summary.py
```diff
@@ -23,7 +23,7 @@
     task_id="main_summary",
     job_name="Main Summary View",
     execution_timeout=timedelta(hours=10),
```
harterrt: Should we decrease the `execution_timeout` in tandem?

mreid-moz: There's some slight possibility that this won't actually speed the job up much - I'd rather tighten up the timeout after we see what execution looks like in the wild.

PR #202, opened 2017-11-15
Increase timeout for longitudinal

fbertsch:
Timed out over the weekend.


fbertsch:
This doesn't seem to be needed any longer.

PR #203, opened 2017-11-15
Run pdsm after main summary

robotblake:
The idea here is to run the tool that updates glue immediately after main summary finishes.

Note: I'm still planning on turning this into a proper operator soon, just fighting code at the moment. 😞


jasonthomas:
I am r+, but @mreid-moz for final review.

dags/main_summary.py
```diff
@@ -210,6 +210,22 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/client_count_daily_view.sh",
     dag=dag)
 
+main_summary_glue = EMRSparkOperator(
+    task_id="main_summary_glue",
+    job_name="Main Summary Update Glue",
+    execution_timeout=timedelta(hours=2),
```
mreid-moz: Please add an `email` param to notify you if something goes wrong

mreid-moz:
Also I'm not sure what's going on with the CI failure.

dags/main_summary.py
```diff
@@ -210,6 +210,22 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/client_count_daily_view.sh",
     dag=dag)
 
+main_summary_glue = EMRSparkOperator(
```
fbertsch: This probably doesn't need to be an EMR instance, saving us a bit of bootstrap time. No change needed, just a consideration for the future.

PR #205, opened 2017-11-22
Temporarily disable scalar bootstrapping in experiment aggregates view

sunahsuh:
I'm working on profiling the job a bit to see where it'd be best to apply effort but for now we should make sure there's up-to-date data in the viewer


jobs/experiment_aggregates_view.sh
```diff
@@ -15,3 +15,4 @@ spark-submit --master yarn \
              --input "s3://$bucket/experiments/v1" \
              --output "s3://$bucket/experiments_aggregates/v1" \
              --date $date
```
fbertsch: Need a backslach here

sunahsuh: Whoops, thanks

PR #213, opened 2017-12-19
Bug 1421787 - Make deletion ping parquet dataset

relud:
Not sure if this will do the schema correctly. If not I think I will need to copy SyncView.scala and edit the schema.

deletion ping doc: https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/deletion-ping.html

### DIFF
--- a/dags/deletion.py
+++ b/dags/deletion.py
@@ -1,0 +1,26 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+
+default_args = {
+    'owner': 'relud@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 12, 18),
+    'email': ['telemetry-alerts@mozilla.com', 'relud@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 2,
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
I added in the `--doc-type` parameter recently to MainSummaryView. It'll probably have extra columns and nulls in all the wrong places. The schema is handcrafted to work with the main pings, and the option was meant for the first-shutdown ping. However, I ran into [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1419116) which mean I had to create a completely separate view. 

It'll probably need a new view but it seems like it doesn't have environment data or a payload, so it should be a really simple schema. Alternatively, the schema could be added to the [mozilla-pipeline-schemas](https://github.com/mozilla-services/mozilla-pipeline-schemas) repository like the new-profile ping. You can use it like the [crash import job](https://github.com/mozilla-services/data-pipeline/blob/master/reports/socorro_import/ImportCrashData.ipynb) or use direct-to-parquet (I think).

relud:
thanks @acmiyaguchi i'll do that

fbertsch:
@relud write the schema in mozilla-pipeline-schemas, that will be relatively straightforward.

PR #215, opened 2018-01-09
Bug 1290140 - Refactor telemetry-batch-view runner

relud:
refactoring a single runner, to start with.

@jezdez is this the kind of thing you had in mind for [Bug 1290140](https://bugzilla.mozilla.org/show_bug.cgi?id=1290140)?


dags/main_summary.py
```diff
@@ -203,15 +204,51 @@
...
     task_id="client_count_daily_view",
     job_name="Client Count Daily View",
     execution_timeout=timedelta(hours=10),
     owner="relud@mozilla.com",
     email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
     instance_count=2,
-    env = {"date": "{{ ds_nodash }}", "bucket": "{{ task.__class__.private_output_bucket }}"},
-    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/client_count_daily_view.sh",
+    env = {
+        "arguments": json.dumps({
```
acmiyaguchi: How about constructing this configuration struct as utility function like [mozetl_envvar](https://github.com/mozilla/telemetry-airflow/blob/master/dags/utils/mozetl.py). Users won't have to worry about serializing to the correct format, and you can change the underlying implementation as needed.

dags/main_summary.py
```diff
@@ -203,15 +204,51 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/retention.sh",
     dag=dag)
 
+client_count_daily_base = '''
```
acmiyaguchi: It might look nicer to represent this is a native list and concatenate using `"".join`.

dags/main_summary.py
acmiyaguchi: It might also be nice to drop the leading command prefix and add is as part of the utility function, but this is mostly stylistic.

jobs/telemetry_batch_view.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+
+from os import chdir
+from os import environ
+from subprocess import call
+from sys import stderr
+import json
+
+if not environ.get('class'):
```
acmiyaguchi: Errors in the configuration should be found at dag definition time. The CI should catch those and prevent us from deploying broken jobs.

relud: fixed

dags/main_summary.py

dags/main_summary.py

dags/main_summary.py
```diff
@@ -203,15 +205,49 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/retention.sh",
     dag=dag)
 
+client_count_daily_base = [
```
acmiyaguchi: Nit: The configuration for client count view feels out of place here since it had been isolated in a separate job file before. I'm okay with this as-is since most jobs tend to have less configuration.

dags/main_summary.py
```diff
@@ -203,15 +205,49 @@
...
+]
+
+client_count_daily_select = [
+    "regexp_extract(subsession_start_date, '(20[0-9]{2}-[0-9]{2}-[0-9]{2})', 1) as activity_date",
+    "devtools_toolbox_opened_count > 0 as devtools_toolbox_opened",
+    "case when distribution_id in ('canonical', 'MozillaOnline', 'yandex') then distribution_id else null end as top_distribution_id",
+] + client_count_daily_base
+
+client_count_daily_groups = [
+activity_date,
```
acmiyaguchi: indentation

dags/utils/tbv.py
```diff
@@ -0,0 +1,24 @@
...
+    environment variables must be prefixed by `TBV_`. For example, a class in
+    telemetry-batch-view taking a `--date` option can use `TBV_DATE` instead.
+
+    :klass string:      name of the class in telemetry-batch-view
+    :options dict:      environment variables to prefix
+    :other dict:        environment variables to pass through
+
+    :returns: a dictionary that contains properly prefixed class and options
+    """
+    prefixed_options = {
```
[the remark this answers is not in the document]
relud: Perhaps we could use two underscore to indicate a dash?

dags/main_summary.py
relud: What about moving it to a separate `.py` file and importing it?

dags/utils/tbv.py
```diff
@@ -0,0 +1,24 @@
+"""Utility functions for launching telemetry-batch-view jobs"""
+
+
+def tbv_envvar(klass, options, other={}):
+    """Set up environment variables for telemetry-batch-view jobs.
+
+    The command line interface can read options from the environment. All
```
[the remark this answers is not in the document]
relud: I like this better than the json approach. it makes `jobs/telemetry_batch_view.py` much cleaner

dags/main_summary.py

dags/main_summary.py
```diff
@@ -2,7 +2,9 @@
 from datetime import datetime, timedelta
 from operators.emr_spark_operator import EMRSparkOperator
 from utils.mozetl import mozetl_envvar
+from utils.tbv import tbv_envvar
 from search_rollup import add_search_rollup
+import json
```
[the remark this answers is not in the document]

dags/utils/tbv.py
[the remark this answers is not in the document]
acmiyaguchi: In the case of mozetl, the envvar wrapper and click definition will both downcase the dash to an underscore so there isn't a mismatch.

 Mixing underscores and dashes is inconsistent anyways. We should just enforce the convention of using dashes in the argument list by throwing an exception and letting CI/local testing catch it.

dags/main_summary.py
acmiyaguchi: Maybe, but it's also nice having the job definition available via the airflow dashboard.

dags/utils/tbv.py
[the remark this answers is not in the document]
relud: > We should just enforce the convention of using dashes in the argument list by throwing an exception and letting CI/local testing catch it.

in `tbv_envvars` or in [telemetry-batch-view](https://github.com/mozilla/telemetry-batch-view)? I can throw errors on `_`s in `tbv_envvars`, but the real place to catch this imo would be in [telemetry-batch-view](https://github.com/mozilla/telemetry-batch-view).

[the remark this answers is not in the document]
acmiyaguchi: Ideally, the cli tooling would tell you that your arguments are ill-formatted, but you would have to extend the functionality of scallop to catch the use of underscores in tbv. It might be more overhead than it's worth though. 

It should be fine as long as something automated catches the error before it reaches production.

jobs/telemetry_batch_view.py
```diff
@@ -0,0 +1,31 @@
...
+    if key.startswith("TBV_")
+], [])
+
+call(
+    'git',
+    'clone',
+    environ.get('GIT_PATH', 'https://github.com/mozilla/telemetry-batch-view.git'),
+    '--branch',
+    environ.get('GIT_BRANCH', 'master'),
+)
```
relud: @acmiyaguchi re-r? on this call that i just modified to support specifying git path and branch

acmiyaguchi: Looks good to me. I've implemented something similar with mozetl ([example](https://github.com/mozilla/telemetry-airflow/blob/master/dags/churn.py#L41) and [implementation](https://github.com/mozilla/python_mozetl/blob/master/bin/mozetl-submit.sh#L78)). It might make sense for me to change the script to be consistent between tbv and mozetl.

dags/utils/tbv.py
```diff
@@ -0,0 +1,31 @@
...
+        "tbv-{}".format(key): value
+        for key, value in options.items()
+    }
+    prefixed_options["tbv-class"] = klass
+    prefixed_options.update(other)
+
+    # raise ValueError if spaces found in non-templated envvar values
+    for item in prefixed_options.values():
+        if "{{" not in item and " " in item:
+            raise ValueError("env cannot contain spaces: '{}'".format(item))
```
relud: r? @acmiyaguchi I added this loop to validate that we aren't trying to pass spaces because that would break things.

I also changed the initial job being converted to one that doesn't need spaces in arguments.

I'll handle spaces in arguments in a separate PR.

jobs/telemetry_batch_view.py
```diff
@@ -0,0 +1,38 @@
...
+chdir("telemetry-batch-view")
+
+command = ["sbt", "assembly"]
+call_exit_errors(command)
+
+command = sum([
+    ["--{}".format(key[4:].lower().replace("_", "-")), value]
+    for key, value in environ.items()
+    if key.startswith("TBV_")
+], [
```
acmiyaguchi: Will the arguments here be put before the script?

acmiyaguchi:
In addition, has this been tested manually?

relud:
manual tests are running. they will confirm that the order of arguments is correct.

relud:
manual test run succeeded [j-35QS49RKPOCCS](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-35QS49RKPOCCS)

PR #221, opened 2018-01-31
Bug 1355990 - Port churn_to_csv to python_etl

acmiyaguchi:
This swaps the job from rtmo to python_mozetl.


acmiyaguchi:
Manual run via local airflow [here](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-OH9C8G8AJ9PF) using `--debug True`.

Checked that the size of the output was the same in the resulting buckets.

PR #222, opened 2018-02-02
Added TAAR ensemble to the airflow DAG.

crankycoder:
WIP: do not merge

@Dexterp37 Can you take a quick look at this to see if I'm setting this up correctly?  

I've got all my code for the taar dynamodb loader in a standalone egg file.

The entry point to run the data load is over here: https://github.com/crankycoder/taar_loader/blob/master/scripts/airflow_job.py 

And the bulk of the code is a straight port of the Scala code from https://github.com/mozilla/telemetry-batch-view/blob/master/src/main/scala/com/mozilla/telemetry/views/HBaseAddonRecommenderView.scala

The new python code currently lives in my repo over at: https://github.com/crankycoder/taar_loader/blob/master/taar_loader/

### DIFF
--- a/dags/longitudinal.py
+++ b/dags/longitudinal.py
@@ -40,6 +40,17 @@ addon_recommender = EMRSparkOperator(
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/addon_recommender.sh",
     dag=dag)
 
+ensemble_recommender = EMRSparkOperator(
+    task_id="ensemble_recommender",
+    job_name="Extract the addon data for each Telemetry user to DynamoDB",
+    execution_timeout=timedelta(hours=10),
+    instance_count=20,
+    owner="vng@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "vng@mozilla.com"],
+    env={},
+    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/ensemble_taar.sh",
+    dag=dag)
+
 game_hw_survey = EMRSparkOperator(
     task_id="game_hw_survey",
     job_name="Firefox Hardware Report",
@@ -114,3 +125,4 @@ cross_sectional.set_upstream(longitudinal)
 distribution_viewer.set_upstream(cross_sectional)
 taar_locale_job.set_upstream(longitudinal)
 taar_legacy_job.set_upstream(longitudinal)
+ensemble_recommender.set_upstream(longitudinal)
--- a/jobs/ensemble_taar.sh
+++ b/jobs/ensemble_taar.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+if [[ -z "$date" ]]; then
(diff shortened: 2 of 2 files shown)


jobs/ensemble_taar.sh
```diff
@@ -0,0 +1,19 @@
...
+   echo "Missing arguments!" 1>&2
+   exit 1
+fi
+
+# Setup a temp directory to work in 
+mkdir tmp_ensemble
+cd tmp_ensemble
+
+# Fetch the airflow job script
+wget https://raw.githubusercontent.com/crankycoder/taar_loader/releases/1.0/scripts/airflow_job.py
```
Dexterp37: I'm not sure if it's ok to load the script off a non-mozilla property

Dexterp37:
I just left a quick comment there. I don't own this repo, but @mreid-moz should be the right person to check this PR.

mreid-moz:
302 @sunahsuh

dags/longitudinal.py
```diff
@@ -40,6 +40,17 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/addon_recommender.sh",
     dag=dag)
 
+ensemble_recommender = EMRSparkOperator(
+    task_id="ensemble_recommender",
+    job_name="Extract the addon data for each Telemetry user to DynamoDB",
```
sunahsuh: This parameter is actually what the EMR cluster will be named -- it'd actually be kind of tough to tie that back to the correct airflow job with this name. Please pick a name that's closer to the task_id.

dags/longitudinal.py
```diff
@@ -40,6 +40,17 @@
...
+    execution_timeout=timedelta(hours=10),
+    instance_count=20,
+    owner="vng@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "vng@mozilla.com"],
```
sunahsuh: Is there another person that could be a secondary owner for this job?

dags/longitudinal.py
```diff
@@ -40,6 +40,17 @@
...
+    env={},
```
sunahsuh: This is the object that populates the environment variables on the cluster -- looks like your script below expects `date` to be set

dags/longitudinal.py
```diff
@@ -40,6 +40,17 @@
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/addon_recommender.sh",
     dag=dag)
 
+ensemble_recommender = EMRSparkOperator(
+    task_id="ensemble_recommender",
+    job_name="Extract the addon data for each Telemetry user to DynamoDB",
+    execution_timeout=timedelta(hours=10),
+    instance_count=20,
```
sunahsuh: Just want to verify that this timeout and instance count make sense for this job -- does it require 20 nodes?

acmiyaguchi:
It looks like this job was migrated to `python_mozetl/taar` and added to airflow in #243. Closing this PR.

PR #226, opened 2018-02-14
Add mango_log_processing DAG. Bug 1437635 

jasonthomas:
I've tested this locally and it seems to work as expected. I named this DAG mango_log_processing since the MR job is named mango. processlogs is a script to setup the EMR environment and run the MR job.

This is different from our other EMR jobs since it makes use of EmrCreateJobFlowOperator. 

Depends on https://github.com/mozilla-services/cloudops-deployment/pull/1603 which includes the EMR job flow json document.

r?


mreid-moz:
> it makes use of EmrCreateJobFlowOperator

Why does this DAG use the above operator?

mreid-moz:
No, I think this is fine, and indeed we should probably try to move our other jobs to the upstream EMR operators.

PR #229, opened 2018-02-16
move cross_sectional_view to telemetry_batch_view.py

relud:
add support for preserving case in tbv_envvar

r? frank as owner of `job_name="Cross Sectional View"`


dags/longitudinal.py
```diff
@@ -59,8 +60,11 @@
     job_name="Cross Sectional View",
     execution_timeout=timedelta(hours=10),
     instance_count=30,
-    env={"date": DS_WEEKLY, "bucket": "{{ task.__class__.private_output_bucket }}"},
-    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/cross_sectional_view.sh",
+    env = tbv_envvar("com.mozilla.telemetry.views.CrossSectionalView", {
+        "executor-cores": 8,
```
acmiyaguchi: @fbertsch is this option even needed? I don't see why it would be desirable to use less than the amount of parallelism available.

jobs/telemetry_batch_view.py
```diff
@@ -25,15 +25,17 @@ def call_exit_errors(command):
 command = ["sbt", "assembly"]
 call_exit_errors(command)
 
-command = sum([
-    ["--{}".format(key[4:].lower().replace("_", "-")), value]
-    for key, value in environ.items()
-    if key.startswith("TBV_") and key != "TBV_CLASS"
-], [
+command = ["spark-submit"]
+if "TBV_executor_cores" in environ:
```
acmiyaguchi: Can options be passed without having to create these special cases?

relud: we could rework how options that go before the jar get defined

dags/longitudinal.py
acmiyaguchi: @harterrt  I think you're the appropriate one to tag here

jobs/telemetry_batch_view.py
```diff
@@ -25,15 +25,17 @@ def call_exit_errors(command):
...
+    command += ["--executor-cores", environ["TBV_executor_cores"]]
+command += [
     "spark-submit",
```
acmiyaguchi: I think you forgot to remove this

dags/longitudinal.py
relud: also used in [jobs/focus_event_longitudinal_view.sh#L13](https://github.com/mozilla/telemetry-airflow/blob/master/jobs/focus_event_longitudinal_view.sh#L13)

jobs/telemetry_batch_view.py
relud: I did. fixed.

dags/longitudinal.py
acmiyaguchi: @harterrt bumping

Aside from that, I think this option should be moved into the `other` argument as opposed to the jobs argument.

harterrt: Sorry - I started a review and never submitted it.  

This is probably copy-pasta from the Longitudinal job, so I'm not attached to it. I certainly can't argue that it's necessary.

relud: okay, dropped it

dags/utils/tbv.py
```diff
@@ -17,7 +17,7 @@ def tbv_envvar(klass, options, other={}):
     :returns: a dictionary that contains properly prefixed class and options
     """
     prefixed_options = {
-        "TBV_{}".format(key.upper().replace("-", "_")): value
+        "TBV_{}".format(key.replace("-", "_")): value
```
relud: @acmiyaguchi I am concerned that this change will lag behind removing `.lower()` in `jobs/telemetry_batch_view.py`, which would cause jobs to fail. Could that happen?

relud:
> Do you think it would be worthwhile to have tests in-line as the script gets more complex?

assuming we're talking about `jobs/telemetry_batch_view.py`

ideally we will avoid letting the script get more complex, but for cases where it's an issue, we should catch known problems and throw useful errors.

dags/utils/tbv.py
acmiyaguchi: This is regards to the github repository not updating the raw file at the master location? I think it's a somewhat rare occurrence, but maybe separating this change into a separate patch would be a good idea.

What was the reasoning behind dropping `upper` in the first place?

relud: we dropped upper because we are preserving case now, because we have camel case args to pass. my concern is not that github will be slow. my concern is that airflow will not get the latest utils/tbv.py as fast as it gets the latest jobs/telemetry_batch_view.py for other jobs already using `jobs/telemetry_batch_view.py`, thus sending it upper cased env vars that become upper cased  args, and failing a dag run.

acmiyaguchi: I think the autodeploy mechanism will also update the utils code. We direcly import the tbv module into all of the dags, which means that the jobs encoded in the dags and the utility python code should get updated at the same time.

I'm also a little wary about preserving case in the environment since the convention is to have them all uppercased, but it looks like it should be fine. Maybe something to think about is an alternative way of passing context to the operator.

PR #242, opened 2018-03-01
Add main summary backfill DAG

sunahsuh:
It's nearly showtime!

Of note: I changed the max_active_runs to 10 so we should run 10 weeks of backfill at a time. At that rate I'm estimating the whole year's backfill will take ~6 days.

Do not merge until the last backfill PR is merged in tbv

### DIFF
--- a/dags/main_summary_backfill.py
+++ b/dags/main_summary_backfill.py
@@ -1,0 +1,36 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+from utils.constants import DS_WEEKLY
+from utils.tbv import tbv_envvar
+
+default_args = {
+    'owner': 'ssuh@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 3, 1),
+    'email': ['telemetry-alerts@mozilla.com', 'ssuh@mozilla.com', 'frank@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 2,
+    'retry_delay': timedelta(minutes=15)
+}
+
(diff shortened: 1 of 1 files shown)


dags/main_summary_backfill.py
```diff
@@ -0,0 +1,36 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from operators.emr_spark_operator import EMRSparkOperator
+from utils.constants import DS_WEEKLY
+from utils.tbv import tbv_envvar
+
+default_args = {
+    'owner': 'ssuh@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2017, 3, 1),
```
fbertsch: This date may or may not be full/partially full/empty when we run, we ran into some problem with that last time. We can just verify when we're done that it has all the rows, and if not just don't copy it over.

dags/main_summary_backfill.py
```diff
@@ -0,0 +1,36 @@
...
+    'email': ['telemetry-alerts@mozilla.com', 'ssuh@mozilla.com'],
```
fbertsch: You can add me on here: `frank@mozilla.com`

dags/main_summary_backfill.py
```diff
@@ -0,0 +1,36 @@
...
+          default_args=default_args,
+          schedule_interval='@weekly',
+          catchup=True,
+          max_active_runs=10)
+
+t0 = EMRSparkOperator(task_id="main_summary_backfill",
+                      job_name="Main Summary Backfill",
+                      execution_timeout=timedelta(hours=48),
+                      instance_count=20,
+                      env=tbv_envvar("com.mozilla.telemetry.views.MainSummaryView", {
```
fbertsch: fancy!

PR #243, opened 2018-03-01
Added DAG entry for TAAR dynamo loader.

crankycoder:
@sunahsuh I've added the TAAR dynamo job here as a dependency on the main_summary.

Am I missing something to schedule this daily?

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -251,6 +251,21 @@ main_summary_glue = EMRSparkOperator(
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/update_glue.sh",
     dag=dag)
 
+taar_dynamo = EMRSparkOperator(
+    task_id="taar_dynamo",
+    job_name="TAAR DynamoDB loader",
+    execution_timeout=timedelta(hours=14),
+    instance_count=6,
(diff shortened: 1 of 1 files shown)


dags/main_summary.py
```diff
@@ -251,6 +251,21 @@
...
 
+taar_dynamo = EMRSparkOperator(
+    task_id="taar_dynamo",
+    job_name="TAAR DynamoDB loader",
+    execution_timeout=timedelta(hours=14),
+    instance_count=6,
+    owner="vng@mozilla.com",
+    email=["mlopatka@mozilla.com", "vng@mozilla.com", "sbird@mozilla.com"],
+    env=mozetl_envvar("taar_dynamo", {
+        "date": "{{ yesterday_ds_nodash }}"
```
sunahsuh: if you use `yesterday_ds_nodash`, this will always run for yesterday's date, regardless of what the scheduled execution date is for the current task run (e.g. if we find an error in a main_summary run a week later and re-run, this task will still run for yesterday's date, even though the DAG is running for last week). You probably want `ds_nodash` instead.

crankycoder: Thanks - I've replaced this now.

PR #247, opened 2018-03-12
move crash_summary_view to telemetry_batch_view.py

relud:
cc @maurodoglio as job owner


relud:
well, we should probably change the job owner then. @acmiyaguchi do you know who the appropriate replacement would be?

sunahsuh:
If you change the owner to please update https://bugzilla.mozilla.org/show_bug.cgi?id=1442338 :)

acmiyaguchi:
I don't know who the right owner is, maybe a better question for @mreid-moz

PR #251, opened 2018-03-14
Bug 1445646: Fix Sensor Timeouts in Main Summary Backfill

sunahsuh:
This PR does two things:
- Set the celery visiblity_timeout to 6 hours, up from the default 1 hour which was causing our step sensors to fail (https://stackoverflow.com/questions/44274381/airflow-long-running-task-in-subdag-marked-as-failed-after-an-hour)
- Serialize each day's step in airflow to run one after the other, in which case sensors should not run past the new 6 hour limit

Setting the visibility_timeout to a higher limit has the downside of slower reaction times when a worker has failed and a task should be requeued. This isn't a very common event, and setting it to 6 hours (vs 1 day, which apparently some people do to work around this airflow issue) seems like a good compromise.

### DIFF
--- a/dags/main_summary_backfill_mar_2018.py
+++ b/dags/main_summary_backfill_mar_2018.py
@@ -5,7 +5,6 @@ from airflow.models import DagBag
 from airflow.utils.trigger_rule import TriggerRule
 from airflow.contrib.operators.emr_terminate_job_flow_operator import EmrTerminateJobFlowOperator
 from airflow.contrib.sensors.emr_step_sensor import EmrStepSensor
-from airflow.operators.sensors import TimeDeltaSensor
 from airflow.operators.subdag_operator import SubDagOperator
 from airflow.operators.moz_emr import (EmrAddStepsOperator,
                                        EmrCreateJobFlowSelectiveTemplateOperator,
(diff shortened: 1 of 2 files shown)


plugins/celery_visibility_timeout_fix.py
```diff
@@ -0,0 +1,5 @@
+from airflow.executors.celery_executor import app
+
+# We need to override visibility_timeout in a redis-backed celery queue since the default
+# is one hour, meaning sensors are marked as failed after one hour
```
mreid-moz: Further reference:
http://docs.celeryproject.org/en/latest/getting-started/brokers/redis.html#id1

dags/main_summary_backfill_mar_2018.py
```diff
@@ -65,13 +64,6 @@ def main_summary_subdag_factory(parent_dag, task_id, day):
     parent_job_flow_id = ("{{{{ task_instance.xcom_pull('setup_backfill_cluster', "
                           "key='return_value', dag_id={}) }}}}".format(parent_dag.dag_id))
 
-    # Try to alleviate throttling issues by introducing some slight jitter on each of the days
```
acmiyaguchi: Is removing this related to the limited number of celery workers?

sunahsuh: No, we just don't need to introduce artificial jitter anymore since we're not submitting the steps all at once

PR #252, opened 2018-03-15
fix bugs in recent PRs

relud:


relud:
apparently integers aren't iterable and that's a problem for i don't know why

PR #253, opened 2018-03-15
Fix `make build` on osx

relud:
I can't run `make build` on osx for testing dags without this


acmiyaguchi:
This seems strange since it should be running within the container, but I can't confirm since I'm running Linux.

relud:
I mean, I assume it works on linux, or someone would have noticed before me, right?

either way, i assumed this was due to having a newest version of python:2-slim that other people have cached

PR #254, opened 2018-03-15
Backfill tweaks

sunahsuh:
3 small changes here:
- Up the parallelism on the backfill to 4 DAGs at once
- Make cluster termination dependent on every main summary day completing successfully (which allows us to selectively re-run days and have the cluster auto-terminate on completion)
- actually use the `action_on_failure` param for job steps


sunahsuh:
> This looks like the graph is short-circuited to cluster termination when a job takes too many retries.

So, the cluster termination operator is set to `all_done`, which means it will only run when all immediately downstream tasks are done (either failed or success). When it was only dependent on the last day, the success criteria is only if the last day is done, so even if there are days further downstream that are running, the termination would run immediately when cleared.

PR #255, opened 2018-03-16
Update dag owners

mreid-moz:


dags/android_clients.py
```diff
@@ -21,7 +21,7 @@
     execution_timeout=timedelta(hours=4),
     instance_count=5,
     owner="frank@mozilla.com",
-    email=["telemetry-alerts@mozilla.com", "frank@mozilla.com", "frank@mozilla.com"],
```
sunahsuh: lol wat

dags/probe_scraper.py
```diff
@@ -22,7 +22,6 @@
     job_name="Probe Scraper",
     execution_timeout=timedelta(hours=4),
     instance_count=1,
-    owner="gfritzsche@mozilla.com",
```
sunahsuh: I'm confused why we're removing these owners

mreid-moz: They're already specified in the `default_args` above. We could remove the emails too by updating the defaults.

PR #256, opened 2018-03-16
Add clients_daily_v6 job

relud:
don't want to interrupt the existing clients_daily dataset, so adding a new task for it

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -198,6 +198,20 @@ clients_daily = EMRSparkOperator(
     uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-submit.sh",
     dag=dag)
 
+clients_daily_v6 = EMRSparkOperator(
+    task_id="clients_daily_v6",
+    job_name="Clients Daily v6",
+    owner="relud@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
+    execution_timeout=timedelta(hours=5),
+    instance_count=10,
+    env=tbv_envvar("com.mozilla.telemetry.views.ClientsDailyView", {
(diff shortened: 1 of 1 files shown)


sunahsuh:
Writing directly to v6 will potentially disrupt folks' workflow as soon as the metastore updates run the first time (probably some time within the first 24 hours of the run) -- even though there will only be partial data. Are you planning to go back as far as you can in history with the backfill?

relud:
I didn't realize v6 would be automatically detected, nor that clients_daily would automatically be updated to it. [Bug 1426172](https://bugzilla.mozilla.org/show_bug.cgi?id=1426172) has meg backfilling to 1 year. v5 currently goes back to 2017-08-31, but is incompatible because it uses activity_date instead of submission_date.

relud:
this is blocked on [Bug 1448134](https://bugzilla.mozilla.org/show_bug.cgi?id=1448134)

relud:
@sunahsuh I've modified this to write to the telemetry-backfill bucket until such a time as we are ready for the v6 dataset

dags/main_summary.py
```diff
@@ -198,6 +198,20 @@
...
+clients_daily_v6 = EMRSparkOperator(
+    task_id="clients_daily_v6",
+    job_name="Clients Daily v6",
+    owner="relud@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
+    execution_timeout=timedelta(hours=5),
+    instance_count=10,
+    env=tbv_envvar("com.mozilla.telemetry.views.ClientsDailyView", {
+        "date": "{{ ds_nodash }}",
+        "output-bucket": "telemetry-backfill" # "{{ task.__class__.private_output_bucket }}"
```
sunahsuh: Please remove the comment with the old code

PR #267, opened 2018-04-11
Add Usage Report

fbertsch:


sunahsuh:
Is not tying the job to a successful run of main_summary going to be an issue?

Otherwise, looks fine to me -- my concern re: mozetl is that this seems to be setting us back a bit from the general goal of reducing the amount of boilerplate code in our repos. Is there a plan going forward to align the job with mozetl (either by moving the spark pieces or by making mozetl constructs a library we can base external repos on, which is probably the better long-term solution.)

PR #268, opened 2018-04-16
Removed explicit temp directory in taar_amodump job 

crankycoder:
Fallback to the default using mkdtemp.

The default path is specified here: 

https://github.com/mozilla/python_mozetl/blob/master/mozetl/taar/taar_amodump.py#L327


crankycoder:
cc/ @acmiyaguchi

PR #269, opened 2018-04-16
Fx usage updates

fbertsch:


jobs/fx_usage_report.sh
```diff
@@ -12,4 +12,5 @@ cd Fx_Usage_Report
 python usage_report/usage_report.py \
     --date $date \
     --sample 10 \
-    --output-bucket $bucket
+    --output-bucket $bucket \
+    --output-prefix "$deploy_environment/usage_report_data"
```
sunahsuh: The intent here is to make it easier to figure out what's writing the data, I'm guessing?

dags/fx_usage_report.py
```diff
@@ -21,7 +21,7 @@
     task_id='wait_for_main_summary',
     external_dag_id='main_summary',
     external_task_id='main_summary',
-    execution_delta=timedelta(hours = -1), # main_summary waits one hour
+    execution_delta=timedelta(days=-7, hours=-1), # main_summary waits one hour, execution date is beginning of the week
```
sunahsuh: Ugh, airflow.

fbertsch:
@sunahsuh yes it is intentional, but unfortunately Airflow doesn't actually allow me to backfill those dates. I'll leave it in because I'm backfilling manually

dags/fx_usage_report.py
fbertsch: yeah, great stuff :)

jobs/fx_usage_report.sh
fbertsch: Yeah, eventually I'll move to have code first deployed to stage and then to prod. So for now these will both be the same.

PR #270, opened 2018-04-20
Update experiments analysis view

sunahsuh:
Removes the --nobootstrapScalars option and removes the job script in favor of tbv_envvar


sunahsuh:
@fbertsch mind taking a look at the last commit? I made a change to fix CI (I'll file another bug for removal of the line when it's fixed upstream)

PR #273, opened 2018-04-23
Bug 1444413 - remove cross_sectional

relud:
@harterrt for review as bugzilla component owner


harterrt:
Hmm, let's make sure this isn't necessary for the distribution viewer. @robhudson can you comment on whether the distribution viewer is actively used?

relud:
@harterrt distribution viewer is decommissioned, so this is ready to go

PR #274, opened 2018-04-24
Bug 1456549 - Fix search rollup timeout and increase instance count

acmiyaguchi:
[Bug Link](https://bugzilla.mozilla.org/show_bug.cgi?id=1456549)

The search rollup job has been failing intermittently over the last few weeks. I left out some key arguments in the DAG definition that prevented the intended email recipients from getting the retry alerts. I've also bumped up the instance count and timeout value.

### DIFF
--- a/dags/search_rollup.py
+++ b/dags/search_rollup.py
@@ -1,12 +1,12 @@
 from datetime import timedelta, datetime
 
 from airflow import DAG
-from airflow.operators import ExternalTaskSensor
 
 from operators.emr_spark_operator import EMRSparkOperator
 from utils.mozetl import mozetl_envvar
 
 
+# These settings only apply to the monthly DAG
 default_args = {
     'owner': 'amiyaguchi@mozilla.com',
     'depends_on_past': False,
@@ -15,12 +15,11 @@ default_args = {
         'telemetry-alerts@mozilla.com',
         'amiyaguchi@mozilla.com',
         'harterrt@mozilla.com',
-        'spenrose@mozilla.com'
     ],
     'email_on_failure': True,
     'email_on_retry': True,
     'retries': 2,
-    'retry_delay': timedelta(minutes=30),
+    'retry_delay': timedelta(hours=2),
 }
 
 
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
@sunahsuh Bumping for review, also feel free to untag yourself.

I verified that there was an issue with the days that timed out related to a OOM error. Bumping the instances up to 3 fixes the underlying problem.

dags/search_rollup.py
```diff
@@ -30,11 +29,17 @@
...
+
+    This can be called with an optional DAG passed into `upstream`. The rollup
+    job will inherit the default values of the referenced DAG.
+    """
 
     search_rollup = EMRSparkOperator(
         task_id="search_rollup_{}".format(mode),
         job_name="{} search rollup".format(mode).title(),
+        owner="amiyaguchi@mozilla.com",
+        email=default_args["email"],
```
sunahsuh: So, wait, default_args doesn't apply the email default arg? 😧

dags/search_rollup.py
```diff
@@ -30,11 +29,17 @@
 
 
 def add_search_rollup(dag, mode, instance_count, upstream=None):
-    """Create a search rollup for a particular date date"""
+    """Create a search rollup for a particular date date
+
+    This can be called with an optional DAG passed into `upstream`. The rollup
```
sunahsuh: I think you mean "task" instead of "DAG" here

dags/search_rollup.py
sunahsuh: Oh wait, I see.. you're referencing the `default_args` variable in the file. Hmm, I think in an ideal world we'd pass in those parameters to the `add_search_rollup` function, or pass in `default_args` with the double-star

dags/search_rollup.py
```diff
@@ -15,12 +15,11 @@
         'telemetry-alerts@mozilla.com',
         'amiyaguchi@mozilla.com',
         'harterrt@mozilla.com',
-        'spenrose@mozilla.com'
     ],
     'email_on_failure': True,
     'email_on_retry': True,
     'retries': 2,
-    'retry_delay': timedelta(minutes=30),
+    'retry_delay': timedelta(hours=2),
```
sunahsuh: Is this something that's relevant just for this DAG's task instance and not for the task in main summary?

acmiyaguchi: Actually, this doesn't do what I meant it to do. This should be omitted.

acmiyaguchi:
@sunahsuh I moved the function to the top for clarity and removed some unnecessary defaults. Hopefully it should read a little clearer now.

PR #277, opened 2018-04-27
t-b-v deploy using artifacts

fbertsch:
@relud  you may want to take a look as well, since you made these scripts.

### DIFF
--- a/dags/utils/deploy.py
+++ b/dags/utils/deploy.py
@@ -1,8 +1,16 @@
 
-def get_artifact_url(slug, region="{{ task.__class__.region }}",
+def get_artifact_url(slug, branch=None, tag=None,
+                     region="{{ task.__class__.region }}",
                      bucket="{{ task.__class__.artifacts_bucket }}",
-                     owner_slug="{{ test.__class__.mozilla_slug }}",
+                     owner_slug="{{ task.__class__.mozilla_slug }}",
                      deploy_tag="{{ task.__class__.deploy_tag }}"):
+    assert not (branch and tag), "Cannot build from both a branch and a tagged release"
+
+    if branch:
+        deploy_tag = branch
+    elif tag:
+        deploy_tag = "tags/{}".format(tag)
+
     return "https://s3-{region}.amazonaws.com/{bucket}/{owner_slug}/{slug}/{deploy_tag}/{slug}.jar".format(
         region=region, bucket=bucket, owner_slug=owner_slug, slug=slug, deploy_tag=deploy_tag)
 
--- a/dags/utils/deploy.py
+++ b/dags/utils/deploy.py
@@ -1,0 +1,8 @@
+
+def get_artifact_url(slug, region="{{ task.__class__.region }}",
+                     bucket="{{ task.__class__.artifacts_bucket }}",
+                     owner_slug="{{ test.__class__.mozilla_slug }}",
+                     deploy_tag="{{ task.__class__.deploy_tag }}"):
+    return "https://s3-{region}.amazonaws.com/{bucket}/{owner_slug}/{slug}/{deploy_tag}/{slug}.jar".format(
+        region=region, bucket=bucket, owner_slug=owner_slug, slug=slug, deploy_tag=deploy_tag)
+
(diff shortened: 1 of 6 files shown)


jobs/telemetry_batch_view.py
```diff
@@ -1,33 +1,30 @@
...
-        "git",
-        "clone",
-        environ.get("GIT_PATH", "https://github.com/mozilla/telemetry-batch-view.git"),
-        "--branch",
-        environ.get("GIT_BRANCH", "master"),
-    ]
-
-    call_exit_errors(command)
+def retrieve_jar():
+    jar_url = environ.get("ARTIFACT_URL")
```
relud: I think we should `retreive_jar` only when `GIT_PATH` and `GIT_BRANCH` are defaults, otherwise we should clone and compile.

fbertsch: We already handle that case with the deploy script. Engineers can just push to a branch on `mozilla/$SLUG` and that branch will get built. Locally they can test Airflow using that artifact.

Instead, let's add an optional `BRANCH` or `TAG` (not both) which will let engineers specify builds for their jobs. `BRANCH` will point to an auto-updated artifact, but then to update jobs that use a previous release, the engineer will have to update their Airflow job manually (kind of annoying).

fbertsch: I want to move away from the build and compile completely, and focus on just using built jars.

jobs/telemetry_batch_view.py
```diff
@@ -40,16 +37,14 @@ def submit_job():
...
         "--deploy-mode", "client",
         "--class", environ["TBV_CLASS"],
-        "target/scala-2.11/telemetry-batch-view-1.1.jar",
+        artifact_file,
     ])
+
     call_exit_errors(command)
 
 
 if environ.get("DO_ASSEMBLY", "True") == "True":
```
whd: If the plan is to move away from build and compile completely, `DO_ASSEMBLY` should probably be removed/renamed.

fbertsch:
@relud updated with those changes. I had to make some changes to the backfill job as well to make things work.

jobs/telemetry_batch_view.py
```diff
@@ -1,33 +1,30 @@
 #!/usr/bin/env python
 
+from requests
```
whd: Probably supposed to be "import requests".

jobs/telemetry_batch_view.py
```diff
@@ -1,33 +1,30 @@
...
+    jar_url = environ.get("ARTIFACT_URL")
 
-    print("+ cd telemetry-batch-view")
-    chdir("telemetry-batch-view")
+    if jar_url is None:
+        exit(1)
 
-    command = ["sbt", "assembly"]
-    call_exit_errors(command)
+    response = requests.get(jar_url)
```
whd: I'm almost offended enough by the potential jar size to say we should do streaming iteration of the request. Almost.

jobs/telemetry_batch_view.py
fbertsch: hmm, concerning that I missed that. I'll make sure to run this locally.

jobs/telemetry_batch_view.py
fbertsch: You're not wrong. Let's revisit in 6 months when the jars are >1GB.

jobs/telemetry_batch_view.py
fbertsch: Annoying that launching locally still retrieves the file from the master Airflow repo. Something else to fix...

dags/utils/deploy.py
```diff
@@ -0,0 +1,16 @@
+
+def get_artifact_url(slug, branch=None, tag=None,
+                     region="{{ task.__class__.region }}",
+                     bucket="{{ task.__class__.artifacts_bucket }}",
+                     owner_slug="{{ test.__class__.mozilla_slug }}",
```
relud: it looks odd that this is test instead of task, but it was like that where it was copied from so 🤷‍♂️

jobs/telemetry_batch_view.py
relud: we've tried this, right? I would not be suprised if >100MB was enough cause a problem with this, but if it works then good enough.

fbertsch:
@relud it's ran fine on two tests so far. The fact that they are both sharing the AWS network, and in fact maybe the same DC, probably helps

PR #285, opened 2018-05-11
Bug 1457268 - Add operator that emails schema changes

relud:


dags/operators/email_schema_emr_spark_operator.py
```diff
@@ -0,0 +1,55 @@
+import boto3
+import re
+from airflow.exceptions import AirflowException
+from airflow.utils.decorators import apply_defaults
+from airflow.utils.email import send_email
+from difflib import unified_diff
+from .emr_spark_operator import EMRSparkOperator
+
+class EmailSchemaEMRSparkOperator(EMRSparkOperator):
```
sunahsuh: I don't think we should conflate responsibilities for operators like this -- seems like it'd make more sense to make this a separate task that runs after main_summary. Consider the case where we fail to send an email because the SMTP service is down, or somesuch -- we'd re-run the entirety of main_summary for that. Also, if we make the email task separate we can inherit from the built-in EmailOperator which seems like a more obvious lineage.

relud: fixed

dags/operators/email_schema_change_operator.py
```diff
@@ -0,0 +1,69 @@
...
+    :param latest_schema: The name of the job.
+    :type latest_schema: string
+
+    :param previous_schema: The e-mail address of the user owning the job.
+    :type previous_schema: string
+
+    and all EmailOperator params except html_content
+    """
+    template_fields = EmailOperator.template_fields + ('latest_schema', 'previous_schema')
+    s3_regex = re.compile('([Ss]3://)?(?P<Bucket>[^/]*)/(?P<Key>.+)')
```
sunahsuh: Any particular reason to do this vs passing in bucket/key/dates as separate params to begin with?

dags/operators/email_schema_change_operator.py
```diff
@@ -0,0 +1,69 @@
...
+    # s3 buckets from env
+    spark_bucket = environ['SPARK_BUCKET']
+    airflow_bucket = environ['AIRFLOW_BUCKET']
+    private_output_bucket = environ['PRIVATE_OUTPUT_BUCKET']
+    public_output_bucket = environ['PUBLIC_OUTPUT_BUCKET']
+    artifacts_bucket = environ['ARTIFACTS_BUCKET']
+
+    @apply_defaults
+    def __init__(self, latest_schema='', previous_schema='',
+                 subject='Airflow: Schema change detected: {{ task }}', *args, **kwargs):
```
sunahsuh: Can we add the dates to the subject?

dags/operators/email_schema_change_operator.py
```diff
@@ -0,0 +1,69 @@
...
+                    **previous_schema.groupdict())['Body'].read().splitlines(True)
+        except client.exceptions.NoSuchKey:
+            logging.info('previous_schema not found, skipping change detection')
+            return  # if there's no previous schema, don't try to compare
+        latest = client.get_object(
+                **latest_schema.groupdict())['Body'].read().splitlines(True)
+        diff = list(unified_diff(previous, latest))
+        if diff:
+            logging.info('schema change detected, sending email')
+            self.html_content = ''.join(diff).replace('\n', '<br>')
```
sunahsuh: Can we add the dates in the body as well?

dags/operators/email_schema_change_operator.py
```diff
@@ -0,0 +1,69 @@
...
+    s3_regex = re.compile('([Ss]3://)?(?P<Bucket>[^/]*)/(?P<Key>.+)')
+    latest_schema = ''
+    previous_schema = ''
+
+    # s3 buckets from env
+    spark_bucket = environ['SPARK_BUCKET']
+    airflow_bucket = environ['AIRFLOW_BUCKET']
+    private_output_bucket = environ['PRIVATE_OUTPUT_BUCKET']
+    public_output_bucket = environ['PUBLIC_OUTPUT_BUCKET']
+    artifacts_bucket = environ['ARTIFACTS_BUCKET']
```
sunahsuh: These are no longer used in the operator, correct?

dags/operators/email_schema_change_operator.py
```diff
@@ -0,0 +1,69 @@
...
+
+    :param previous_schema: The e-mail address of the user owning the job.
+    :type previous_schema: string
+
+    and all EmailOperator params except html_content
+    """
+    template_fields = EmailOperator.template_fields + ('latest_schema', 'previous_schema')
+    s3_regex = re.compile('([Ss]3://)?(?P<Bucket>[^/]*)/(?P<Key>.+)')
+    latest_schema = ''
+    previous_schema = ''
```
sunahsuh: Don't think we need to declare these here if we populate then in `__init__`

dags/main_summary.py
```diff
@@ -28,10 +30,19 @@
...
         "to": "{{ ds_nodash }}",
+        "schema-report-location": main_summary_schema+"{{ ds_nodash }}",
         "bucket": "{{ task.__class__.private_output_bucket }}"}),
     uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/telemetry_batch_view.py",
     dag=dag)
 
+main_summary_schema = EmailSchemaChangeOperator(
+    task_id="main_summary_schema",
+    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
+    to=["telemetry-alerts@mozilla.com"],
```
sunahsuh: IIRC in the bug, we talked about limiting the `to` to folks that are specifically interested (you can add me! :D)

dags/operators/email_schema_change_operator.py
relud: private_output_bucket is used to define the schema location

PR #286, opened 2018-05-16
Added more workers and extended runtime for taar_lite job

crankycoder:
The taar_lite job is currently timing out in airflow as we do not have enough workers or time to complete the job.  This just bumps the instance count to 5 and extends the runtime to 2 hours which should be more than sufficient to execute the job.


crankycoder:
cc/ @mlopatka

PR #287, opened 2018-05-16
Sync probe-scraper cache before and after running

fbertsch:
DO NOT MERGE YET

@jasonthomas, do you have any recommendations for which bucket location we should use to store the probe-scraper cache?

@Dexterp37 also note that with this change we should probably remove the requests cache database object.

### DIFF
--- a/jobs/probe_scraper.sh
+++ b/jobs/probe_scraper.sh
@@ -3,9 +3,12 @@
 # Print all the executed commands to the terminal.
 set -x
 
-BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
+OUTPUT_BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
+CACHE_BUCKET="telemetry-private-analysis-2/probe-scraper"
 CACHE_DIR="probe_cache"
 OUTPUT_DIR="probe_data"
+
+aws s3 sync s3://$CACHE_BUCKET/ $CACHE_DIR/ --delete
 
 # Clone and setup the scraper
 git clone https://github.com/mozilla/probe-scraper.git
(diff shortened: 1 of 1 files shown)


jasonthomas:
You could reuse the `$OUTPUT_BUCKET` something like `CACHE_BUCKET="$OUTPUT_BUCKET/cache"`. 

During the `$OUTPUT_BUCKET` sync you would need to exclude `$CACHE_BUCKET` path so it's not removed:

```
    aws s3 sync $OUTPUT_DIR/ s3://$OUTPUT_BUCKET/ \
           --delete \
           --exclude "$CACHE_BUCKET/*" \
           --content-type 'application/json' \
           --cache-control 'max-age=28800' \
           --acl public-read
```

fbertsch:
@jasonthomas if I don't set the acl as `public-read`, will it default to not being public? I think that's the desired behavior here.

fbertsch:
I wasn't clear, I'm asking the opposite - I don't want this data to be public.

Dexterp37:
@fbertsch , right, I think you're correct on the `request_cache` thing. Would you kindly file an issue in the probe scraper repo?

jasonthomas:
> I wasn't clear, I'm asking the opposite - I don't want this data to be public.

Yes it will default to being private/inaccessible via public. You can also explicitly set a object level acl using `--acl private` during sync.

fbertsch:
@Dexterp37 @georgf and I decided to pursue a strategy where we assume we're dealing with git repos (instead of files accessible over http), which makes this change unnecessary. Looking at run-times of probe-scraper I'm not sure it's worth it from a perf standpoint, thoughts?

fbertsch:
Great, we'll look at this again if performance becomes an issue in the future.

PR #296, opened 2018-06-01
Change probe scraper arguments

fbertsch:


jobs/probe_scraper.sh
```diff
@@ -17,7 +17,7 @@ pip install -r requirements.txt
 mkdir $CACHE_DIR $OUTPUT_DIR
 
 # Finally run the scraper.
-python probe_scraper/runner.py --outdir $OUTPUT_DIR --tempdir $CACHE_DIR
+python probe_scraper/runner.py --out-dir $OUTPUT_DIR --temp-dir $CACHE_DIR
```
jasonthomas: I no longer see an option for `--temp-dir` https://github.com/mozilla/probe-scraper/blob/0d8d84394c9249dd175b7213b0095a694d1a7517/probe_scraper/runner.py#L175

Should this be `--cache-dir` instead?

fbertsch:
@jasonthomas good catch, verified the line works now.

PR #297, opened 2018-06-04
Bug 1462810 - Bump default EMR version to 5.13.0 for Spark 2.3.0

acmiyaguchi:
[Bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1462810)

This PR contains a patch for bumping [EMR to 5.13.0](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-5x.html), which is the first version with support for [Spark 2.3.0](https://spark.apache.org/releases/spark-release-2-3-0.html).

@jklukas recently ran the Airflow dag against most of the jobs. I'm going to run the jobs against the new version. There shouldn't be any errors because the Spark release is backwards compatible.

### DIFF
--- a/dags/events_to_amplitude.py
+++ b/dags/events_to_amplitude.py
@@ -35,7 +35,6 @@ events_to_amplitude = EMRSparkOperator(
     job_name="Focus Android Events to Amplitude",
     execution_timeout=timedelta(hours=8),
     instance_count=FOCUS_ANDROID_INSTANCES,
-    release_label="emr-5.13.0",
     env={
         "date": "{{ ds_nodash }}",
         "max_requests": FOCUS_ANDROID_INSTANCES * VCPUS_PER_INSTANCE,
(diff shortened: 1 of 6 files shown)


acmiyaguchi:
I'm running the following jobs:
* [Longitudinal](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-1Z3QUOWGFIY16)
* [MainSummaryView](https://us-west-2.console.aws.amazon.com/elasticmapreduce/home?region=us-west-2#cluster-details:j-1RBTU1SS2R7ST)

acmiyaguchi:
@jezdez Have you experienced any serious issues with EMR 5.13.0? I think if the above two jobs run fine, it should be alright to bump the default tag.

acmiyaguchi:
I've run all the `python_mozetl` tests using Spark 2.3, but I haven't run the entire `main_summary` dag. However, `main_summary` works fine on EMR `5.13`. This shouldn't affect the longitudinal job because its already on `5.13`.

acmiyaguchi:
We should be able to remove the pinned versions without breaking anything. I'm familiar with a few of the pinned tags which were meant to pull in bugfixes from upstream Spark.

PR #301, opened 2018-06-08
Prevent backfill by default on local Airflow instance

acmiyaguchi:
I set the `catchup_by_default` configuration variable to False via the Dockerfile.dev environment. When I tried this last year, it wasn't working for me.

I tested this by rerunning `make build` and  setting the start date of the debug DAG to 3 days ago. When true, it started to run the tasks from the start date. When set false, it only ran the most recent day. I also tried this on `churn`, and it correctly ran only the most recent Wednesday. 

Also, I documented a workaround for Linux users where the user id causes permission errors with the mounted volume.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -31,7 +31,8 @@ ENV AIRFLOW_HOME=/app \
     # AIRFLOW_SMTP_HOST= \
     # AIRFLOW_SMTP_USER= \
     # AIRFLOW_SMTP_PASSWORD= \
-    AIRFLOW_SMTP_FROM=telemetry-alerts@airflow.dev.mozaws.net
+    AIRFLOW_SMTP_FROM=telemetry-alerts@airflow.dev.mozaws.net \
+    AIRFLOW__SCHEDULER__CATCHUP_BY_DEFAULT=False
 
 EXPOSE $PORT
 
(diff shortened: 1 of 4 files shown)


Dockerfile.dev
```diff
@@ -31,7 +31,8 @@ ENV AIRFLOW_HOME=/app \
     # AIRFLOW_SMTP_HOST= \
     # AIRFLOW_SMTP_USER= \
     # AIRFLOW_SMTP_PASSWORD= \
-    AIRFLOW_SMTP_FROM=telemetry-alerts@airflow.dev.mozaws.net
+    AIRFLOW_SMTP_FROM=telemetry-alerts@airflow.dev.mozaws.net \
+    AIRFLOW__SCHEDULER__CATCHUP_BY_DEFAULT=False
```
jklukas: It might be worth adding a comment linking to https://airflow.apache.org/configuration.html above this ENV directive for easy reference of how these env vars are used, especially since we're now using one of the more complex double-underscore overrides. In particular, it's nice to see docs about env vars taking precedence over config values.

dags/debug.py
```diff
@@ -5,7 +5,7 @@
 default_args = {
     'owner': 'nobody@example.com',
     'depends_on_past': False,
-    'start_date': datetime(2099, 5, 31),
+    'start_date': datetime.now() - timedelta(3),
```
jklukas: Is it intentional to include this change in this PR? Do you think there's any danger here for prod?

acmiyaguchi: Yes, the simplest way to verify backfill behavior is to turn this DAG on. 

I don't think there should be any danger on production, this DAG should be turned off there (and state is maintained between deployment). The worst case I can imagine is where all the SleepOperators use up the execution slots for scheduling other things.

PR #302, opened 2018-06-12
Make longitudinal use TBV release process

fbertsch:


jobs/telemetry_batch_view.py
```diff
@@ -43,8 +43,16 @@ def submit_job():
     call_exit_errors(command)
 
 
+def update_metastore(location):
+    cmd = "parquet2hive -ulv 1 --sql {} | beeline -u jdbc:hive2://${HIVE_SERVER}:10000".format(location)
+    call_exit_errors(cmd)
+
+
```
wlach: unnecessary extra newline?

dags/longitudinal.py
```diff
@@ -24,8 +24,14 @@
...
     release_label="emr-5.11.0",
-    env={"date": DS_WEEKLY, "bucket": "{{ task.__class__.private_output_bucket }}"},
-    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/longitudinal_view.sh",
+    env=tbv_envvar(
+        "com.mozilla.telemetry.views.LongitudinalView",
+        {
+            "bucket": "{{ task.__class__.private_output_bucket }}",
+            "to": DS_WEEKLY
+        },
+        metastore_location="s3://telemetry-parquet/longitudinal"),
```
relud: instead of adding a parameter here, you should just use other to pass it:
`other={"METASTORE_LOCATION": "s3://telemetry-parquet/longitudinal"}`

fbertsch: I thought it might be a good idea to include it explicitly, since we'll be making a similar change to all of the other jobs (at some point - it's on the docket for soonish)

PR #307, opened 2018-06-22
Bug 1467860 - Load database creds from S3 for aggregator job

jklukas:
Adapts to the envvar-based creds needed since https://github.com/mozilla/python_mozaggregator/pull/63

The new credentials file has been uploaded to S3 and I've tested that _preparedb() returns successfully with this code on an ATMO cluster.

### DIFF
--- a/jobs/telemetry_aggregator.py
+++ b/jobs/telemetry_aggregator.py
@@ -1,12 +1,30 @@
 #!/mnt/anaconda2/bin/ipython
 
+import json
 import logging
 from os import environ
+from boto.s3.connection import S3Connection
 from pyspark import SparkContext, SparkConf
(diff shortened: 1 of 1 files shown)


jobs/telemetry_aggregator.py
```diff
@@ -1,21 +1,36 @@
...
+for k, v in creds.items():
+    environ[k] = v
+
+# Attempt a database connection now so we can fail fast if credentials are broken.
+_preparedb()
+
+# Send jobs to the spark workers.
 conf = SparkConf().setAppName('telemetry-aggregates')
 sc = SparkContext(conf=conf)
-sc.addPyFile(package_file)
```
jklukas: I'm not sure why this line went away. Investigating.

PR #319, opened 2018-07-02
Add Event Ping Events dataset job

sunahsuh:
Scheduling the event ping events dataset from mozilla/telemetry-streaming#149 (tested locally and everything worked fine, and the urls for telemetry-batch-view jobs seem to be fine)

### DIFF
--- a/dags/utils/tbv.py
+++ b/dags/utils/tbv.py
@@ -2,7 +2,8 @@
 
 from utils.deploy import get_artifact_url
 
-def tbv_envvar(klass, options, branch=None, tag=None, other={}, metastore_location=None):
+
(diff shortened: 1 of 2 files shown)


dags/utils/tbv.py
```diff
@@ -19,8 +20,11 @@ def tbv_envvar(klass, options, branch=None, tag=None, other={}, metastore_locati
 
     :returns: a dictionary that contains properly prefixed class and options
     """
-    slug = "{{ task.__class__.telemetry_batch_view_slug }}"
-    url = get_artifact_url(slug, branch=branch, tag=tag)
+    if artifact_url is not None:
```
jklukas: I'd switch these branches around to be able to avoid `not None`.

PR #321, opened 2018-07-09
[Bug 1473090] Migrate telemetry-airflow to circle 2.0

haroldwoo:


.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
+# These environment variables must be set in CircleCI UI
+#
+# DOCKERHUB_REPO - docker hub repo, format: <username>/<repo>
+# DOCKER_EMAIL   - login info for docker hub
```
jasonthomas: nit: `DOCKER_EMAIL` is no longer used here

PR #323, opened 2018-07-13
Fix #308 - add input/output buckets for engagement ratio

### LINKED ISSUE #308: main_summary.engagement_ratio does not expose input/output bucket
https://github.com/mozilla/telemetry-airflow/blob/1d2cc3b82c40b71e18936b45217abd9861fce967/dags/main_summary.py#L56-L63

This job does not accept the bucket location from the environment, which breaks integration testing.
The engagement ratio job is also not exposed to the mozetl cli, and can't be run using `mozetl_submit`.

acmiyaguchi:
This fixes #308 by setting the correct input and output buckets. This job also depends on the corresponding [mozetl PR](https://github.com/mozilla/python_mozetl/pull/242) to use the `mozetl-submit` script. Unfortunately, the job uses a different location from the `PRIVATE_OUTPUT_BUCKET`, so the conditional is needed to specify the right bucket.

The jinja template uses the `-` sign in front of the variable block in order to strip whitespace in the comment string. There's some documentation on [whitespace control](http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control). I've also included a small snippet that I used to play with the behavior.

```python
from jinja2 import Template
tmpl = """
{% if cond %}
    {{- truthy }} 
{% else %}
    {{- 'a string' }}  
{% endif %}""".strip()
print(Template(tmpl).render({"cond": True, "truthy": "some string"}))

# some string 
```

I made sure that the job rendered correctly on my local machine:

`MOZETL_COMMAND=engagement_ratio MOZETL_ENGAGEMENT_RATIO_INPUT_BUCKET=telemetry-test-bucket MOZETL_ENGAGEMENT_RATIO_OUTPUT_BUCKET=telemetry-test-bucket`

I then changed `DEPLOY_ENVIRONMENT=prod` in `Dockerfile.dev` and got the expected result:
`MOZETL_COMMAND=engagement_ratio MOZETL_ENGAGEMENT_RATIO_INPUT_BUCKET=telemetry-test-bucket MOZETL_ENGAGEMENT_RATIO_OUTPUT_BUCKET=net-mozaws-prod-us-west-2-pipeline-analysis`


dags/main_summary.py
```diff
@@ -57,7 +57,16 @@
...
     instance_count=10,
-    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/engagement_ratio.sh",
+    env=mozetl_envvar("engagement_ratio", {
+        "input_bucket": "{{ task.__class__.private_output_bucket }}",
+        "output_bucket": """
+            {% if task.__class__.deploy_environment == 'dev' %}
+                {{- task.__class__.private_output_bucket }}
+            {% else %}
+                {{- 'net-mozaws-prod-us-west-2-pipeline-analysis' }}
+            {% endif %}""".strip(),
```
relud: why are you using bloth `{{-` and `.strip()` ?

acmiyaguchi:
Since the the template is in a docstring, there's a newline at the beginning. Without the `.strip()` and `{{-`, it renders like this:

image

When I just add in `.strip()`, it keeps the preceeding newline from the line break after the triple quotes:

image

I had to use both of them to render the text without any line breaks.

dags/main_summary.py
acmiyaguchi: Whoops, addressed at a top-level comment.

PR #325, opened 2018-07-17
[Bug 1455730] Upgrade airflow to 1.9

haroldwoo:
- Removes mozlog.diff patch since new version will require new diff file and we don't ship to logging2.0 anyways
- Removes pip install of 9.0.3
- Upgrade airflow to 1.9.0
- Update readme to include fernet key fix for initdb


haroldwoo:
That environment variable takes precedence over the value in airflow.cfg.
https://airflow.apache.org/configuration.html

PR #326, opened 2018-07-18
Bug 1476067 - Set sane defaults for main_summary DAG when the deploy environment is dev

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1476067)


acmiyaguchi:
It looks like this mostly works, aside from the `experiment_error_aggregates`, which is actually a job that doesn't depend on `main_summary`.

acmiyaguchi:
@sunahsuh This is still a work in progress, but do you have opinions on setting the instance count based on the environment?

sunahsuh:
I'm actually interested in seeing support for dev environments deeper in the code -- for instance, maybe we can add a param to EMRSparkOperator to turn itself into a no-op task in dev, and a param to specify the number of dev instances. We could also add an argument to `tbv_envvar` to pass in dev-only flags. I think those changes would let you remove all the conditionals from the DAG file, and would be useful if we decided we want to do this for other DAGs in the future.

acmiyaguchi:
Thanks for the feedback @sunahsuh, I'll try to move as much of the development logic into the operator. I'm I'll make sure to update the documentation on how to run the tests locally too.

acmiyaguchi:
image

I made some changes to the EMR operator, but there are still a couple of places the dev/prod logic needs to exist in the DAG itself. It looks like most of the jobs complete relatively quickly. 

@sunahsuh do you know of a way to make the `Experiments Aggregates View` faster? It seems like this particular sub-tree is a bottleneck for completing the job.

dags/main_summary.py
```diff
@@ -26,11 +41,17 @@
     job_name="Main Summary View",
     execution_timeout=timedelta(hours=14),
     instance_count=40,
-    env=tbv_envvar("com.mozilla.telemetry.views.MainSummaryView", {
-        "from": "{{ ds_nodash }}",
-        "to": "{{ ds_nodash }}",
-        "schema-report-location": "s3://{{ task.__class__.private_output_bucket }}/schema/main_summary/submission_date_s3={{ ds_nodash }}",
-        "bucket": "{{ task.__class__.private_output_bucket }}"}),
+    env=tbv_envvar("com.mozilla.telemetry.views.MainSummaryView", cond_env(
```
sunahsuh: Wouldn't it be a bit cleaner to add a `dev` param to `tbv_envvar` and have that function check the deploy environment?

dags/operators/emr_spark_operator.py
```diff
@@ -68,23 +68,44 @@ def on_kill(self):
...
+        :param instance_count: The number of instances to use in production
+        :param dev_instance_count: The number of instances to use in development
+        :param disable_on_dev: Turn the job into a no-op if run in development
+        :param release_label: The EMR release label (i.e. 'emr-5.13.0')
+        :param output_visibility: 'public' or 'private', specifying the default
+                                    output bucket for data
+        :param env: A dictionary of environment variables to pass during runtime
+        :param dev_env: Additional environment variables to pass in development
+        :param arguments: Passed to `airflow.sh`
+        """
```
sunahsuh: \o/

acmiyaguchi:
@sunahsuh this should look up to standard now:

* modified tbv_envvar and mozetl_envvar to take dev options
* experiments error aggregates now uses tbv_envvar
* engagement ratio uses the dev_options parameter now

It should take roughly 1-2 hours to run the entire DAG on c3.2xlarge. The retention and experiment error aggregates were bumped to 3 nodes each so they run within an hour. 

image

I'm going to merge this in tomorrow morning, I don't think there are any serious issues.

acmiyaguchi:
I wrote a script to parse through the `telemetry-test-bucket` for out data.

```
$ aws s3 ls s3://telemetry-test-bucket/ --recursive | grep "^2018-08-01" > filelisting_20180801.txt
$ python aggregate_airflow.py filelisting_20180801.txt 
                                                                 size
name                                       date       hour           
addons                                     2018-08-01 15     24633416
addons/agg                                 2018-08-01 15      2828340
client_count_daily                         2018-08-01 15      4466433
clients_daily                              2018-08-01 15     18063292
events                                     2018-08-01 15      1678473
experiment_error_aggregates                2018-08-01 15      1731098
experiments                                2018-08-01 15    173849533
experiments_aggregates                     2018-08-01 15        59715
harter/searchdb                            2018-08-01 15      1101510
main_summary                               2018-08-01 15    209457999
mreid/maudau/engagement_ratio.20180801.csv 2018-08-01 15        36832
mreid/maudau/engagement_ratio.csv          2018-08-01 15        36832
search_clients_daily                       2018-08-01 15      6765948
txp_mau_dau_simple                         2018-08-01 15         1105
```

There a few things that are missing from this report. The retention dataset is missing because I manually terminated the job. There is also only one line for clients_daily, which is because I omit the version number as an aggregation key.

This verification is slow, but it shows the relevant information for this specific DAG run.

PR #329, opened 2018-07-30
Bug 1479580 - Log which JAR is being used in telemetry-batch-view

fbertsch:
Previously, there were no logs about which artifact
was being used. That made it hard to debug problems.
This change will make it obvious which JAR is being
used to run a job.

### DIFF
--- a/jobs/telemetry_batch_view.py
+++ b/jobs/telemetry_batch_view.py
@@ -4,7 +4,7 @@ import requests
 from os import chdir
 from os import environ
 from subprocess import call, PIPE, Popen
-
+from urlparse import urlparse
 
 artifact_file = "artifact.jar"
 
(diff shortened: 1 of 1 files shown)


jobs/telemetry_batch_view.py
```diff
@@ -22,6 +22,42 @@ def retrieve_jar():
     if jar_url is None:
         exit(1)
 
+
+    print("Retrieving JAR: {}".format(jar_url))
```
jklukas: Do we have a habit of `print`ing in this project? Should we instead call `logging.info`?

fbertsch: Currently, yes. I'd be fine with switching to `logging.info`, but I would prefer we tackle that in a follow-up.

jobs/telemetry_batch_view.py
```diff
@@ -22,6 +22,42 @@ def retrieve_jar():
...
+
+    if response.status_code != 404:
+        stripped = response.content.strip()
+        if len(stripped.split("\n")) == 1:
+            # Handle historical version
+            uri_query = stripped
+            build_url = "Build URL not available"
+        else:
+            # Output both query string and build URL
+            uri_query, build_url = stripped.split("\n")
```
jklukas: This would be slightly cleaner and safer using `.partition()`:

```
uri_query, _, build_url = response.content.strip().partition('\n')
if build_url is None:
  # Handle historical version
  build_url = "Build URL not available"
```

fbertsch: Good recommendation!

PR #333, opened 2018-08-06
Bug 1469388 - Add job for Desktop DAU

relud:


relud:
I'll set a calendar reminder to check up on it in a month and reduce the number of nodes if this turns out to be excessive.

PR #334, opened 2018-08-07
Update instance count and timeout for desktop_dau

relud:
and add missing protocol prefix


dags/main_summary.py
```diff
@@ -318,11 +318,11 @@
...
     owner="relud@mozilla.com",
     email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
-    execution_timeout=timedelta(hours=5),
-    instance_count=10,
+    execution_timeout=timedelta(hours=1),
+    instance_count=1,
     env=tbv_envvar("com.mozilla.telemetry.views.dau.DesktopDauView", {
         "to": "{{ ds_nodash }}",
-        "bucket": "{{ task.__class__.private_output_bucket }}",
+        "bucket": "s3://{{ task.__class__.private_output_bucket }}",
```
sunahsuh: shouldn't we change this bit in the job code instead? bucket is misleadingly named if it needs a protocol..

relud: the job's test depends on being able to specify a non-s3 location.

also bucket could mean gcs or s3, and this allows it to support both.

How about I add a `--bucket-protocol` option to the job code?

PR #337, opened 2018-08-15
Bug 1481274 - Add tbv and mozetl compatible Databricks operator to Airflow

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1481274)

This adds a `MozDatabricksSubmitRunOperator` as a plugin. The interface should be a drop-in replacement for the `EMRSparkOperator` if the job uses the `tbv_envvar` or `mozetl_envvar` job wrappers. My notes from the Databricks POC can be found [here](https://gist.github.com/acmiyaguchi/205519fb1d7c8f5dcb61fb92b6236b76), which have been useful for adding functionality.

In addition to faster startup times, the autoscaling functionality might also be useful in practice. The Spark history and executor logs are also properly captured in this environment. 

The connection settings should be set in the Airflow UI under `Admin > Connections  > databricks_default`:

```
Host: https://dbc-caf9527b-e073.cloud.databricks.com
Extras: {"token": "<GENERATED_TOKEN>"}
```

These can also be passed through the environment or `Dockerfile.dev`: `AIRFLOW_CONN_DATABRICKS_DEFAULT_HOST=...`


acmiyaguchi:
I'm running into a bug where clusters keep spinning up (~1 per second) when run via `make run COMMAND="test main_summary main_summary 20180801"`. I've worked around this by running `docker-compose up` once, and then using `docker-compose exec` afterwards. 

In one window, launch airflow:
```
$ make up
```

In another one, run this script:
```
#!/bin/bash

HOST='https://dbc-caf9527b-e073.cloud.databricks.com'
EXTRA='{"token":"<TOKEN>"}'

docker-compose exec web \
    airflow connections \
        --delete \
        --conn_id databricks_default

docker-compose exec web\
    airflow connections \
        --add \
        --conn_id databricks_default \
        --conn_type databricks \
        --conn_host ${HOST} \
        --conn_extra ${EXTRA}

docker-compose exec web \
    airflow test main_summary main_summary 20180801
```

It looks like the whole process is roughly 10 minutes.
image

The cluster spin-up is 3 minutes, and the job takes another 7.

acmiyaguchi:
@sunahsuh This is mostly ready for review, but I need to address a few issues related to deployment. This operator should work as a drop-in replacement, I've tested it with the two jobs in the most recent commit. Most of the jobs we currently have fail because they include the explicit call to `spark.stop()` which causes a failure via the API. 

@haroldwoo Do you have any insights into how the connection string for Databricks would be configured in production? In particular, I want to be able to reference the IAM role for spinning up clusters with access to the telemetry-landfill bucket. Should this be exposed to the environment or set up during airflow init?

acmiyaguchi:
FYI, the `[testing]` commit should not be merged in and will be dropped at the end of review. This PR is in  a state where it can be tested locally.

plugins/moz_databricks.py
```diff
@@ -0,0 +1,171 @@
...
+        if enable_autoscale:
+            new_cluster["autoscale"] = {
+                "min_workers": min_workers,
+                "max_workers": max_workers,
+            }
+        else:
+            new_cluster["num_workers"] = min_workers
+
+        # Parse the environment variables to bootstrap the tbv/mozetl workflow
+        if env.get("TBV_CLASS"):
```
sunahsuh: Suggestion: instead of this if clause, use an abstract base class with the shared stuff and create two child classes, one each for jobs that use tbv_envvar and mozetl to set the bits in each clause

plugins/moz_databricks.py
```diff
@@ -0,0 +1,171 @@
...
+            jar_task = {
+                "main_class_name": env["TBV_CLASS"],
+                # Reconstruct TBV parameters from the environment, scallop does
+                # not support reading arguments in this form
+                "parameters": sum([
+                    ["--{}".format(key[4:].replace("_", "-")), value]
+                    for key, value in env.items()
+                    if key.startswith("TBV_") and key != "TBV_CLASS"
```
sunahsuh: I understand the impetus to try to make this class as backwards compatible as possible, but I think we should consider passing parameters through without tbv_envvar if we can avoid this re-munging things we just munged earlier. I'm not sure this is worth the interface compatibility

acmiyaguchi: I don't think this operator is worth the interface incompatibility because I think the environment is the right level to think about the job calling convention. For example, if one day we migrated our jobs to submit via a Docker master, the easiest way to pass arguments would be through the environment. The `tbv_envvar` and `mozetl_envvar` interface with the corresponding runners have been working well, so far. 

It's a bit convoluted to do this extra work when we could pass in the raw string, but that's the price for choosing environment variables for passing context between run-times. My preferred solution to this would be [environment variable support like click](http://click.pocoo.org/5/options/#values-from-environment-variables).

Also since the interface is compatible, I also think it would be cool to monkey-patch EMRSparkOperator with the MozDatabricksRunSubmitOperator later.

plugins/moz_databricks.py
acmiyaguchi: I'm going to hold off on this one. I was actually thinking it might be interesting to think about a base class for Mozilla-specific information we would want to have in our jobs. The EMRSparkOperator should probably be moved into the plugins directory, and the common templating variables and constructor values could be moved into a common class.

acmiyaguchi:
Thanks for the review @sunahsuh :+1: . There are still a few things that need to be figured out to production and evaluate on a larger scale ([Bug 1484337](https://bugzilla.mozilla.org/show_bug.cgi?id=1484337)), but this will probably be touched a few more times.

I'm going to test this one more time, remove the `[testing]` commit, and merge this in after the weekend.

PR #340, opened 2018-08-22
Bug 1482208 - Specify environment and user in Airflow EMR cluster names

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1482208)

This adds a prefix to EMR cluster names for easier visual filtering.

In production, the behavior will remain the same. On stage and dev, the job name will be prefixed with the environment name:

`[dev] Main Summary View`

If `DEV_USERNAME` is also set in the Dockerfile or exported during `make run`, then it will also be appended to the name:

`[dev][amiyaguchi] Main Summary View`

EDIT:
I had failures related to the bugzilla_database job, related to my specific setup. Everything in this PR should look good.


acmiyaguchi:
It turns out that the EMR search is actually regex powered, which makes searching for the tags kind of annoying:

```
\[dev\]\[amiyaguchi\]
```

The format described in the bug is more regex friendly:

`dev-amiyaguchi: Main Summary View`

PR #343, opened 2018-09-06
Use master for focus-android events to amplitude

fbertsch:


fbertsch:
@sunahsuh this did get tested on dev, and everything seemed to be in order.

PR #349, opened 2018-09-12
Aggregate release telemetry

fbertsch:
I do have a concern about overloading the database, but we use a read-replica so we can handle that later, if it happens.

### DIFF
--- a/dags/telemetry_aggregates.py
+++ b/dags/telemetry_aggregates.py
@@ -16,8 +16,8 @@ default_args = {
 dag = DAG('telemetry_aggregates', default_args=default_args, schedule_interval='@daily')
 
 prerelease_telemetry_aggregate_view = EMRSparkOperator(
-    task_id = "telemetry_aggregate_view",
-    job_name = "Telemetry Aggregate View - Prerelease",
+    task_id = "prerelease_telemetry_aggregate_view",
+    job_name = "Prerelease Telemetry Aggregate View",
     instance_count = 10,
     execution_timeout=timedelta(hours=12),
     env = {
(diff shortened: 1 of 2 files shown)


dags/telemetry_aggregates.py
```diff
@@ -15,11 +15,26 @@
...
-    env = {"date": "{{ ds_nodash }}"},
+    env = {
+      "date": "{{ ds_nodash }}",
+      "channels": "nightly, aurora, beta"
+    },
+    uri = "https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/run_telemetry_aggregator.sh",
+    dag = dag)
+
+release_telemetry_aggregate_view = EMRSparkOperator(
+    task_id = "telemetry_aggregate_view",
```
mreid-moz: `task_id` should be unique here, maybe `"telemetry_aggregate_view_release"`?

dags/telemetry_aggregates.py
```diff
@@ -15,11 +15,26 @@
...
+    job_name = "Telemetry Aggregate View - Release",
+    instance_count = 40,
+    execution_timeout=timedelta(hours=12),
+    env = {
+      "date": "{{ ds_nodash }}",
+      "channels": "release"
```
mreid-moz: Should we include `esr` too?

fbertsch: ESR isn't in the spec, and requires some changes to the aggregator. Let's confirm that's something people want and tackle that separately.

mreid-moz: Sounds good to me.

mreid-moz:
> I do have a concern about overloading the database, but we use a read-replica so we can handle that later, if it happens.

We could also set release aggregates as being "downstream" from pre-release if that becomes an issue.

fbertsch:
@mreid-moz that would probably help, but I'm more concerned by the 40 machines from that one job simultaneously committing. I'll probably have to limit the number of parallel tasks at that stage if it becomes an issue.

PR #351, opened 2018-09-20
Add telemetry aggregates parquet job (bug 1475358)

robhudson:


dags/telemetry_aggregates_parquet.py
```diff
@@ -0,0 +1,28 @@
...
+    'email': ['telemetry-alerts@mozilla.com', 'robhudson@mozilla.com', 'frank@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 3,
+    'retry_delay': timedelta(minutes=30),
+}
+
+dag = DAG('telemetry_aggregates_parquet', default_args=default_args, schedule_interval='@daily')
+
+telemetry_aggregate_parquet_view = EMRSparkOperator(
```
fbertsch: Want to run this on databricks?

dags/telemetry_aggregates_parquet.py
```diff
@@ -0,0 +1,28 @@
...
+    task_id="telemetry_aggregate_parquet_view",
+    job_name="Telemetry Aggregate Parquet View",
+    instance_count=40,
```
fbertsch: Does this really need 40 nodes?

dags/telemetry_aggregates_parquet.py
robhudson: I'm not sure? Should we? How does one do that?

dags/telemetry_aggregates_parquet.py
robhudson: I copied what the telemetry aggregates needs. It does seem high but I don't know how to judge how many nodes is really needed.

fbertsch: Those are the release aggregates. I would imagine this doesn't need more than 5 nodes, can you confirm?

dags/telemetry_aggregates_parquet.py
fbertsch: The Longitudinal job does that, check how it's done there.

dags/telemetry_aggregates_parquet.py
robhudson: When I was testing it I was using a 5 node cluster on ATMO and less a few times and it ran fine.

dags/telemetry_aggregates_parquet.py
robhudson: Cool, I'll check it out and switch to that.

acmiyaguchi: Only jobs that run via `mozetl_envvar` and `tbv_envvar` and the corresponding runners are supported via Databricks now. Custom scripts won't work without modifying the operator. Sorry for the lack of documentation on this behavior, it should be a bit more explicit.

robhudson:
@fbertsch This has now been tested and updated and rebased.

PR #355, opened 2018-10-08
Fix unicode issues in airflow emails

sunahsuh:
Unicode 😭

### DIFF
--- a/dags/operators/emr_spark_operator.py
+++ b/dags/operators/emr_spark_operator.py
@@ -246,12 +246,13 @@ class EMRSparkOperator(BaseOperator):
                 spark_log_location = self.get_spark_log_location()
 
                 raise AirflowException(
-                    'Spark job {} terminated with errors: {} - {}<br>'
+                    u'Spark job {} terminated with errors: {} - {}<br>'
                     'Cluster Stderr: {}<br>'
                     'Cluster Stdout: {}<br>'
                     'Spark Driver Log Location: {}'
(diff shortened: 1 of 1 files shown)


dags/operators/emr_spark_operator.py
```diff
@@ -246,12 +246,13 @@ def describe_cluster():
...
 
                 raise AirflowException(
-                    'Spark job {} terminated with errors: {} - {}<br>'
+                    u'Spark job {} terminated with errors: {} - {}<br>'
                     'Cluster Stderr: {}<br>'
                     'Cluster Stdout: {}<br>'
                     'Spark Driver Log Location: {}'
                     .format(self.job_name, reason_code, reason_message,
                             step_logs['stderr'], step_logs['stdout'], spark_log_location)
+                    .format('ascii', 'replace')
```
acmiyaguchi: Should this be `encode` instead? I tried the following in a 2.7 session:

```python
>>> u'hello world {}'.format(b"\xF0\x9F\x98\x81".decode('utf')).format('ascii', 'replace')
u'hello world \U0001f601'
>>> u'hello world {}'.format(b"\xF0\x9F\x98\x81".decode('utf')).encode('ascii', 'replace')
'hello world ?'
```

sunahsuh: And this, my friend, is why we have code review (and why we should have a testable operator but..)

PR #362, opened 2018-11-09
Bug 1506252 - Register socorro crash reports with Statuspage

acmiyaguchi:
[buglink](https://bugzilla.mozilla.org/show_bug.cgi?id=1506252)

This renames `socorro_import` to `crash_reports_parquet`. The dag name and the resulting dataset name are still `socorro_import`. The registered name will be "Socorro Crash Reports Parquet" on statuspage.

[Bug 1498228](https://bugzilla.mozilla.org/show_bug.cgi?id=1498228) is blocking.

### DIFF
--- a/dags/socorro_import.py
+++ b/dags/socorro_import.py
@@ -31,6 +31,6 @@ crash_report_parquet = EMRSparkOperator(
 
 register_status(
     crash_report_parquet,
-    "Socorro Crash Reports Parquet",
+    crash_report_parquet.job_name,
     "Convert processed crash reports into parquet for analysis",
 )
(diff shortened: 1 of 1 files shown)


dags/socorro_import.py
```diff
@@ -1,26 +1,36 @@
...
     env={"date": "{{ ds_nodash }}"},
     uri="https://raw.githubusercontent.com/mozilla-services/data-pipeline/master/reports/socorro_import/ImportCrashData.ipynb",
     output_visibility="public",
-    dag=dag)
+    dag=dag,
+)
+
+register_status(
+    crash_report_parquet,
+    "Socorro Crash Reports Parquet",
```
mreid-moz: nit: Can we use `crash_report_parquet.job_name` here to avoid repeating this?

PR #363, opened 2018-11-15
Bug 1498231 - Register MainSummaryView with Statuspage

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1498231)

This registers MainSummaryView failures with statuspage. If there does happen to be a failure, then someone will need to go into the statuspage management page and turn it back to operational after fixing it. This creates a component for the dataset, so incidents can be filed even if the node in the DAG doesn't trigger.

The second half of the bug will end up being a query from redash using email to alert on the dataset status outside of airflow.


acmiyaguchi:
Woops, I should have looked at the calendar before reviewing request.

PR #367, opened 2018-11-29
Switch devtools prerelease amplitude job to new config schema name

sunahsuh:
PR Part 1: update the current job to specify prerelease-only

### DIFF
--- a/dags/events_to_amplitude.py
+++ b/dags/events_to_amplitude.py
@@ -6,8 +6,6 @@ from utils.mozetl import mozetl_envvar
 from utils.deploy import get_artifact_url
 
 FOCUS_ANDROID_INSTANCES = 10
-# Currently devtools is shipping only nightly events while they debug some issues,
-# so 2 nodes is plenty
 DEVTOOLS_INSTANCES = 10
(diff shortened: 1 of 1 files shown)


dags/events_to_amplitude.py
```diff
@@ -51,9 +49,9 @@ def key_file(project):
...
     dag=dag)
 
-devtools_events_to_amplitude = EMRSparkOperator(
-    task_id="devtools_events_to_amplitude",
-    job_name="DevTools Events to Amplitude",
+devtools_prerelease_events_to_amplitude = EMRSparkOperator(
+    task_id="devtools_prerelease_events_to_amplitude",
+    job_name="DevTools Prerelease Events to Amplitude",
     execution_timeout=timedelta(hours=8),
     instance_count=DEVTOOLS_INSTANCES,
```
fbertsch: nit: `PRERELEASE_DEVTOOLS_INSTANCES`?

sunahsuh: I was going to use the same number I can get the job to run fast enough on release :D I'll change it in the follow-up PR if that's not the case

PR #369, opened 2018-12-05
Bug 1505583: Use tbv_envvar for devtools release e2a job

sunahsuh:
Adding the devtools release events to amplitude job sampled at 50% to the dag -- the other events to amplitude jobs use a custom job script but I've made the changes needed to use the generic tbv job script instead. I'll start with the new release job using the generic runner and move the rest of the jobs next week after this has run for a few days (and when it's not all-hands :))

### DIFF
--- a/jobs/telemetry_batch_view.py
+++ b/jobs/telemetry_batch_view.py
@@ -15,7 +15,7 @@ def call_exit_errors(command):
     print("+ {}".format(" ".join(command)))
     rc = call(command, env=environ.copy())
     if rc > 0:
-        exit(rc)
+       exit(rc)
 
 
 def retrieve_jar():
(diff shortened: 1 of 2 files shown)


jobs/telemetry_batch_view.py
```diff
@@ -92,6 +110,9 @@ def update_metastore(location, hive_server):
 if environ.get("DO_RETRIEVE", "True") == "True":
     retrieve_jar()
 
+if environ.get("DO_EVENTS_TO_AMPLITUDE_SETUP") == "True":
+    events_to_amplitude_setup(environ.get("TBV_config_file_path"), environ.get("KEY_BUCKET"), environ.get("KEY_PATH"))
```
relud: I think to match the style of `retrieve_jar()` above, this should be a function with no arguments, and the values should be extracted from `environ` in the body of `events_to_amplitude_setup`

sunahsuh:
> Non-blocking, I feel like the changes to telemetry_batch_view.py don't really belong there. I would rather see a wrapper script that downloads and executes telemetry_batch_view.py, but I'm not sure that's feasible.

So, `telemetry_batch_view.py` is already a script that's already downloaded and executed, amusingly (by [airflow.sh](https://github.com/mozilla/telemetry-airflow/blob/master/airflow.sh)). I do agree this is getting rather unwieldy and I think if we were planning to stay with EMRSparkOperator for the long-term there would be some avenues to explore, but I'm not sure it's worth it at this junction.

PR #374, opened 2018-12-18
[Bug 1515145] Reduce timeout of main summary and add pagerduty email

haroldwoo:


dags/main_summary.py
```diff
@@ -13,7 +13,7 @@
     'owner': 'frank@mozilla.com',
     'depends_on_past': False,
     'start_date': datetime(2015, 11, 3),
-    'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
+    'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com', 'main_summary_dataset@moz-svc-ops.pagerduty.com'],
```
fbertsch: This will email us on any failure/retry on any of the tasks in the DAG. We should add it explicitly to the main_summary task only.

haroldwoo: default args is only used in main_summary

haroldwoo: would you rather it be called main_summary_args

fbertsch: default args is used by the `main_summary` DAG, which contains a `main_summary` task. We only want to add it to the latter

haroldwoo:
ah ok

PR #375, opened 2018-12-19
Move main summary to databricks

fbertsch:


fbertsch:
Currently being tested at https://dbc-caf9527b-e073.cloud.databricks.com#job/864/run/1

dags/main_summary.py
```diff
@@ -49,24 +49,30 @@
...
+    ebs_volume_size=250,
     env=tbv_envvar("com.mozilla.telemetry.views.MainSummaryView",
         options={
             "from": "{{ ds_nodash }}",
             "to": "{{ ds_nodash }}",
             "schema-report-location": "s3://{{ task.__class__.private_output_bucket }}/schema/main_summary/submission_date_s3={{ ds_nodash }}",
-            "bucket": "{{ task.__class__.private_output_bucket }}"
+            "bucket": "{{ task.__class__.private_output_bucket }}",
+            "read-mode": "aligned",
+            "input-partition-multiplier": "400"
```
jklukas: Can you explain what behavior you expect out of these changes to read-mode and input-partition-multiplier? Are we expecting small dataset behavior due to the high input-partition-multiplier?

jklukas:
Should we hold off on merging until we confirm that the test run is successful?

dags/main_summary.py
fbertsch: It's actually the opposite - we expect a behavior change when we switch to autoscaling because the cluster size is initially small, which means that the parallelism is low. Forcing the system to use an exact amount of parallelism will cause scale-up of up to our max number of nodes (but possibly less, hence allowing autoscaling at all).

fbertsch:
@jklukas considering it's already failing, and that the test is only on nightly, my vote is to merge it in (especially considering this models the MSWAH, which has been succeeding this whole time).

PR #390, opened 2018-12-27
moved taar_locale job into the main_summary DAG from longitudinal

crankycoder:
This is a simple move of the taar_locale job onto the main_summary DAG under the clients_daily_v6 job.


crankycoder:
@acmiyaguchi r?

PR #396, opened 2019-01-03
Quick dag to check for main summary delays

haroldwoo:


acmiyaguchi:
It seems like the [ExternalTaskSensor](https://airflow.incubator.apache.org/code.html?highlight=external_task_sensor#airflow.sensors.external_task_sensor.ExternalTaskSensor) might be appropriate for checking the task state.

I also did a similar thing in #395. I check for `_SUCCESS` to be written out to S3 using a sensor that polls every 30 minutes for a timeout of 8 hours, and I set the status to a `partial_outage` on failure.

Issue #397, opened 2019-01-04
`register_status` does not create a new component until first failure


acmiyaguchi:
https://github.com/mozilla/telemetry-airflow/blob/a74f4e7f8b1f7dc4dad8d24f83ce4faa98ae13cb/dags/utils/status.py#L4-L7

This function should register a data-set, regardless of it's execution status.

PR #398, opened 2019-01-04
Fix `register_status` to update statuspage

acmiyaguchi:
Currently, if a dataset is registered using `register_status`, the component is only added after it's first failure. This fixes issue #397 by adding another operator that runs `get_or_create` after a dataset is generated.

### DIFF
--- a/dags/utils/status.py
+++ b/dags/utils/status.py
@@ -1,3 +1,6 @@
+import logging
+
+from airflow.operators.python_operator import PythonOperator
 from airflow.operators.dataset_status import DatasetStatusOperator
 from airflow.hooks.dataset_status import DatasetStatusHook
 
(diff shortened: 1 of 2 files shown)


dags/utils/status.py
```diff
@@ -1,4 +1,8 @@
+import logging
+
+from airflow.operators.python_operator import PythonOperator
 from airflow.operators.dataset_status import DatasetStatusOperator
+from airflow.hooks.dataset_status import DatasetStatusHook
```
haroldwoo: tests are failing with ImportError: No module named dataset_status

acmiyaguchi: It looks like my unittests are a bit brittle in this section, I'll give it a look over and re-tag you once I figure out how to untangle this.

acmiyaguchi: re: this is part of the unittest that I fixed where imports are relative to the root path. Airflow edits the sys import paths for the plugins so they look consistent.

tests/test_dag_utils_status.py
```diff
@@ -105,6 +111,7 @@ def exception():
 
     expected = {
         "test": State.FAILED,
+        "test_register": State.NONE,
```
acmiyaguchi: This is actually not ideal because the job should be registered regardless on the success state of the parent. In practice, this is fine because on failure, the StatuspageOperator will execute and update the page.

haroldwoo: put this in a comment?

PR #400, opened 2019-01-07
Allow configuring driver node instance type in Databricks operator

akkomar:


plugins/moz_databricks.py
```diff
@@ -161,7 +167,7 @@ def __init__(self, job_name, env, instance_count,
                 "parameters": [env["MOZETL_COMMAND"]]
             }
 
-            # Proper pip dependencies in Databriks is only supported via pypi.
```
acmiyaguchi: :+1:

plugins/moz_databricks.py
```diff
@@ -61,7 +62,9 @@ def __init__(self, job_name, env, instance_count,
         :param release_label: Databricks Runtime versions,
             run `databricks clusters spark-versions` for possible values.
         :param iam_role: An Amazon Resource Name (ARN) specifying an iam role
-        :param instance_type: An EMR instance type
+        :param instance_type: An EC2 instance type (worker nodes and driver, if noc specified)
```
acmiyaguchi:
```suggestion
        :param instance_type: An EC2 instance type (worker nodes and driver, if not specified)
```

PR #406, opened 2019-01-08
Add an operator to check for _SUCCESS files

### LINKED ISSUE #404: S3FSCheckSuccess fails when __HIVE_DEFAULT_PARTITION__ exists
https://github.com/mozilla/telemetry-airflow/blob/9c14df6a2fc7cb3931226e5a6a7dfa67c2049433/plugins/s3fs_check_success.py#L9-L19

This fails when there is a `__HIVE_DEFAULT_PARTITION__` in the dataset. 

```
$aws s3 ls s3://telemetry-parquet/main_summary/v4/submission_date_s3=20190105/ --recursive | grep _SUCCESS | wc -l

101
``` 

This should be modified so the number of partitions is the minimum required `_SUCCESS` partitions to be written, or an explicit list of partitions.

acmiyaguchi:
This PR adds an operator version of a sensor written in #395. This works in the same way as the `S3FSCheckSuccessSensor`, except it is not a long lived job.

I noticed a few things when watching the sensor last night.

* The sensor worked as expected during a catch-up backfill when other jobs were not being run
* The sensor would fail when it ran into unexpected partitions (the `__HIVE_DEFAULT_PARTITION__` of #404)
* The sensor timed out during the run for `2019-01-07`. The logs show that the poke method was never called during the lifetime of the sensor.
  - This might be caused by the low concurrency limit, where the pokes are treated as a low-priority and is never given a chance to run.

In light of this, I've disabled the `dataset_alerts` DAG. Instead of using a long-lived sensor, I think a better solution in our current Airflow configuration is to run an operator once (with a few retries), when the dataset should definitely be written to disk. This way, the scheduler should pick up the task to be run once another task has finished. This reuses most of the code from the last PR.

There are a couple of small things in this PR too:

* Adding a sensor to the plugins imports `snakebite`, breaking python 3 support in pre-1.10.1. I've disabled py36 tests in tox. See https://issues.apache.org/jira/browse/AIRFLOW-2474. 
* I've fixed #404 so the expected number of partitions is a lower bound.

### DIFF
--- a/dags/dataset_alerts.py
+++ b/dags/dataset_alerts.py
@@ -1,6 +1,6 @@
 from airflow import DAG
 from datetime import datetime, timedelta
-from airflow.operators.s3fs_check_success import S3FSCheckSuccessSensor
+from airflow.sensors.s3fs_check_success import S3FSCheckSuccessSensor
 from airflow.operators.dataset_status import DatasetStatusOperator
 
 default_args = {
--- a/plugins/s3fs_check_success.py
+++ b/plugins/s3fs_check_success.py
@@ -16,7 +16,7 @@ def check_s3fs_success(bucket, prefix, num_partitions):
     objects = s3.Bucket(bucket).objects.filter(Prefix=prefix)
     success = set([obj.key for obj in objects if "_SUCCESS" in obj.key])
     logging.info("Found {n_success} files".format(n_success=len(success)))
-    return len(success) == num_partitions
+    return len(success) >= num_partitions
 
 
 class S3FSCheckSuccessOperator(BaseOperator):
--- a/plugins/s3fs_check_success.py
+++ b/plugins/s3fs_check_success.py
@@ -2,7 +2,7 @@ import logging
 
 import boto3
 from airflow.exceptions import AirflowException
-from airflow.operators import BaseSensorOperator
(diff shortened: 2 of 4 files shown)


acmiyaguchi:
The third observation is only partially correct, there were a total of 3 tries that happened during the run on `2018-01-07`. https://r3e51de5feeeb5619-tp.appspot.com/admin/airflow/log?task_id=check_main_summary&dag_id=dataset_alerts&execution_date=2019-01-07T01:00:00

1. This was the initial run which failed to start executing. This was cleared around 7:00 UTC.
2. This log shows the pokes correctly running at 30 minute intervals. However this didn't succeed because there were actually 101 partitions.
  - `aws s3 ls s3://telemetry-parquet/main_summary/v4/submission_date_s3=20190107/ --recursive | grep _SUCCESS | wc -l`
3. This log was created at 13:00 UTC overlapping with run 2. Not sure why this exists.

The concurrency limit may actually not be an issue here, but something strange did happen on the run last night. 

The concurr

acmiyaguchi:
I'm going to enable the `dataset_alerts` dag again one this is merged in. If there are issues with the sensors for mysterious reasons, I'm going to swap them out with the operator instead since it's simpler.

PR #407, opened 2019-01-08
Remove experiment aggregates job 

robhudson:


robhudson:
Updated after the response in the bug.

PR #416, opened 2019-01-23
Increase number of api retries for main summary

sunahsuh:
Quick bandaid fix for the databricks API issues -- sets the retry limit to 20, which means we'll keep retrying every 30 seconds for 10 minutes. DatabricksSubmitRunOperator doesn't implement exponential backoff, as far as I can tell :/ Possible contribution upstream?

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -72,6 +72,7 @@ main_summary = MozDatabricksSubmitRunOperator(
         dev_options={
             "channel": "nightly",   # run on smaller nightly data rather than release
         }),
+    databricks_retry_limit=20, # at a default polling rate of once every 30 seconds, this is 10 min of retries
     dag=dag)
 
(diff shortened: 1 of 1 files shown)


sunahsuh:
Oh no, looks like it actually just retries with no wait between the tries at all 🙀 https://github.com/apache/airflow/blob/v1-9-stable/airflow/contrib/hooks/databricks_hook.py#L115-L137

acmiyaguchi:
I believe this is the patch we're interested in, which sleeps between retries for errors like `TEMPORARILY_UNAVAILABLE`. 

https://issues.apache.org/jira/browse/AIRFLOW-2709

We'll need the operator from v1.10.1 in order to get these changes.

sunahsuh:
Closing in favor of remediation items from the postmortem

PR #419, opened 2019-01-24
Backport Databricks hook and operator from apache/airflow:v1-10-stable

acmiyaguchi:
This backports patches made to the Databricks hook and operator as a separately maintained plugin. We should be able to access the operator like any other plugin. We can modify the `MozDatabricksRunSubmitOperator` to use the plugin instead of the one from within the tree.

The backport process was a 3 part process (broken down by commit).

1. Copy the relevant files from https://github.com/apache/airflow/tree/v1-10-stable
2. Fix the imports and mocks to reference the plugins directory e.g. `plugins.databricks.databricks_hook` instead of `airflow.contrib.hooks.databricks_hook`.
3. Export the plugin in the module initialization.

Other than the minor fixes entailed in step 2, the tests and backported functionality should be plug and play.

### DIFF
--- a/plugins/databricks/__init__.py
+++ b/plugins/databricks/__init__.py
@@ -1,0 +1,11 @@
+from airflow.plugins_manager import AirflowPlugin
+from databricks import databricks_hook as hook, databricks_operator as operator
+
+
+class DatasetStatusPlugin(AirflowPlugin):
+    name = "dataset_status"
+    hooks = [hook.DatabricksHook]
+    operators = [
+        operator.DatabricksSubmitRunOperator,
+        operator.DatabricksRunNowOperator,
+    ]
--- a/plugins/databricks/__init__.py
+++ b/plugins/databricks/__init__.py
@@ -1,9 +1,23 @@
+"""
+A backport of the Databricks hooks and operators from apache/airflow v1.10 stable.
+
+The Mozilla Data Platform currently runs airflow v1.9, which includes a Databricks
+operator with a failure mode that can cause duplicate data. The operator intermittenly
+fails due to insufficient time between retries, leaving a zombie cluster that continues
+to run in the background. This plugin includes an operator that patches this bug and
+relieves mitigations around this bug, in particular [1] and [2].
+
+[1]
+[2] https://github.com/mozilla/telemetry-airflow/pull/417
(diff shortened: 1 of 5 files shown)


plugins/databricks/__init__.py
```diff
@@ -0,0 +1,11 @@
+from airflow.plugins_manager import AirflowPlugin
+from databricks import databricks_hook as hook, databricks_operator as operator
+
+
+class DatasetStatusPlugin(AirflowPlugin):
+    name = "dataset_status"
```
sunahsuh: Looks like some copypasta here and in the class name

plugins/databricks/__init__.py
```diff
@@ -0,0 +1,11 @@
+from airflow.plugins_manager import AirflowPlugin
+from databricks import databricks_hook as hook, databricks_operator as operator
+
+
```
sunahsuh: Can you add a comment here about the reasons for backporting?

plugins/databricks/databricks_hook.py
```diff
@@ -0,0 +1,276 @@
+# -*- coding: utf-8 -*-
```
sunahsuh: Can you add a comment pointing to the original code you copied from?

plugins/databricks/databricks_operator.py
```diff
@@ -0,0 +1,480 @@
+# -*- coding: utf-8 -*-
```
sunahsuh: Same re: pointer back to the original

tests/test_databricks_hook.py
```diff
@@ -0,0 +1,453 @@
+# -*- coding: utf-8 -*-
```
sunahsuh: Same re: pointer back to the original

tests/test_databricks_operator.py
```diff
@@ -0,0 +1,438 @@
+# -*- coding: utf-8 -*-
```
sunahsuh: Here too

acmiyaguchi:
@sunahsuh I've modified the comments to point to the original branch, but it looks like the 1.10-stable branch is being updated somewhat regularly. Should I point to the specific revision that the file was take from instead?

PR #422, opened 2019-02-01
Revert "Disable retries for Databricks tasks"

acmiyaguchi:
Reverts mozilla/telemetry-airflow#417 with the inclusion of #419 and #421, which adds the databricks operator from airflow v1.10-stable and increases the number of retries and retry delay to every 30 seconds for 10 minutes.


acmiyaguchi:
It may be safer to not merge this particular commit until we've verified that the databricks plugin is working correctly.

acmiyaguchi:
This should be good to go with the succesful run of #430.

PR #423, opened 2019-02-05
Send Fennec iOS events to Amplitude

akkomar:
This should be merged after https://github.com/mozilla/telemetry-streaming/pull/224


dags/events_to_amplitude.py
```diff
@@ -91,6 +92,23 @@ def key_path(project):
     dag=dag
 )
 
+fennec_ios_events_to_amplitude = EMRSparkOperator(
+    task_id="fennec_ios_events_to_amplitude",
+    job_name="Fennec iOS Events to Amplitude",
+    execution_timeout=timedelta(hours=8),
```
fbertsch: Consider reducing this if it never gets close. I've had it in the back of my mind to reduce e.g. Focus's but never got around to it.

PR #424, opened 2019-02-05
Make the tab severity generator etl job use mozetl-submit

wlach:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1525022 and
https://github.com/mozilla/python_mozetl/pull/309.


wlach:
I tested this locally and it timed out. I don't see how that would be related to these changes though, so I'm just going to merge this as-is. We can always back out/revert/fix if it turns out there was a real issue. I'm not feeling so inclined to spend a lot more time debugging the local airflow environment.

PR #425, opened 2019-02-06
Set AIRFLOW_BUCKET to be telemetry-airflow in dev mode

wlach:
This is where the airflow steps.sh actually lives, which we need for
a development environment to live. Nothing should actually be written
to this location.


wlach:
Hey Anthony, I can't remember if this was ok to land or not? I recall that this used to work without the patch in the days before ops didn't clean out `telemetry-test-bucket ` regularly

PR #426, opened 2019-02-06
Fix relative import in Databricks plugin

acmiyaguchi:
Airflow is running on python 2, so relative imports are not valid syntax. This uses the relative module path.


wlach:
wfm, it will be nice when we start running all the things on python 3!

acmiyaguchi:
I'm looking forward to a python 3 environment, too!

PR #427, opened 2019-02-07
Revert "Increase databricks retries and retry delay to avoid api errors"

acmiyaguchi:
Reverts mozilla/telemetry-airflow#421

```
Broken DAG: [/home/airflow/gcs/dags/main_summary.py] Use keyword arguments when initializing operators
Broken DAG: [/home/airflow/gcs/dags/landfill.py] Use keyword arguments when initializing operators
Broken DAG: [/home/airflow/gcs/dags/longitudinal.py] Use keyword arguments when initializing operators 
```

### DIFF
--- a/plugins/moz_databricks.py
+++ b/plugins/moz_databricks.py
@@ -2,7 +2,7 @@ from os import environ
 from pprint import pformat
 
 from airflow.plugins_manager import AirflowPlugin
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
Tagging for review after the fact -- all of the dags will be delayed by a few hours.

PR #431, opened 2019-02-20
Use Miniconda for Py 3.7

fbertsch:
We need Python 3.7 for the `glean_parser`, and also then the probe-scraper. I don't think I want to make these changes for everyone though.

### DIFF
--- a/jobs/probe_scraper.sh
+++ b/jobs/probe_scraper.sh
@@ -7,17 +7,31 @@ BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
 CACHE_DIR="probe_cache"
 OUTPUT_DIR="probe_data"
 
+ANACONDA_PATH=/mnt/miniconda3
+ANACONDA_SCRIPT=Miniconda3-4.5.12-Linux-x86_64.sh
+
+wget --no-clobber --no-verbose -P /mnt https://repo.anaconda.com/miniconda/$ANACONDA_SCRIPT
+rm -rf $ANACONDA_PATH
+bash /mnt/$ANACONDA_SCRIPT -b -p $ANACONDA_PATH
+
+export PATH=$ANACONDA_PATH/bin:$PATH
+rm /mnt/$ANACONDA_SCRIPT
+
+# Just to ensure, when looking at logs, we are using the correct python binary
+echo $(which python3)
+
 # Clone and setup the scraper
 git clone https://github.com/mozilla/probe-scraper.git
 cd probe-scraper
-python setup.py bdist_egg
-pip install -r requirements.txt
+python3 setup.py bdist_egg
+python3 -m pip install -r requirements.txt
 
 # Create the output and cache directories.
 mkdir $CACHE_DIR $OUTPUT_DIR
 
 # Finally run the scraper.
-python probe_scraper/runner.py --out-dir $OUTPUT_DIR --cache-dir $CACHE_DIR
+# Using -m allows for relative imports in runner.py
+python3 -m probe_scraper.runner --out-dir $OUTPUT_DIR --cache-dir $CACHE_DIR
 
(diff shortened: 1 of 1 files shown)


fbertsch:
DO NOT MERGE until https://github.com/mozilla/probe-scraper/pull/69

jobs/probe_scraper.sh
```diff
@@ -7,17 +7,25 @@ BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
 CACHE_DIR="probe_cache"
 OUTPUT_DIR="probe_data"
 
+ANACONDA_PATH=/mnt/anaconda2
```
fbertsch: This comes directly from `emr-boostrap-spark`.

jobs/probe_scraper.sh
```diff
@@ -7,17 +7,25 @@ BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
...
+ANACONDA_SCRIPT=Anaconda3-5.3.1-Linux-x86_64.sh
```
mdboom: You could probably get away with Miniconda for this, since we don't need large parts of the scientific stack (Scipy etc.)

jobs/probe_scraper.sh
```diff
@@ -7,17 +7,25 @@ BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
...
+rm -rf $ANACONDA_PATH
+bash /mnt/$ANACONDA_SCRIPT -b -p $ANACONDA_PATH
+rm /mnt/$ANACONDA_SCRIPT
+
 # Clone and setup the scraper
 git clone https://github.com/mozilla/probe-scraper.git
 cd probe-scraper
-python setup.py bdist_egg
-pip install -r requirements.txt
+python3 setup.py bdist_egg
```
fbertsch: At this point `python3` is just a `python` alias, but it does make it obvious what we're using.

jobs/probe_scraper.sh
mdboom: Maybe `anaconda3` since that's what's being installed?

jobs/probe_scraper.sh
```diff
@@ -7,17 +7,25 @@ BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
 CACHE_DIR="probe_cache"
 OUTPUT_DIR="probe_data"
 
+ANACONDA_PATH=/mnt/anaconda2
+ANACONDA_SCRIPT=Anaconda3-5.3.1-Linux-x86_64.sh
+
+wget --no-clobber --no-verbose -P /mnt http://repo.continuum.io/archive/$ANACONDA_SCRIPT
+rm -rf $ANACONDA_PATH
+bash /mnt/$ANACONDA_SCRIPT -b -p $ANACONDA_PATH
+rm /mnt/$ANACONDA_SCRIPT
```
mdboom: Don't we need to active the conda environment (at least by adding it to `PATH`)?  Do we know what `which python3` is below?

jobs/probe_scraper.sh
fbertsch: Good point. I'll test it on miniconda.

jobs/probe_scraper.sh
fbertsch: So this and the following comment (adding to the path) are both because I'm co-opting the existing files and path: https://github.com/mozilla/emr-bootstrap-spark/blob/master/ansible/files/bootstrap/telemetry.sh#L330.

That's probably not the best method, it should be fine to add this to `anaconda3` and point `PATH` to that first.

fbertsch:
@mdboom I updated this to use Miniconda and a new mount location. I ran this with the `2to3` branch of `probe-scraper`, and compared it to the output of the `master` using python2, and they matched exactly. This should be good to merge.

PR #439, opened 2019-02-22
Add mpopova to error notification list for iOS Amplitude job

akkomar:


akkomar:
Yes, you're right, better stick to the Statuspage for out-of-team notifications. I'll close this and will open another PR.

PR #443, opened 2019-02-25
Revert "[hotfix] Set default python version to 2"

acmiyaguchi:
Reverts mozilla/telemetry-airflow#441. The py3 import errors are fixed with mozilla/python_mozetl#319, which adds a `test_cli_import` that should prevent this from happening again.


acmiyaguchi:
@fbertsch bump, re: r?

PR #452, opened 2019-03-14
Re-add test dags since we've switched to mysql on gke

haroldwoo:
Will remove later.  Just want to run a quick check on dag execution now that we will use mysql backend.


haroldwoo:
not easily.  Unless we want to change the way we build the container to
allow root access

PR #453, opened 2019-03-14
Fix container to use mysql backend. pymysql wasn't working

haroldwoo:


acmiyaguchi:
From the logs: `E: Package 'libmysqlclient-dev' has no installation candidate`

haroldwoo:
yeah building with libmysqlclient-dev was working locally for me, but using default-libmysqlclient-dev works on circleci. At some point we'll revisit the dev docker container with postgres

PR #455, opened 2019-03-26
modifying dev container to use mysql to match gke env

haroldwoo:


haroldwoo:
the `make tests` fails on master also (you need to rm apache-airflow[mysql] from requirements.txt to get the dev build to work) for test 'Increase text size for MySQL (not relevant for other DBs' text types)' with TypeError: Incorrect padding.

README.md
```diff
@@ -39,12 +39,14 @@ An Airflow container can be built with
 make build
 ```
 
-You should then run the database migrations to complete the container initialization with
+You should then initialize containers with:
 
 ```bash
-make migrate
+nohup make up &
```
acmiyaguchi: Maybe mention that `make clean` is the appropriate way stop the backgrounded service

haroldwoo: rm'd.  The make has a help prompt and the readme mentions make clean later

PR #472, opened 2019-04-18
Make clients_last_seen dependent on clients_daily_v6_bq_load

jklukas:
In the current configuration, I think the clients_daily query
is going to end up missing new records for the day because it
runs immediately after the EMR job for clients_daily, not waiting
for the data to load to BQ.


jasonthomas:
Thanks for catching this!

PR #479, opened 2019-04-19
Export clients_last_seen_v1 to parquet

relud:
and firefox_desktop_exact_mau28_by_dimensions_v1

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -3,8 +3,9 @@ from airflow.utils import trigger_rule
 from airflow.operators.dummy_operator import DummyOperator
 from airflow.operators.subdag_operator import SubDagOperator
 
+from airflow.contrib.hooks.aws_hook import AwsHook
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
-from airflow.contrib.operators.dataproc_operator import DataprocClusterCreateOperator, DataprocClusterDeleteOperator, DataProcSparkOperator # noqa
+from airflow.contrib.operators.dataproc_operator import DataprocClusterCreateOperator, DataprocClusterDeleteOperator, DataProcSparkOperator, DataProcPySparkOperator # noqa
 from operators.gcp_container_operator import GKEPodOperator
 from airflow.contrib.operators.bigquery_table_delete_operator import BigQueryTableDeleteOperator # noqa
 from airflow.contrib.operators.s3_to_gcs_transfer_operator import S3ToGoogleCloudStorageTransferOperator # noqa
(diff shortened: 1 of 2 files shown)


dags/utils/gcp.py
```diff
@@ -265,3 +265,83 @@ def reprocess_parquet(parent_dag_name,
...
+    with models.DAG(dag_id=dag_prefix + dag_name, default_args=default_args) as dag:
+
+        create_dataproc_cluster = DataprocClusterCreateOperator(
+            task_id="create_dataproc_cluster",
+            cluster_name=cluster_name,
+            gcp_conn_id=gcp_conn_id,
+            project_id=connection.project_id,
+            properties={
+                "core:fs.s3a.access.key": models.Variable.get("s3_export_access_key"),
+                "core:fs.s3a.secret.key": models.Variable.get("s3_export_secret_key"),
```
relud: These variables would require configuring in airflow.

I'm not sure if this is the best way to do this, because this method makes the credentials visible to anyone that can describe the dataproc cluster, and IDK if `models.Variable` is an acceptable place to store even very restricted credentials.

I could alternatively write to a `gs://` location in from spark, and then call a kubernetes operator to copy from GCS to S3, but for clients_last_seen this is 375 GiB being copied, and I would prefer to take advantage of spark's parallelism to do it.

cc @haroldwoo and @jasonthomas

relud:
This should wait until the next back fill of `clients_last_seen_v1`. There are two columns that need to be dropped (`last_seen_date` and `generated_time`), and [two new columns](https://bugzilla.mozilla.org/show_bug.cgi?id=1545799) being added to `clients_daily_v6` for the last 6 months or so, which will need to propagate through.

dags/utils/gcp.py
haroldwoo: Why do we need to write to s3? Having the aws credentials visible to dataproc users may not be that bad if the keys are restricted to only write to a particular bucket. Another option may be to put the creds in a gcs bucket and have the run_dataproc_pyspark task use the file.

jasonthomas: We are writing back to S3 so that this data is available in our existing AWS tool chains. 

I think doing it this way is fine since the project this will be running in will have limited access. Can we use Airflow Connections here instead to store the AWS credentials [1]? 

When we do provision credentials for this we should limit based on dataset prefix.

[1] https://airflow.readthedocs.io/en/stable/howto/manage-connections.html

relud: we need to get these tables to s3 so they are accessible in athena, presto, and spark. This is blocking us from disabling client_count_daily, desktop_dau and desktop_active_dau parquet tables, as well as blocking some MAU-type analysis for an intern project.

as for limiting creds to write a specific bucket, makes sense, though I would probably want to further restrict it to specific paths within the telemetry-parquet bucket, just so that it can't mess with any unexpected tables. it's probably fine because everyone that can see dataproc in that project probably also has full access to the dev iam on aws.

none of this addresses the question of using `models.Variable` to store credentials though. also, would it work to just use `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` from the environment, now that this runs in k8s?

haroldwoo: those keys are already in use and have a lot more access (full emr, etc.)

relud: using a connection seems great, I'll see if I can figure out how to pass connection creds to dataproc

relud: okay, pretty sure I'm doing that correctly now.

dags/utils/gcp.py
```diff
@@ -265,3 +266,88 @@ def reprocess_parquet(parent_dag_name,
...
+    :param Optional[Dict[str, Any]] default_args: DAG configuration
+    :param str gcp_conn_id:                       Airflow connection id for GCP access
+    :param str dataproc_storage_bucket:           Dataproc staging GCS bucket
+    :param str dataproc_zone:                     GCP zone to launch dataproc clusters
+    :param int num_preemptible_workers:           Number of Dataproc preemptible workers
+
+    :return: airflow.models.DAG
+    """
+
+    cluster_name = table.replace("_", "-") + "-{{ ds_nodash }}"
```
jasonthomas: cluster_name needs to be 55 characters or less. This seems fine for `firefox-desktop-exact-mau28-by-dimensions-v1-YYYYMMDD` which ends up being 54 chars but adding a note for future reference.

relud: turns out it needs to be `firefox-desktop-exact-mau28-by-dimensions-v1-YYYYMMDD` is 53 characters, and it needs to be under 52 >_<

PR #482, opened 2019-04-24
Import direct2parquet bigquery load jobs

jasonthomas:
This is a continuation of https://github.com/mozilla/telemetry-airflow/pull/465 but for D2P datasets. 

Before we merge we will need to configure `aws_prod_iam_s3` connection in WTMO Airflow.

I've tested this locally.  r?

### DIFF
--- a/dags/direct2parquet_bigquery_load.py
+++ b/dags/direct2parquet_bigquery_load.py
@@ -1,0 +1,116 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from airflow.operators.subdag_operator import SubDagOperator
+from airflow.contrib.operators.bigquery_operator import BigQueryOperator
+from utils.gcp import load_to_bigquery
+
+default_args = {
+    'owner': 'jthomas@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2019, 4, 23),
+    'email': ['telemetry-alerts@mozilla.com', 'jthomas@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'retries': 2,
+    'retry_delay': timedelta(minutes=30),
+}
+
+
+dag_name = 'direct2parquet_bigquery_load'
+
+with DAG(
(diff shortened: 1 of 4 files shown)


dags/direct2parquet_bigquery_load.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            kwargs['date_submission_col'] = values['date_submission_col']
+        except KeyError:
+            pass
+
+        tasks[task_name] = SubDagOperator(
+                            subdag=load_to_bigquery(**kwargs),
+                            task_id=task_name)
+
+    core_clients_daily = BigQueryOperator(
```
jasonthomas: @jklukas Does `depends_on_past` need to be set for these?

relud: It's only needed for clients_last_seen

jklukas: `depends_on_past` needs to be set for each of the `*_last_seen` jobs, but not for `*_daily`.

dags/direct2parquet_bigquery_load.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    tasks['telemetry_core_parquet_bigquery_load'] >> core_clients_daily
+
+    core_clients_last_seen = BigQueryOperator(
+        task_id='core_clients_last_seen',
+        bql='sql/core_clients_last_seen_v1.sql',
+        destination_dataset_table='telemetry.core_clients_last_seen_v1${{ds_nodash}}', # noqa
+        write_disposition='WRITE_TRUNCATE',
+        use_legacy_sql=False,
```
jklukas: `depends_on_past=True` here

dags/direct2parquet_bigquery_load.py
jklukas: Relatedly, I opened a PR for making this same change for the fxa dag: https://github.com/mozilla/telemetry-airflow/pull/483

PR #495, opened 2019-05-01
added pypi library support to match python_mozetl repo

crankycoder:
Support for passing in PyPI libraries was added to the
python_mozetl/bin/mozetl-databricks runner in :

https://github.com/mozilla/python_mozetl/commit/b1789afbd4b1ea94daf89014ec84363111883099

This is a minimal patch to add support to extend the 'libraries' attribute in run-submit JSON blob to get PyPI libraries to load in the Databricks cluster


crankycoder:
The dependencies in setup.py for taar itself are not meaningful for running taar as a library.  taar is normally a web service for AMO.  

We ended up merging the two taar repos (library vs web service) after doing a lot of cleanup - and there wasn't a good reason to keep 'just' the webservice when it wasn't doing anything significant to warrant it's own repository.  The webservice ended up being just flask and a call to the `import` library to import taar itself.

The intent of passing in the pypi libraries to the runner directly lets us run the ETL job without having to import all the external libraries that a webservice is going to require.  This patch is to support running the ensemble ETL job which needs to run the collaborative, similarity and locale recommenders in TAAR.

PR #496, opened 2019-05-01
Fix bigquery-etl helper

relud:
- Convert tuple to list in bigquery-etl helper
- Add default for name in bigquery-etl helper


relud:
fixes currently broken `main_summary` dag

PR #503, opened 2019-05-14
Add kpi_dashboard DAG

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/120

### DIFF
--- a/dags/kpi_dashboard.py
+++ b/dags/kpi_dashboard.py
@@ -1,0 +1,29 @@
+import datetime
+
+from airflow import models
+from utils.gcp import bigquery_etl_query
+
+default_args = {
+    'owner': 'jklukas@mozilla.com',
+    'start_date': datetime.datetime(2019, 5, 12),
+    'email': ['telemetry-alerts@mozilla.com', 'jklukas@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'depends_on_past': False,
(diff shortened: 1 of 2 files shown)


dags/kpi_dashboard.py
```diff
@@ -0,0 +1,29 @@
...
+
+with models.DAG(
+        dag_name,
+        # KPI dashboard refreshes at 16:00 UTC, so run this 15 minutes beforehand.
+        schedule_interval='45 15 * * *',
+        default_args=default_args) as dag:
+
+    kpi_dashboard = bigquery_etl_query(
+        task_id='kpi_dashboard',
+        destination_table='firefox_kpi_dashboard_v1',
```
relud:
```suggestion
        destination_table='firefox_kpi_dashboard_v1',
        partitioned=False,
```

dags/kpi_dashboard.py
```diff
@@ -0,0 +1,29 @@
...
+dag_name = 'kpi_dashboard'
+
+with models.DAG(
+        dag_name,
+        # KPI dashboard refreshes at 16:00 UTC, so run this 15 minutes beforehand.
+        schedule_interval='45 15 * * *',
+        default_args=default_args) as dag:
+
+    kpi_dashboard = bigquery_etl_query(
+        task_id='kpi_dashboard',
```
relud: I don't think this should override the default of using `destination_table` for `task_id`
```suggestion
```

PR #505, opened 2019-05-15
Bug 1551853 - Update source data version

mreid-moz:
Update telemetry-core-parquet from v2 to v3.


mreid-moz:
Agreed re: deprecating. That's the next step :)  Thanks!

PR #506, opened 2019-05-15
Fix regression in building SQL file names for queries

jklukas:
[Observed failure](https://workflow.telemetry.mozilla.org/log?task_id=core_clients_daily&dag_id=direct2parquet_bigquery_load&execution_date=2019-05-14T02%3A00%3A00%2B00%3A00):

> /app/script/entrypoint: line 31: sql/core_clients_daily_v1$20190514.sql: No such file or directory

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -397,6 +397,7 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
+    sql_file_path = "sql/{}.sql".format(destination_table)
     if date_partition_parameter is not None:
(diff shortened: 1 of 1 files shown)


dags/utils/gcp.py
```diff
@@ -397,6 +397,7 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
+    sql_file_path = "sql/{}.sql".format(destination_table)
```
jklukas: We need to build the sql file path _before_ we modify the destination_table to include a date suffix

jklukas:
Going ahead and merging so that I can get started on making sure necessary jobs run, but looking for retroactive approval on the changes.

PR #508, opened 2019-05-17
Update main_summary dag to run clients_last_seen_raw

jklukas:
This needs to be merged at the same time as
https://github.com/mozilla/bigquery-etl/pull/128


dags/main_summary.py
```diff
@@ -363,16 +363,16 @@
...
-    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
+    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com", "jklukas@mozilla.com"],
     depends_on_past=True,
     start_date=datetime(2019, 4, 15),
     dag=dag)
 
 clients_last_seen_export = SubDagOperator(
     subdag=export_to_parquet(
-        table="clients_last_seen_v1",
+        table="clients_last_seen_raw_v1",
```
jklukas: @relud Do you know if it's possible for export_to_parquet to export a view rather than a table? I assume at least that it would more expensive since exporting a table can rely on the Copy API, but I think it would be worth eating the query cost to export the view that includes `days_since_seen` rather than just the bit arrays which would be difficult to work with in Athena, etc. But I'm not sure if the connector can handle targeting a view.

relud: i haven't tried, let me check

relud:
Are you sure you didn't mean to link https://github.com/mozilla/bigquery-etl/pull/128 in the description?

jklukas:
> Are you sure you didn't mean to link mozilla/bigquery-etl#128 in the description?

I don't know what you're talking about. It's said 128 in the description this whole time 😉  (you're right; I updated the description).

dags/main_summary.py
relud: > java.lang.RuntimeException: Table type VIEW is not supported

Nope. I can make a utility function that creates a subdag to export a day from a view, or we could make a view in athena. I favor the latter.

jklukas: I forgot that Athena supports views now. If that's feasible, I agree that's a better option. Hopefully, we can make the view union the existing clients_last_seen_v1 data with the new clients_last_seen_raw_v1 data so that we don't have to backfill.

jklukas: I just confirmed that we can do the position calculation as `log2(bitwise_and(16, -16))` in athena, so that's good.

jklukas: I'll coordinate with @robotblake about how best to make these changes.

jklukas: It looks like redash does not display Presto/Athena views in the schema browser. I tried creating a view and querying it, which worked fine, but users would not be able to discover the view on their own.

Issue #502, opened 2019-05-08
`make run COMMAND=...` fails due to failed requests from long startup time


acmiyaguchi:
The `make run` command does not work properly because the mysql database takes too long to respond. The wait script does not check whether the database is accepting requests, and airflow is able to make queries before it is ready.

PR #511, opened 2019-05-17
Update README.md

### LINKED ISSUE #501: Broken dags due to direct boto3 usage in MozDatabricksOperator
The boto3 object causes the MozDatabrickOperator to fail during dag construction in the dev environment if the corresponding environment variables are not defined:

https://github.com/mozilla/telemetry-airflow/blob/b8a44075d00ad66cfe6162509e86359e28c0ac0e/plugins/moz_databricks.py#L177

sunahsuh:
- Adds a note about allowing database migrations to run on first run (or after a build/clean)
- Adds a note about needing environment variables w AWS/DB creds set before dags will parse. Will be obsolete if we fix #501

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,19 @@ An Airflow container can be built with
 make build
 ```
 
+### Export Credentials
+
+For now, DAGs that use the Databricks operator won't parse until the following environment variables are set (see issue #501):
+```
+AWS_SECRET_ACCESS_KEY
+AWS_ACCESS_KEY_ID
+DB_TOKEN
+```
+
+### Migrate Database
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,19 @@ An Airflow container can be built with
...
+### Export Credentials
+
+For now, DAGs that use the Databricks operator won't parse until the following environment variables are set (see issue #501):
+```
+AWS_SECRET_ACCESS_KEY
+AWS_ACCESS_KEY_ID
+DB_TOKEN
+```
+
+### Migrate Database
```
acmiyaguchi: This also agrees with my observations from https://github.com/mozilla/telemetry-airflow/issues/502, and is probably enough to close it.

PR #512, opened 2019-05-18
Bug 1551853 - Fix up channel aliasing

mreid-moz:
3rd time's a charm?


fbertsch:
*sigh* can't believe we missed it earlier

PR #513, opened 2019-05-20
Bump search_clients version for BigQuery load tasks Bug 1552893

jasonthomas:
r?


dags/main_summary.py
```diff
@@ -274,7 +274,7 @@
         dataset_s3_bucket="telemetry-parquet",
         aws_conn_id="aws_dev_iam_s3",
         dataset="harter/searchdb",
-        dataset_version="v4",
+        dataset_version="v5",
         bigquery_dataset="search",
         gke_cluster_name="bq-load-gke-1",
         p2b_table_alias="search_aggregates_v4",
```
acmiyaguchi: This will rename the search dashboard to v5, but it will still effectively be the same content as `search_aggregates_v4`?

jasonthomas: I am not sure. I just noticed that `harter/searchdb/v4` no longer has new data after `2019-05-12` but `harter/searchdb/v5` dones.

acmiyaguchi: It might be a good idea to remove this alias or update it for the version bump. I took a look at the source to figure out what would happen: 

https://github.com/mozilla/telemetry-airflow/blob/9f7f620d9e1c41198fde36bf6413187c5bc1e51e/dags/utils/gcp.py#L112

It would be confusing to use a table named with a different version than the source dataset.

jasonthomas: Err, apologies. Yes, `search_aggregates_v4` should be updated to `search_aggregates_v5`.

PR #515, opened 2019-05-22
Export clients_last_seen_raw to clients_last_seen Parquet

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/151


jklukas:
cc @robotblake 

This should allow us to avoid Athena/Presto views completely, and we'll keep on populating the existing 's3://telemetry-parquet/clients_last_seen/v1/` path rather than a raw one.

PR #519, opened 2019-05-28
Bug 1549695 - Sync probe-scraper cache to s3

fbertsch:
The only question is which bucket to use, currently it's `telemetry-airflow`. @jasonthomas, is this acceptable?

Long-term we'll switch this all to GCP, but this is a short-term fix.

### DIFF
--- a/jobs/probe_scraper.sh
+++ b/jobs/probe_scraper.sh
@@ -4,6 +4,7 @@
 set -x
 
 BUCKET="net-mozaws-prod-us-west-2-data-pitmo"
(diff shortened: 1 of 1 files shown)


fbertsch:
I'll run this once with an extended timeout to seed the cache.

jasonthomas:
> The only question is which bucket to use, currently it's `telemetry-airflow`. @jasonthomas, is this acceptable?

Yes.

fbertsch:
The cache has been seeded with data from a current run.

PR #527, opened 2019-06-04
Add functionality to run prio validation. Fix Gke cluster creation/deletion

haroldwoo:


dags/prio.py
```diff
@@ -0,0 +1,71 @@
...
+
+run_prio = GKEPodOperator(
+    task_id='run_prio_a',
+    gcp_conn_id=gcp_conn_id,
+    project_id=connection.project_id,
+    location='us-west1-b',
+    cluster_name=cluster_name,
+    name='run-hello-world',
+    namespace='default',
+    image='hello-world:latest',
```
haroldwoo: When Anthony has a prio container to run, we would execute it here.  Maybe we would want two dags for the two separate projects?

dags/operators/gcp_container_operator.py
```diff
@@ -78,3 +82,46 @@ def _set_env_from_extras(self, extras):
             os.environ[G_APP_CRED] = service_key.name
             os.environ[GCLOUD_APP_CRED] = service_key.name
             service_key.close()
+
+class GKEClusterCreateOperator(GKEClusterCreateOperator):
+    """
+    We override methods to fix AIRFLOW-4518
```
whd: Can you explain in what way we're overriding these? A link to the actual issue instead of its tag would also be helpful, since my initial search for "AIRFLOW-4518" in two search engines did not elucidate.

dags/prio.py
```diff
@@ -0,0 +1,71 @@
...
+    service_account="prio-runner-a@moz-fx-priotest-project-a.iam.gserviceaccount.com",
+    owner_label="hwoo",
+    team_label="dataops"
+    )
+
+gke_dag = DAG(
+    'gke_prio_a',
+    default_args=DEFAULT_ARGS,
+    dagrun_timeout=timedelta(hours=6),
+    schedule_interval='0 2 * * *'
```
whd: I don't know what the scheduling is supposed to look like other than weekly.

dags/prio.py
whd: Two separate dags would make handing off a single one to another party easier, but it's reasonable to make them a single dag as well (scheduling is an implementation detail).

dags/utils/gke.py
```diff
@@ -0,0 +1,60 @@
...
+                        "owner": owner_label,
+                        "team": team_label
+                    },
+                    "preemptible": preemptible,
+                    "diskType": disk_type
+                },
+                "initialNodeCount": 1,
+                "autoscaling": {
+                    "enabled" : True,
+                    "minNodeCount": 1,
```
whd: This helper function seems fine to me but we may want to extend it in future.

dags/prio.py
acmiyaguchi: You can use "mozilla/prio:latest" and run "scripts/test-cli-integration" as the command if you want a more reasonable set of logs to test with. 

https://github.com/mozilla/python-libprio/blob/e8d4fc07bf3ae0059af6218673e655285a62bb09/.circleci/config.yml#L66

dags/prio.py
haroldwoo: oh yeah, thanks

haroldwoo:
The image was working before but may have changed recently. I was planning to get this merged, keep the dags toggled off in wtmo, and you could iterate and test it locally yourself.

PR #534, opened 2019-06-12
increase instance_count for bgbb_pred

wcbeard:
* shrink window to 90 days


wcbeard:
@acmiyaguchi this should stop the jobs from timing out.

Also, I decreased the window on the *fit* job earlier, but not the predictions job, hence the continued timeouts.

PR #535, opened 2019-06-18
New dag to backfill 10 years of adi processing from vertica

haroldwoo:
Will need to be executed via airflow backfill -s <date> -e 2019-04-17


jasonthomas:
I would just suggest updating your commit message to reflect what is being added here. r+wc

PR #538, opened 2019-06-25
Use databricks to run crash summary

wlach:
Essentially using the same configuration as we do for "main summary
all histograms", with a lower max instance size.

### DIFF
--- a/dags/crash_summary.py
+++ b/dags/crash_summary.py
@@ -1,7 +1,7 @@
 from airflow import DAG
+from airflow.operators.moz_databricks import MozDatabricksSubmitRunOperator
 from airflow.operators.subdag_operator import SubDagOperator
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


wlach:
This should make the crash summary backfill faster. @fbertsch I will be out later today, but feel free to merge this after making any modifications you deem appropriate.

dags/crash_summary.py
```diff
@@ -18,10 +18,16 @@
 
 dag = DAG('crash_summary', default_args=default_args, schedule_interval='@daily')
 
-crash_summary_view = EMRSparkOperator(
+crash_summary_view = MozDatabricksSubmitRunOperator(
     task_id="crash_summary_view",
     job_name="Crash Summary View",
-    instance_count=20,
+    instance_count=5,
+    max_instance_count=40,
```
fbertsch: We _may_ want to configure this with less, but we can change it later if the nodes seem under utilized.

PR #539, opened 2019-06-25
Update environment variables in schema generator

acmiyaguchi:
This updates the schema generator job with the interface exposed in mozilla/mozilla-schema-generator#37.

### DIFF
--- a/dags/probe_scraper.py
+++ b/dags/probe_scraper.py
@@ -45,7 +45,12 @@ with DAG('probe_scraper',
         namespace='default',
         image='mozilla/mozilla-schema-generator:latest',
         image_pull_policy='Always',
-        env_vars={"MPS_SSH_KEY_BASE64": "{{ var.value.mozilla_pipeline_schemas_secret_git_sshkey_b64 }}"},
+        env_vars={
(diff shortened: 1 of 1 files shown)


fbertsch:
@acmiyaguchi in retrospect, maybe we should have `MPS_BRANCH_PUBLISH` be a config-level argument for Airflow that defaults locally to `test-generated-schemas`, and is `generated-schemas` on prod. Downside is one more level of indirection which is more difficult to reason about, but could prevent accidental overwrites.

acmiyaguchi:
This should be doable with a jinja macro. It's not clear how to set up the local airflow environment with access to the production kubernetes cluster, so I'm not too worried about accidental writes here. It's a good thing to note and watch out for though.

fbertsch:
It doesn't need access to the prod cluster, one can overwrite mozilla/generated-schemas using a local Docker instance with an allowable ssh key. This can happen if one sets the correct variable in the local Airflow UI.

fbertsch:
I only know because I did it once :)

fbertsch:
Ah oops, I see what you're saying. Running on the GKE cluster is hard. Ignore this!

PR #543, opened 2019-07-05
Use only one node for crash summary

wlach:
It seems like this job does not benefit from using more than one machine,
so let's just use one and save some costs.


wlach:
Spark UI showing near identical results with a one load cluster:

image

I did find some significant areas of this job which were quite inefficient which I addressed in https://github.com/mozilla/telemetry-batch-view/pull/546

PR #545, opened 2019-07-16
bump search clients daily bq load version

relud:
it looks like search clients daily was bumped to v7, and backfilled, but bigquery is still only only importing v6

cc @harterrt


jasonthomas:
FWIW I noticed that v7 was missing this morning and have already transferred the parquet data over to GCS. I am planning on loading all the v7 data into bigquery later today.

PR #547, opened 2019-07-17
Export Fenix events to Amplitude

fbertsch:
@acmiyaguchi for the connections changes
@haroldwoo for the BQ pieces

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -25,6 +25,30 @@ wait_for() {
   done
 }
 
+init_connections() {
+    airflow connections --add \
+        --conn_id databricks_default \
+        --conn_type databricks \
+        --conn_host dbc-caf9527b-e073.cloud.databricks.com:443 \
+        --conn_extra token=${DB_TOKEN}
+
+    airflow connections --add \
+        --conn_id google_cloud_prio_a \
+        --conn_type google_cloud_platform
+
+    airflow connections --add \
+        --conn_id google_cloud_prio_b \
+        --conn_type google_cloud_platform
+
+    airflow connections --add \
+        --conn_id google_cloud_derived_datasets \
+        --conn_type google_cloud_platform
+
+    airflow connections --add \
+        --conn_id amplitude_s3_conn \
+        --conn_type s3
+}
+
 [ $# -lt 1 ] && usage
 
 # Only wait for backend services in development
@@ -38,6 +62,8 @@ case $1 in
   web)
     airflow initdb
     airflow upgradedb
+    init_connections
+
     exec airflow webserver -p ${PORT} --workers 4
     ;;
   worker)
--- a/bin/run
+++ b/bin/run
@@ -62,7 +62,9 @@ case $1 in
   web)
     airflow initdb
     airflow upgradedb
-    init_connections
+
+    # Only init connections in dev
+    [ ! -z ${DEVELOPMENT+check} ] && init_connections
 
     exec airflow webserver -p ${PORT} --workers 4
     ;;
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -71,7 +71,7 @@ with models.DAG(
         delimiter=extension,
         google_cloud_storage_conn_id=gcp_conn_id,
         dest_aws_conn_id=amplitude_s3_conn,
-        dest_s3_key='s3://{bucket}/{dir}'.format(bucket=amplitude_s3_bucket, dir=directory),
+        dest_s3_key='s3://{bucket}'.format(bucket=amplitude_s3_bucket),
         replace=True
     )
 
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -17,6 +17,8 @@ default_args = {
     'retry_delay': datetime.timedelta(minutes=10),
 }
 
+environment = '{{ task.__class__.deploy_environment }}'
+
 dag_name = 'bq_events_to_amplitude'
 
 # This connection needs access to both BQ and GCS
@@ -40,13 +42,13 @@ with models.DAG(
         gke_cluster_name='bq-load-gke-2',
         destination_table=table_name,
         date_partition_parameter=None,
-        parameters=('submission_date' + ":DATE:{{ds}}",)
+        parameters=('submission_date' + ':DATE:{{ds}}',)
     )
 
     project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
 
-    gcs_bucket = "moz-fx-data-amplitude-export"
-    directory = 'fenix/{{ ds_nodash }}/'
+    gcs_bucket = 'moz-fx-data-amplitude-export'
+    directory = '{env}/fenix/{{ ds_nodash }}/'.format(env=environment)
     extension = '.tsv.gz'
 
     # Export from bq to gcs
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -46,7 +46,7 @@ with models.DAG(
 
     project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
 
-    gcs_bucket = 'moz-fx-data-amplitude-export'
+    gcs_bucket = 'moz-fx-data-derived-datasets-amplitude-export'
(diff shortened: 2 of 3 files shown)


dags/bq_events_to_amplitude.py
```diff
@@ -0,0 +1,78 @@
...
+    'email_on_retry': True,
+    'depends_on_past': False,
+    'retries': 2,
+    'retry_delay': datetime.timedelta(minutes=10),
+}
+
+dag_name = 'bq_events_to_amplitude'
+
+# This connection needs access to both BQ and GCS
+gcp_conn_id = 'google_cloud_derived_datasets'
```
fbertsch: We may want to make a separate gcp connection, with a different project, to use for the GCS access. That way we don't need to change the `google_cloud_derived_datasets` service account to use GCS.

bin/run
```diff
@@ -25,6 +25,30 @@ wait_for() {
   done
 }
 
+init_connections() {
```
haroldwoo: why are we adding this? How does this behave if the connections already exist on wtmo?

dags/bq_events_to_amplitude.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
+
+    gcs_bucket = 'moz-fx-data-amplitude-export'
+    directory = '{env}/fenix/{{ ds_nodash }}/'.format(env=environment)
+    extension = '.tsv.gz'
+
+    # Export from bq to gcs
+    # Docs: https://github.com/apache/airflow/blob/master/airflow/contrib/operators/bigquery_to_gcs.py#L28
```
haroldwoo: wtmo runs tag 1.10.2, does master branch match for this file?

docker-compose.yml
```diff
@@ -30,11 +30,6 @@ services:
     environment:
       - AWS_SECRET_ACCESS_KEY
       - AWS_ACCESS_KEY_ID
-      # https://docs.databricks.com/api/latest/authentication.html#generate-a-token
```
haroldwoo: this way may be desirable since it doesn't affect prod

bin/run
fbertsch: Check out the commit, it has a rationale for doing it this way.

dags/bq_events_to_amplitude.py
fbertsch: Yes, I tested this out on dev with 1.10.2. There are some changes for 1.10.4, but nothing that effects us.

dags/bq_events_to_amplitude.py
```diff
@@ -0,0 +1,79 @@
...
+        source_project_dataset_table='{project}.telemetry.{table}'.format(project=project_id, table=table_name),
+        destination_cloud_storage_uris=['gs://{bucket}/{dir}*{ext}'.format(bucket=gcs_bucket, dir=directory, ext=extension)],
+        bigquery_conn_id=gcp_conn_id,
+        compression='GZIP',
+        export_format='CSV',
+        field_delimiter='\t',
+        print_header=True
+    )
+
+    # Push the data to S3
```
haroldwoo: Do we need the data in s3 for amplitude? Anyway we can avoid this step? It seems like a move in the wrong direction (AWS)

dags/bq_events_to_amplitude.py
```diff
@@ -0,0 +1,79 @@
...
+    table_name = 'fenix_events_amplitude_v1'
+    table_creation = bigquery_etl_query(
+        destination_table=table_name,
+        date_partition_parameter=None,
+        parameters=('submission_date' + ':DATE:{{ds}}',)
+    )
+
+    project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
+
+    gcs_bucket = 'moz-fx-data-amplitude-export'
```
fbertsch: @haroldwoo we'll need to create a gcs bucket for use with this job

dags/bq_events_to_amplitude.py
haroldwoo: this comment will become outdated though

dags/bq_events_to_amplitude.py
fbertsch: We don't own this s3 bucket, this is for Amplitude only.

fbertsch: This ingestion method has already been approved by Amplitude, IT, and Legal.

dags/bq_events_to_amplitude.py
fbertsch: IMO better than no doc link, it will be close to where the class is defined.

bin/run
haroldwoo: Setting the connections in docker-compose guarantees that we do not introduce it into our prod deployment. We also want to see errors if these connections are not set in prod, rather than guessing if they have been set in the UI correctly.

dags/bq_events_to_amplitude.py
haroldwoo: which gcp project? We will eventually migrate from derived-datasets to the shared-prod project

bin/run
fbertsch: I've confirmed that this does _not_ overwrite existing connections.

As it is, we're guessing now if they have been set correctly - we don't know until the job runs. All we can check is that they are set, so this doesn't change that behavior. Thoughts?

dags/bq_events_to_amplitude.py
fbertsch: I'm not sure. AFAIK you all are in charge of deciding which projects to use and for what purposes. It would be simpler for this all to be in derived-datasets (or shared-prod when we migrate) because then the derived tables and exported TSV data will be in the same project.

bin/run
haroldwoo: I don't see a need to modify our existing prod deployment. This introduces a new set of potential failure cases

dags/bq_events_to_amplitude.py
fbertsch: Oh, right now the project used is whichever is defined in `google_cloud_derived_datasets`

bin/run
fbertsch: Understood. I've updated to just set these in development.

bin/run
```diff
@@ -38,6 +62,10 @@ case $1 in
   web)
     airflow initdb
     airflow upgradedb
+
+    # Only init connections in dev
+    [ ! -z ${DEVELOPMENT+check} ] && init_connections
```
acmiyaguchi: Could we split this out into another file and source it instead? It'll be easier to point someone to `bin/connections.sh` since it'll be single purposed.

docker-compose.yml
acmiyaguchi: Unfortunately airflow reads from the environment over the settings set in the UI (which is an unexpected imo, applications should start reading values closest to the user's control). The dummy GCP connections can't be overwritten without creating the connection string.

docker-compose.yml
```diff
@@ -30,11 +30,6 @@ services:
...
-      - "AIRFLOW_CONN_DATABRICKS_DEFAULT=databricks://:@dbc-caf9527b-e073.cloud.databricks.com:443?token=${DB_TOKEN}"
```
acmiyaguchi: @fbertsch what was the exact issue you were facing when you were setting the environment variable for the GCP connection string? If you were trying to set the extra's field, they get set as url-encoded params:

```
google-cloud-platform://?extra__google_cloud_platform__key_path=%2Fkeys%2Fkey.json&extra__google_cloud_platform__scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform&extra__google_cloud_platform__project=airflow
```
https://airflow.apache.org/howto/connection/gcp.html#configuring-the-connection

It's still probably a lot easier to use the command-line tool.

bin/run
fbertsch: I attempted that, but it never worked. I think we can follow-up with that change as it will take some time to figure out why it wasn't running.

docker-compose.yml
fbertsch: Right, that's what I did for the GCP connection string. However that still didn't work either :/

bin/run
fbertsch: Created in issue: https://github.com/mozilla/telemetry-airflow/issues/548

PR #549, opened 2019-07-17
Bug 1566941 - Move away from EMR, and increase instance count and timeout

acmiyaguchi:
This should fix some of the intermittent timeouts that have been seen in `first_shutdown_summary`.


acmiyaguchi:
See testing logs: https://dbc-caf9527b-e073.cloud.databricks.com/#job/3650/run/1

PR #550, opened 2019-07-17
Load main_summary to clustered bq table in telemetry_raw

relud:
This is blocked on moving all 25 tables at once that have a string `sample_id`, or load a `list` or `map` from parquet


dags/main_summary.py
```diff
@@ -95,7 +95,13 @@
         dataset="main_summary",
         dataset_version="v4",
         gke_cluster_name="bq-load-gke-1",
-        ),
+        bigquery_dataset="telemetry_raw",
+        cluster_by=["sample_id"],
+        drop=["submission_date_s3"],
+        replace=[
+            "submission_date_s3 AS submission_date",
```
jklukas: Does a field named `submission_date` already exist? If not, then this will fail as a REPLACE clause. It may need to come before the `* REPLACE`, which may require an additional new parameter.

relud: yes, it does already exist. for now, this will work. for other tables hiding `submission_date_s3` may required another parameter, correct.

jasonthomas:
I've cut release https://github.com/mozilla/parquet2bigquery/releases/tag/20190717

dags/utils/gcp.py
```diff
@@ -32,7 +32,10 @@ def load_to_bigquery(parent_dag_name=None,
                      spark_gs_dataset_location=None,
                      bigquery_dataset='telemetry',
                      dataset_gcs_bucket='moz-fx-data-derived-datasets-parquet',
-                     gcp_conn_id='google_cloud_derived_datasets'):
+                     gcp_conn_id='google_cloud_derived_datasets'
```
jasonthomas: Missing a comma after `gcp_conn_id='google_cloud_derived_datasets'`

relud:
```suggestion
                     gcp_conn_id='google_cloud_derived_datasets',
```

dags/main_summary.py
relud: https://github.com/mozilla/parquet2bigquery/pull/30

PR #553, opened 2019-07-19
Update airflow bigquery loads to telemetry_raw

relud:


dags/direct2parquet_bigquery_load.py
```diff
@@ -54,6 +54,7 @@
             },
         'firefox-installer-install-parquet': {
             'dataset_version': 'v1',
+            'bigquery_dataset': 'telemetry',
```
jklukas: Why is this one specially pointing at the non-raw dataset?

relud: it's pointed at the non-raw dataset because it didn't move last night, because it doesn't use `submission_date`, `sample_id`, maps, or lists.

PR #554, opened 2019-07-19
Bug: 1566623 The TAAR Ensemble airflow job points to an invalid filename

crankycoder:
The TAAR similarity job points to an invalid databricks airflow runner.  The fileextension should be `.py`, not `.sh`


acmiyaguchi:
This likely won't fix any of the issues seen in bug 1566623 because the runner is ignored in the databricks operator. The URL parameter is an artifact of maintaining backwards compatibility with the EMRSparkOperator.

PR #555, opened 2019-07-19
move clients_last_seen_raw_v1 to telemetry_raw

relud:


dags/main_summary.py
```diff
@@ -378,6 +378,7 @@
 
 clients_last_seen = bigquery_etl_query(
     task_id="clients_last_seen",
+    parameters=("--dataset_id", "telemetry_raw"),
     destination_table="clients_last_seen_raw_v1",
```
jklukas:
```suggestion
    destination_table="clients_last_seen_v1",
```

I want to avoid double-raw naming. So as we move tables named with a `_raw` suffix into the `_raw` datasets, they should lose their raw suffix. This may not be technically feasible at the moment. If so, we should make sure that we factor that renaming into the move into shared-prod.

relud: i will do that as part of https://github.com/mozilla/bigquery-etl/issues/189 and work on that monday

PR #561, opened 2019-07-22
Prepare Prio DAGs for weekly/daily processing

acmiyaguchi:
This adds the Prio DAG for daily processing of aggregates. This will read pings from the data platform, preprocess the data into appropriate partitions, load the data into the prio-processor, aggregate, and finally copy the data into a final location.


acmiyaguchi:
I'm currently running into the following issue:

```
[2019-07-23 00:04:34,268] {logging_mixin.py:95} INFO - [2019-07-23 00:04:34,267] {pod_launcher.py:104} INFO - processor/bin/process: line 31: GOOGLE_APPLICATION_CREDENTIALS: parameter null or not set
```

@jasonthomas I'm not sure the environment variable in the GKEPodOperator is being populated in the container as per:
https://github.com/mozilla/telemetry-airflow/blob/a283877e239bfffff68065690e0d03e8e80c1e45/dags/operators/gcp_container_operator.py#L59-L60

Do you know the best way to set `GOOGLE_APPLICATION_CREDENTIALS` within the running container?

jasonthomas:
> I'm currently running into the following issue:
> 
> ```
> [2019-07-23 00:04:34,268] {logging_mixin.py:95} INFO - [2019-07-23 00:04:34,267] {pod_launcher.py:104} INFO - processor/bin/process: line 31: GOOGLE_APPLICATION_CREDENTIALS: parameter null or not set
> ```
> 
> @jasonthomas I'm not sure the environment variable in the GKEPodOperator is being populated in the container as per:
> 
> https://github.com/mozilla/telemetry-airflow/blob/a283877e239bfffff68065690e0d03e8e80c1e45/dags/operators/gcp_container_operator.py#L59-L60
> 
> Do you know the best way to set `GOOGLE_APPLICATION_CREDENTIALS` within the running container?

The method that you are referring to IIRC is what airflow is using to authenticate and access the GKE cluster. I am assuming the error you are seeing is coming from the actual container running on the cluster?

acmiyaguchi:
I see, I misunderstood where the service account was being written. Is there a way to mount a keyfile into the container? My integration tests are based on [mounting a local keyfile to the container](https://github.com/mozilla/prio-processor/blob/438c1a983225bf319c6f1168fba033b10ac43e5e/processor/docker-compose.yml#L10-L12).

acmiyaguchi:
## Setup instructions

There are three projects that are currently involved:

name | project_id | airflow_conn
-|-|-
prio-admin-prod | moz-fx-prio-admin | google_cloud_prio_admin
priotest-a-prod | moz-fx-priotest-project-a | google_cloud_prio_a
priotest-b-prod | moz-fx-priotest-project-b | google_cloud_prio_b

Create a service account keyfile for each of these projects and save them locally to disk.

Start up the Airflow service using `make up`. Navigate to `Admin > Connections` and fill in `Project Id` and `Keyfile JSON` for each connection.

My keyfiles are stored locally at `~/gcp-service-credentials`, so the following command is an example for copying the credentials:

```bash
jq -c '.' ~/gcp-service-credentials/moz-fx-prio-admin-9fe77690f33b.json 
```

Now, fill in the necessary airflow variables:

    prio_private_key_hex_internal
    prio_public_key_hex_internal
    prio_private_key_hex_external
    prio_public_key_hex_external

These should be set using the production keys. The `priotest-a-prod` and `priotest-b-prod` projects contain keys that can be viewed using the following commands.

```
gcloud config set project moz-fx-priotest-project-a
gsutil cat gs://project-a-secrets/prio-a-secrets.json | jq

gcloud config set project moz-fx-priotest-project-b
gsutil cat gs://project-b-secrets/prio-b-secrets.json | jq
```

Navigate to `Admin > Variables` and set each of the variables appropriately.

Turn on the `prio_processor` dag and let it run.

acmiyaguchi:
@wlach This is now ready for review. This DAG will result in data located in the `prio-admin-prod` project under `gs://moz-fx-data-prio-data/processed/`. This should be followed by a post-processing job that loads data into BigQuery, but this can be done after merging this DAG and verifying that the job is processing correctly.

@hwoo I've tagged you for review since this will require updating the connection strings in Airflow with the service accounts that you've set up. Is it a good idea to have a separate dev/prod environment for this project, and to clean up the names so they are consistent?

```bash
$ gsutil ls -r gs://moz-fx-data-prio-data/processed/
gs://moz-fx-data-prio-data/processed/:

gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/:

gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/batch_id=content.blocking_blocked_TESTONLY-0/:
gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/batch_id=content.blocking_blocked_TESTONLY-0/part-00000-30da8c72-e645-4fd8-94f2-47f7c96645f2.c000.json

gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/batch_id=content.blocking_blocked_TESTONLY-1/:
gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/batch_id=content.blocking_blocked_TESTONLY-1/part-00000-30da8c72-e645-4fd8-94f2-47f7c96645f2.c000.json
```

```bash
$ gsutil cat gs://moz-fx-data-prio-data/processed/submission_date=2019-08-12/batch_id=content.blocking_blocked_TESTONLY-0/part-00000-30da8c72-e645-4fd8-94f2-47f7c96645f2.c000.json
[250, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

dags/prio/processor.py
```diff
@@ -0,0 +1,276 @@
...
+bucket located in each server's project.
+
+The SubDag operator simplifies the definition of each phase. The clusters are
+ephemeral -- there are situations where clusters can become orphaned. These are
+usually resolved by rerunning the individual SubDag, which should clean-up
+properly.
+
+The following Airflow variables should be set:
+
+    prio_private_key_hex_internal
```
haroldwoo: Should we name these a/b instead of internal/external

dags/utils/gke.py
```diff
@@ -17,13 +20,10 @@ def create_gke_config(
...
     """
-    
+
     cluster_def_dict = {
         "name": name,
-        "masterAuthorizedNetworksConfig":
-            {
-            "enabled": True,
-            },
+        "masterAuthorizedNetworksConfig": {"enabled": not is_dev},
```
haroldwoo: thanks for this. Should we add a comment as to why we want to toggle this?

dags/prio/processor.py
acmiyaguchi: These are consistent with the options in the processor: https://github.com/mozilla/prio-processor/blob/074869a8fb1786fe0a5f5d6295374d41563f6bc4/prio/cli/options.py#L11-L28

It makes it easier to reuse code for each server, since the keys are relative to the servers instead of relative to the admin.

acmiyaguchi: It could be clearer to rename these to a/b, but this makes more sense if `processor_b` was out of our control and we didn't have the private key to processor B.

dags/utils/gke.py
acmiyaguchi: The container won't run locally without this, since my development machine is not in the authorized network.

haroldwoo: I meant, adding a comment like "Setting this to false opens the GKE cluster to the world"

dags/prio/processor.py
```diff
@@ -0,0 +1,276 @@
...
+        child_dag_name="processor_a",
+        default_args=DEFAULT_ARGS,
+        server_id="a",
+        gcp_conn_id="google_cloud_prio_a",
+        service_account="prio-runner-a@moz-fx-priotest-project-a.iam.gserviceaccount.com",
+        arguments=["processor/bin/process"],
+        env_vars={
+            "DATA_CONFIG": "/app/processor/config/content.json",
+            "SERVER_ID": "A",
+            "SHARED_SECRET": "m/AqDal/ZSA9597GwMM+VA==",
```
wlach: Should this be here?

dags/prio/processor.py
```diff
@@ -0,0 +1,276 @@
...
+            # TODO: this is the built-in testing key
+            "PRIVATE_KEY_HEX": "{{ var.value.prio_private_key_hex_internal }}",
+            "PUBLIC_KEY_HEX_INTERNAL": "{{ var.value.prio_public_key_hex_internal }}",
+            "PUBLIC_KEY_HEX_EXTERNAL": "{{ var.value.prio_public_key_hex_external }}",
+            "BUCKET_INTERNAL_PRIVATE": "project-a-private",
+            "BUCKET_INTERNAL_SHARED": "project-a-shared",
+            "BUCKET_EXTERNAL_SHARED": "project-b-shared",
+            "RETRY_LIMIT": "5",
+            "RETRY_DELAY": "3",
+            "RETRY_BACKOFF_EXPONENT": "2",
```
wlach: It looks like there is a substantial amount of duplicated code in this `env_vars` variable between processor a and b. Maybe you should define the common parts in a variable above and merge in whatever differs?

dags/prio/processor.py
acmiyaguchi: Good point, this should probably be part of the airflow variables too.

dags/prio/processor.py
acmiyaguchi: If this were python 3, I would be less hesitant since I would be able to do this in place (`{**a, **b}`). The `RETRY_*` and `DATA_CONFIG` variables are the only shared variables, so it's not too bad here.

acmiyaguchi:
I've tested this DAG locally -- it takes 20-30 minutes to run per day due to the overhead of spinning up and down Kubernetes clusters. 

I use this script to update the airflow connections and variables after downloading service keys to my local machine:

https://gist.github.com/acmiyaguchi/9d0b403be8c0d852a514acf3380d8cf5

dags/prio/kubernetes.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Running the pod without any time in-between will cause the scope-based
+        # authentication in Google Cloud Platform to fail. For example:
+        #
+        # `ServiceException: 401 Anonymous caller does not have
+        # storage.objects.get access to moz-fx-prio-dev-a-private/processed/`
+        #
+        # Sleeping by a small amount solves this problem. This issue was first
+        # noticed intermittently on 2019-09-09.
+        sleep = BashOperator(task_id="sleep", bash_command="sleep 15", dag=dag)
```
haroldwoo: How was 15 seconds determined? This seems like it may be an unreliable workaround, as opposed to having the container poll/retry. Another option would be pod readiness probes but it doesn't look like they are supported in k8spod or gkepod operators.

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+
+There are two sets of servers that run in Google Kubernetes Engine (GKE) under
+Data Operations. The servers are assigned service accounts with access to a
+private storage bucket associated with its project, and a pair of mutually
+shared buckets. An administrative account oversees interop with ingestion,
+including a process and transfer job from Firefox Telemetry into a receiving
+bucket located in each server's project.
+
+The SubDag operator simplifies the definition of each phase. The clusters are
+ephemeral -- there are situations where clusters can become orphaned. These are
```
haroldwoo: what situations

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+These variables are encrypted and passed into the prio-processor container via
+the environment.
+"""
+from datetime import datetime, timedelta
+from os import environ
+
+from airflow import DAG
+from airflow.contrib.hooks.gcs_hook import GoogleCloudStorageHook
+from airflow.contrib.operators.gcs_to_gcs import (
+    GoogleCloudStorageToGoogleCloudStorageOperator,
```
haroldwoo: list not required

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 0,
+    "retry_delay": timedelta(minutes=15),
+    "schedule_interval": "@daily",
+    "dagrun_timeout": timedelta(hours=2),
+}
+
+# use a less than desirable method of generating the service account name
+IS_DEV = environ.get("DEPLOY_ENVIRONMENT") == "dev"
```
haroldwoo: You want to default to 'prod' in all cases that it != dev? I would think we would want it the other way around.
environ.get also supports default values.

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+BUCKET_BOOTSTRAP_ADMIN = "moz-fx-data-{}-prio-bootstrap".format(ENVIRONMENT)
+
+dag = DAG(dag_id="prio_processor", default_args=DEFAULT_ARGS)
+
+
+# Copy the dependencies necessary for running the `prio-processor staging` job
+# into the relevant GCS locations. This includes the runner and the egg
+# containing the bundled dependencies.
+
+prio_staging_bootstrap = SubDagOperator(
```
haroldwoo: The use of the word staging here is confusing. The top of file comment doesn't mention a staging job.

dags/prio/kubernetes.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def container_subdag(
+    parent_dag_name,
+    child_dag_name,
+    default_args,
+    gcp_conn_id,
+    service_account,
+    server_id,
+    env_vars,
```
haroldwoo: should 'env_vars' and 'arguments' have empty dict/lists (respectively) as defaults so we don't need to pass in env vars = {} like line 114 of dags/prio/processor.py ?

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+        gcp_conn_id="google_cloud_prio_admin",
+        service_account=SERVICE_ACCOUNT_ADMIN,
+        arguments=[
+            "bash",
+            "-c",
+            "cd processor; prio-processor bootstrap --output gs://{}".format(
+                BUCKET_BOOTSTRAP_ADMIN
+            ),
+        ],
+        env_vars={},
```
haroldwoo: this could be removed if we set default to empty dict

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    ),
+    task_id="bootstrap",
+    dag=dag,
+)
+
+# Run the PySpark job for range-partioning Prio pings.
+prio_staging = SubDagOperator(
+    subdag=dataproc.spark_subdag(
```
haroldwoo: I think this could use the moz_pyspark_runner if if was modified to take a few addl arguments

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+# into the relevant GCS locations. This includes the runner and the egg
+# containing the bundled dependencies.
+
+prio_staging_bootstrap = SubDagOperator(
+    subdag=kubernetes.container_subdag(
+        parent_dag_name=dag.dag_id,
+        child_dag_name="bootstrap",
+        default_args=DEFAULT_ARGS,
+        server_id="admin",
+        gcp_conn_id="google_cloud_prio_admin",
```
haroldwoo: use variable, since this is repeated and used on lines 71-73

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+# Run the PySpark job for range-partioning Prio pings.
+prio_staging = SubDagOperator(
+    subdag=dataproc.spark_subdag(
+        parent_dag_name=dag.dag_id,
+        child_dag_name="staging",
+        default_args=DEFAULT_ARGS,
+        gcp_conn_id="google_cloud_prio_admin",
```
haroldwoo: use variable again

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    logging.info("Deleted {} objects".format(total))
+
+
+clean_processor_a = PythonOperator(
+    task_id="clean_processor_a",
+    python_callable=clean_buckets,
+    op_kwargs={
+        "private_bucket": BUCKET_PRIVATE_A,
+        "shared_bucket": BUCKET_SHARED_A,
+        "google_cloud_storage_conn_id": "google_cloud_prio_a",
```
haroldwoo: use variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+clean_processor_b = PythonOperator(
+    task_id="clean_processor_b",
+    python_callable=clean_buckets,
+    op_kwargs={
+        "private_bucket": BUCKET_PRIVATE_B,
+        "shared_bucket": BUCKET_SHARED_B,
+        "google_cloud_storage_conn_id": "google_cloud_prio_b",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+load_processor_a = GoogleCloudStorageToGoogleCloudStorageOperator(
+    task_id="load_processor_a",
+    source_bucket=BUCKET_DATA_ADMIN,
+    source_object="staging/submission_date={{ ds }}/server_id=a/*",
+    destination_bucket=BUCKET_PRIVATE_A,
+    destination_object="raw/submission_date={{ ds }}/",
+    google_cloud_storage_conn_id="google_cloud_prio_admin",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+load_processor_b = GoogleCloudStorageToGoogleCloudStorageOperator(
+    task_id="load_processor_b",
+    source_bucket=BUCKET_DATA_ADMIN,
+    source_object="staging/submission_date={{ ds }}/server_id=b/*",
+    destination_bucket=BUCKET_PRIVATE_B,
+    destination_object="raw/submission_date={{ ds }}/",
+    google_cloud_storage_conn_id="google_cloud_prio_admin",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+trigger_processor_a = GoogleCloudStorageToGoogleCloudStorageOperator(
+    task_id="trigger_processor_a",
+    source_bucket=BUCKET_DATA_ADMIN,
+    source_object="staging/_SUCCESS",
+    destination_bucket=BUCKET_PRIVATE_A,
+    destination_object="raw/_SUCCESS",
+    google_cloud_storage_conn_id="google_cloud_prio_admin",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+trigger_processor_b = GoogleCloudStorageToGoogleCloudStorageOperator(
+    task_id="trigger_processor_b",
+    source_bucket=BUCKET_DATA_ADMIN,
+    source_object="staging/_SUCCESS",
+    destination_bucket=BUCKET_PRIVATE_B,
+    destination_object="raw/_SUCCESS",
+    google_cloud_storage_conn_id="google_cloud_prio_admin",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+# done for convenience, since both processors are running on the same
+# infrastructure (but partitioned along project boundaries).
+
+processor_a = SubDagOperator(
+    subdag=kubernetes.container_subdag(
+        parent_dag_name=dag.dag_id,
+        child_dag_name="processor_a",
+        default_args=DEFAULT_ARGS,
+        server_id="a",
+        gcp_conn_id="google_cloud_prio_a",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+    dag=dag,
+)
+
+processor_b = SubDagOperator(
+    subdag=kubernetes.container_subdag(
+        parent_dag_name=dag.dag_id,
+        child_dag_name="processor_b",
+        default_args=DEFAULT_ARGS,
+        server_id="b",
+        gcp_conn_id="google_cloud_prio_b",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+)
+
+
+insert_into_bigquery = SubDagOperator(
+    subdag=kubernetes.container_subdag(
+        parent_dag_name=dag.dag_id,
+        child_dag_name="insert_into_bigquery",
+        default_args=DEFAULT_ARGS,
+        server_id="admin",
+        gcp_conn_id="google_cloud_prio_admin",
```
haroldwoo: variable

dags/prio/processor.py
```diff
@@ -0,0 +1,330 @@
...
+            "RETRY_LIMIT": "60",
+            "RETRY_DELAY": "10",
+            "RETRY_BACKOFF_EXPONENT": "1",
+        },
+    ),
+    task_id="processor_b",
+    dag=dag,
+)
+
+
```
haroldwoo: what are we inserting?

dags/prio/dataproc.py
```diff
@@ -0,0 +1,78 @@
+from airflow import DAG
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.contrib.operators.dataproc_operator import (
+    DataprocClusterCreateOperator,
+    DataprocClusterDeleteOperator,
+    DataProcPySparkOperator,
+)
+
+
+def spark_subdag(
```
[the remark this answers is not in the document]
acmiyaguchi: I'd rather follow up with a separate PR instead of expanding the scope of this one. I don't think the dataproc runner was available when I started this PR.

dags/prio/processor.py
acmiyaguchi: Unclean shutdown of airflow, primarily.

dags/prio/processor.py
acmiyaguchi: This is enforced by black due to the length of the line.

dags/prio/processor.py
acmiyaguchi: I think technically `DEPLOY_ENVIRONMENT` could be "stage", so there's this extra logic to make sure it's either dev or prod. Swapped the logic so the fallback is sane.

dags/prio/processor.py
acmiyaguchi: Used staging instead of partitioning to keep the terminology consistent. If this gets changed, it should be changed in the processor job so things are consistent. I do agree that the naming could be better though.

dags/prio/processor.py
acmiyaguchi: I'm going to leave this as is, but it might be worthwhile in the future to refactor to increase code reuse.

dags/prio/processor.py
acmiyaguchi: Aggregates are being inserted into BigQuery.

dags/prio/kubernetes.py
acmiyaguchi: I tried setting `retry=3` with `retry_delay=timedelta(seconds=30)`, but I can't seem to make the subdag task to retry. I did noticed that the 15 second timeout is sometimes not long enough, so I've bumped the timeout to 60 seconds. It's frustrating that the GKE cluster still needs more time to initialize the authorization flow.

I think the next round of deployment should run directly on top of GCE instances with the service account being generated in the instance and mounted into a docker instance. I think this would also help with the 5-7 minutes of GKE startup and also being able to run this on another provider.

haroldwoo: Oh i meant having the prio container retry until authentication is set up

acmiyaguchi: I'm not sure this would solve it, I think it might be at the cluster level as opposed to the container. I'll look into reproducing this in the `prio-processor` side using GCE.

PR #564, opened 2019-07-25
Bug 1566623 - Fix the start_date parameter in the taar_weekly DAG

crankycoder:
The start_date parameter has a bad default causing regularly scheduled runs to fail.  The changes the string to a datetime type for the taar ensemble run operator.

### DIFF
--- a/dags/taar_weekly.py
+++ b/dags/taar_weekly.py
@@ -24,11 +24,14 @@ taar_weekly = DAG(
 )
 
 wait_for_clients_daily = ExternalTaskSensor(
-    task_id='clients_daily',
-    external_dag_id='main_summary',
-    external_task_id='clients_daily',
-    execution_delta=timedelta(days=-7, hours=-1),  # main_summary waits one hour, execution date is beginning of the week
-    dag=taar_weekly)
+    task_id="clients_daily",
+    external_dag_id="main_summary",
+    external_task_id="clients_daily",
+    execution_delta=timedelta(
+        days=-7, hours=-1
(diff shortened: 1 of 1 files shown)


dags/taar_weekly.py
```diff
@@ -43,14 +45,13 @@
...
     enable_autoscale=True,
-    start_date='20190527',
-    pypi_libs=['mozilla-taar3==0.4.5', 'mozilla-srgutil==0.1.10', 'python-decouple==3.1'],
-    env=mozetl_envvar(
-        "taar_ensemble",
-        {
-            "date": "{{ ds_nodash }}",
-        },
-    ),
+    start_date=datetime(2019, 7, 27),
```
acmiyaguchi: Is this date really necessary, since it's provided as a default argument via the DAG.

dags/taar_weekly.py
```diff
@@ -10,7 +10,7 @@
 default_args_weekly = {
     "owner": "vng@mozilla.com",
     "depends_on_past": False,
-    "start_date": datetime(2019, 5, 31),
+    "start_date": datetime(2019, 7, 27),
```
acmiyaguchi: I would suggest setting this to 2019-07-14, so we can run this once it makes it into the production cluster.

dags/taar_weekly.py
crankycoder: I've dropped this parameter in my latest patch.

dags/taar_weekly.py
crankycoder: Changed as per your comment now.

PR #565, opened 2019-07-25
Bug 1566623 - Remove start_date parameter from the taar_weekly DAG

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1566623)

The start_date parameter in the taar_ensemble job is set incorrectly as a string, causing it to fail on scheduled triggers.

@crankycoder I don't think this parameter is necessary since there is already a default start_date for the DAG.


acmiyaguchi:
Oh woops, duplicate of #564

PR #566, opened 2019-07-25
Bug: 1566623 - taar_ensemble job needs a start_date parameter

crankycoder:
Oops.  We actually do need the start_date parameter in the taar_ensemble task.


dags/taar_weekly.py
```diff
@@ -51,6 +51,7 @@
         "python-decouple==3.1",
     ],
     env=mozetl_envvar("taar_ensemble", {"date": "{{ ds_nodash }}"}),
+    start_date=datetime(2019, 7, 14),
     uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-databricks.py",
     output_visibility="private",
```
acmiyaguchi: Alternatively, I think adding the following would solve this too:

```python
    dag=taar_weekly,
```

PR #567, opened 2019-07-26
Add bigquery_etl_copy_deduplicate tasks

jklukas:
Relies on https://github.com/mozilla/bigquery-etl/pull/262

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -7,7 +7,8 @@ from operators.email_schema_change_operator import EmailSchemaChangeOperator
 from utils.mozetl import mozetl_envvar
 from utils.tbv import tbv_envvar
 from utils.status import register_status
-from utils.gcp import bigquery_etl_query, export_to_parquet, load_to_bigquery
+from utils.gcp import bigquery_etl_query, bigquery_etl_copy_deduplicate
+from utils.gcp import export_to_parquet, load_to_bigquery
 
 
 default_args = {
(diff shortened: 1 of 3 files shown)


dags/main_summary.py
```diff
@@ -7,7 +7,8 @@
 from utils.mozetl import mozetl_envvar
 from utils.tbv import tbv_envvar
 from utils.status import register_status
-from utils.gcp import bigquery_etl_query, export_to_parquet, load_to_bigquery
+from utils.gcp import bigquery_etl_query, bigquery_etl_copy_deduplicate
+from utils.gcp import export_to_parquet, load_to_bigquery
```
relud: I think we should do this how python-black does:
```
from utils.gcp import (
    bigquery_etl_query,
    bigquery_etl_copy_deduplicate,
    export_to_parquet,
    load_to_bigquery,
)
```

dags/main_summary.py
```diff
@@ -25,6 +26,18 @@
...
 
+# We copy yesterday's main pings from telemetry_live to telemetry_stable
+# at the root of this DAG because telemetry_stable.main_v4 will become
+# the source for main_summary, etc. once we are comfortable retiring parquet
+# data imports.
+copy_deduplicate_main_ping = bigquery_etl_copy_deduplicate(
+    task_id='copy_deduplicate_main_ping',
+    target_project_id='moz-fx-data-shared-prod',
+    only_tables=['telemetry_live.main_v4'],
+    owner="jklukas@mozilla.com",
```
relud: this uses a mix of `'` and `"`, I think throughout this PR we should consistently use `"` unless the string contains a `"`, like python-black does.

PR #569, opened 2019-07-30
Add options for controlling bigquery-etl project, dataset, and sql path

relud:
and rename the `telemetry_raw` dataset to `telemetry_derived`


dags/utils/gcp.py
```diff
@@ -385,6 +387,9 @@ def bigquery_etl_query(
     destination_table,
     parameters=(),
     arguments=(),
+    dataset_id="telemetry",
```
jklukas: This is not marked as Optional below, but there's a default here. I think it makes sense to remove the default and make this a required param.

jklukas: Or maybe the default here can be a transitional thing until we have all derived tables moved into the telemetry_derived locations.

relud: let's remove it now, just in case we don't get another chance

PR #571, opened 2019-07-31
WIP: 1569184: Add Android SDK to dependencies

mdboom:
This downloads and sets up the Android SDK so that probe_scraper can determine gradle-based dependencies.  (Java is already around on EMR, because Spark...)


fbertsch:
Closing this, as we're not finding dependencies automatically any longer.

PR #572, opened 2019-07-31
Build sql_file_path before appending date to destination_table

jklukas:
Fixes bug introduced in #569 that lead to some failures this morning like:

    /app/script/entrypoint: line 31: sql/telemetry_derived/clients_last_seen_v1$20190730.sql: No such file or directory

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -426,6 +426,7 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
+    sql_file_path = sql_file_pathor or "sql/{}/{}.sql".format(dataset_id, destination_table)
     if date_partition_parameter is not None:
         destination_table = destination_table + "${{ds_nodash}}"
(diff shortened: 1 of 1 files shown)


dags/utils/gcp.py
```diff
@@ -426,6 +426,7 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
+    sql_file_path = sql_file_pathor or "sql/{}/{}.sql".format(dataset_id, destination_table)
```
jklukas: By building sql_file_path here, we avoid picking up the `$20190730` suffix that we append to `destination_table` in the next step.

jklukas:
Merging now to get these queries running. Seeking retroactive review from @relud and we can address any additional changes in an additional PR.

dags/utils/gcp.py
scholtzan: Looks like `sql_file_pathor` is a typo here and should be `sql_file_path`?

scholtzan: Ah nvm, fixed in https://github.com/mozilla/telemetry-airflow/pull/573

jklukas: Yep 😭

PR #574, opened 2019-07-31
Bug 1570405 - Export active_profiles_v1 to BigQuery

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1570405)


dags/main_summary.py
```diff
@@ -605,6 +605,23 @@
     dag=dag
 )
 
+bgbb_pred_bigquery_load = SubDagOperator(
+    subdag=load_to_bigquery(
+        parent_dag_name=dag.dag_id,
+        dag_name="bgbb_pred_bigquery_load",
+        default_args=default_args,
+        dataset_s3_bucket="telemetry-parquet",
+        objects_prefix="bgbb",
```
acmiyaguchi: This is needed because the full path of the parquet dataset is `telemetry-parquet/bgbb/active_profiles/v1`

jasonthomas: `objects_prefix` is misleading and I don't think it will work here as expected.

Instead we should remove `objects_prefix`, set `dataset=bgbb/active_profiles' and also set p2b_table_alias="active_profiles_v1".

PR #582, opened 2019-08-08
Pin numpy version

fbertsch:
Pandas requires numpy>=1.12, so it tries to install numpy 1.17, which is Python 3 only. numpy 1.16.* supports Python 2, and is the latest one to do so.


fbertsch:
Thanks @jklukas!

PR #585, opened 2019-08-14
Bug 1570611 - Decommission Topline summary view

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1570611)


acmiyaguchi:
There shouldn't be any other consumers, so this should be safe to merge in.

PR #594, opened 2019-09-05
Migrate socorro import to dataproc

haroldwoo:
Will need to change a few testing vars before merging.
- [x] The start_date
- [x] bq_gcp_conn_id
- [x] dest_s3_key
- [x] bq_dataset

And WTMO will need a few connections configured
- [x] conn - google_cloud_airflow_dataproc
- [x] conn - aws_socorro_readonly_s3
- [x] conn - aws_dev_socorro_telemetry_parquet_s3 
- [x] conn - google_cloud_derived_datasets


dags/socorro_import.py
```diff
@@ -18,39 +38,154 @@
...
+    image="google/cloud-sdk:242.0.0-alpine",
+    arguments=gsutil_args,
+    env_vars={
+        "AWS_ACCESS_KEY_ID": "{{ var.value.socorro_access_key_id }}",
+        "AWS_SECRET_ACCESS_KEY": "{{ var.value.socorro_secret_access_key }}"
+    },
+    dag=dag,
+)
+
+# We dont use load_to_bigquery since our source data is s3 json.
```
akkomar: This is unclear to me, I'd write:
```suggestion
# We don't use load_to_bigquery since our source data is on GCS.
```
instead. Unless by `source data` you mean original Socorro files, but that's not relevant to BQ load?

dags/socorro_import.py
```diff
@@ -18,39 +38,154 @@
...
+    namespace='default',
+    image=docker_image,
+    arguments=gke_args,
     dag=dag,
 )
 
 register_status(
-    crash_report_parquet,
-    crash_report_parquet.job_name,
+    bq_load,
```
akkomar: :+1:

dags/utils/dataproc.py
```diff
@@ -21,7 +21,7 @@ def __init__(self,
                  cluster_name=None,
                  num_workers=2,
                  image_version='1.4',
-                 zone='us-central1-b',
+                 zone='us-west1-b',
```
akkomar: Why are zones changed to `west` here?

jobs/socorro_import_crash_data.py
```diff
@@ -0,0 +1,196 @@
...
+from datetime import datetime as dt, timedelta, date
+from os import environ
+
+from pyspark.context import SparkContext
+from pyspark.sql.session import SparkSession
+from pyspark.sql.types import *
+
+"""
+Adapted from:
+https://raw.githubusercontent.com/mozilla-services/data-pipeline/master/reports/socorro_import/ImportCrashData.ipynb
```
akkomar: Better link to specific revision:
```suggestion
https://github.com/mozilla-services/data-pipeline/blob/0c94d328f243338d21bae360547c300ac1b82b12/reports/socorro_import/ImportCrashData.ipynb
```

dags/utils/dataproc.py
haroldwoo: We have created the gcs buckets in that region, so we want to avoid cross region transfers. Originally I had set up us-central1-b because I had thought we had resource limits in west1-b, but I was confusing that for gke clusters.

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
-        dataset_version="v2",
-        date_submission_col="crash_date",
-        gke_cluster_name="bq-load-gke-1",
-        bigquery_dataset="telemetry_derived",
-        ),
-    task_id="crash_report_parquet_bigquery_load",
-    dag=dag)
-
-crash_report_parquet >> crash_report_parquet_bigquery_load
+s3_to_gcs >> crash_report_parquet >> gcs_to_s3 >> remove_bq_table_partition >> bq_load
```
acmiyaguchi:
```suggestion
s3_to_gcs >> crash_report_parquet 
crash_report_parquet >> gcs_to_s3 
crash_report_parquet >> remove_bq_table_partition >> bq_load
```

It doesn't look like there's a dependency between the bq load and the gcs to s3 transfer.

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+bq_table_name = '{}_{}'.format(dataset, dataset_version)
+
+docker_image='docker.io/mozilla/parquet2bigquery:20190722' # noqa
+
+gke_args = [
+    '-d', bq_dataset,
+    '-c', '10',
+    '-b', gcs_data_bucket,
+    '-R',
+    '-p', objects_prefix,
```
acmiyaguchi: This should use the long form of the options so the code is self-documenting.

```
--dataset
--concurrency
--bucket
--no-resume
--prefix
```

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+    gcp_conn_id=bq_gcp_conn_id,
+    project_id=bq_connection.project_id,
+    location=gke_location,
+    cluster_name=gke_cluster_name,
+    name='copy-parquet-data-back-to-s3',
+    namespace='default',
+    image="google/cloud-sdk:242.0.0-alpine",
+    arguments=gsutil_args,
+    env_vars={
+        "AWS_ACCESS_KEY_ID": "{{ var.value.socorro_access_key_id }}",
```
acmiyaguchi: The default AWS credentials that are used on EMR/databricks clusters don't work with this dataset?

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+# Required to write parquet data back to s3://telemetry-parquet
+write_aws_conn_id='aws_dev_socorro_telemetry_parquet_s3'
+
+gcstj_object_conditions = {
+    'includePrefixes':  'v1/crash_report/{{ ds_nodash }}'
+}
+
+gcstj_transfer_options = {
+    'deleteObjectsUniqueInSink': True
+}
```
acmiyaguchi: If the variables are not used more than once, I think it makes sense to put them next to the operator they take place in. In particular, `gcstj_object_conditions` and `gcstj_transfer_options` don't have any real meaning outside of the gcs transfer operator.

dags/socorro_import.py
```diff
@@ -1,14 +1,35 @@
...
+s3://telemetry-parquet/socorro_crash/v2
+
+Until we migrate socorro to gcp (https://bugzilla.mozilla.org/show_bug.cgi?id=1512641),
+or at least modify it to write crash stats to gcs, this job will now copy json from s3
+to gcs, use dataproc to rewrite the data to parquet in gcs, and load the parquet data
+into bigquery.
+
+Copying the parquet data back from gcs to s3 will be necessary until we have migrated
+all workloads using the old athena datasource to it's bigquery counterpart.
+"""
```
acmiyaguchi: Could you document the necessary variables and connections in the docstring?

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+    subdag = moz_dataproc_pyspark_runner(
+        parent_dag_name=dag.dag_id,
+        dag_name='crash_report_parquet',
+        default_args=default_args,
+        cluster_name=cluster_name,
+        job_name="Socorro_Crash_Reports_to_Parquet",
+        # TODO - CI/CD for gcs and repo sync for artifact bucket jobs and gcp bootstrap folders
+        python_driver_code="gs://moz-fx-data-prod-airflow-artifacts/jobs/socorro_import_crash_data.py",
+        py_args=[
+            "-d", "{{ ds_nodash }}",
```
acmiyaguchi:
```suggestion
            "--date", "{{ ds_nodash }}",
```

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+gke_args = [
+    '-d', bq_dataset,
+    '-c', '10',
+    '-b', gcs_data_bucket,
+    '-R',
+    '-p', objects_prefix,
+    '--cluster-by', 'crash_date'
+]
+
+remove_bq_table_partition = BigQueryTableDeleteOperator(
```
acmiyaguchi: Why does the partition for the current date need to be removed? Does the `--glob-load` option do the right thing in this case if added to the `bq_load` step?

dags/socorro_import.py
haroldwoo: thanks good catch

dags/socorro_import.py
haroldwoo: They would, but those credentials are to broad

dags/socorro_import.py
haroldwoo: Not sure, just following what we currently do for all of our other derived datasets.

dags/socorro_import.py
haroldwoo: yes

dags/socorro_import.py
```diff
@@ -18,39 +39,155 @@
...
+# bucket prefix, and unfortunately socorro data in s3 has prefix version/dataset instead 
+# of having the dataset name come first
+gcs_data_bucket = 'moz-fx-data-socorro-data'
+
+dataset = 'socorro_crash'
+dataset_version='v2'
+date_submission_col='crash_date'
+ds_type='ds_nodash'
+
+objects_prefix='{}/{}/{}={{{{{}}}}}'.format(dataset,
```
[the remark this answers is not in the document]
haroldwoo: I just copy pasted from our load to bq code. Not sure if your way will work with the templating

haroldwoo:
Yes this dag has ran successfully. I've been writing results to a separate bq table and verifying it matches the current one. Also have been doing checks that the parquet data rsync'd back to s3 matches the original s3 parquet data output by emr

willkg:
Is there anything we need to do in socorro for this in the short term?

Longer term, is there anything we should do in socorro to make this "better" in some way?

acmiyaguchi:
@willkg It seems like the largest bit of ugliness in this job is the cross-cloud transfer, being able to read dumps directly from a GCS-accessible location would help quite a bit. Another is to migrate any dashboards or queries away from Athena/Presto and to use the BigQuery dataset instead. This would cut out the data transfer back into AWS.

willkg:
Socorro is in maintenance mode right now, but we're eventually going to move it to GCP. Maybe 2020q1? When we do that, we can fix the cross-cloud transfer silliness. I wrote up a bug to cover that. https://bugzilla.mozilla.org/show_bug.cgi?id=1579266

We don't manage any dashboards or queries that use the socorro_crash data set. Someone would have to track the owners of those down and get them to update. Once this lands, I (or anyone else) can do an announcement on the stability mailing list. I think that'll catch at least some users.

haroldwoo:
If aws_default has write access to telemetry-parquet already, then i'm fine changing to use it. I can change the GKE Pod copy to use the same credentials as well.  I don't think the aws_default keys have access to read the aws prod socorro crash json bucket though

haroldwoo:
I don't think the aws_default connection has any credentials attached. We were relying on the EMR clusters IAM role for permissions before

acmiyaguchi:
Ah right, we're using environment variables to spin up the EMR clusters:

https://github.com/mozilla/telemetry-airflow/blob/ac753516e2deb1ba53bd74aeb739fbc4c4193531/dags/operators/emr_spark_operator.py#L169-L173

I'm still in favor of consolidating the credentials for this job into a single key instead of 2 (or 3, if the variables are a third key instead of a duplicate of one of the other 2 keys).

haroldwoo:
Going to stick with the two separate aws connections to avoid cross account iam permissions, bucket policies, and object acl sadness.

dags/socorro_import.py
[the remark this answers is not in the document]
whd: This particular issue appears to have been resolved, so I'm resolving it.

jobs/socorro_import_crash_data.py
```diff
@@ -0,0 +1,196 @@
...
+Adapted from:
+https://github.com/mozilla-services/data-pipeline/blob/0c94d328f243338d21bae360547c300ac1b82b12/reports/socorro_import/ImportCrashData.ipynb
+
+Original source json data (not ndjson):
+s3://crashstats-telemetry-crashes-prod-us-west-2/v1/crash_report
+
+Original destination parquet data:
+s3://telemetry-parquet/socorro_crash/v2/crash_date=20190801
+
+This job now reads json from GCS, and writes parquet to GCS.
```
whd: Nit: there's some inconsistency in your capitalization of proper nouns and acronyms in your doc strings. I've ignored it elsewhere but this appears to the the first time you've properly capitalized them (I find the capitalization easier to read).

jobs/socorro_import_crash_data.py
```diff
@@ -0,0 +1,196 @@
...
+from datetime import datetime as dt, timedelta, date
+from os import environ
+
+from pyspark.context import SparkContext
+from pyspark.sql.session import SparkSession
+from pyspark.sql.types import *
+
+"""
+Adapted from:
+https://github.com/mozilla-services/data-pipeline/blob/0c94d328f243338d21bae360547c300ac1b82b12/reports/socorro_import/ImportCrashData.ipynb
```
whd: I'm not familiar with this file (though apparently I'm the last person to have edited it heh) but I'm going to assume this is a relatively straight-forward adaption of existing working code, and that this adaption has been tested.

dags/socorro_import.py
whd: Can you add a comment to this affect? As someone not familiar with how we load derived datasets this also confused me.

dags/socorro_import.py
```diff
@@ -18,39 +46,153 @@
...
+            "--dest-gcs-path", "gs://{}/{}".format(gcs_data_bucket, dataset)
+        ],
+        idle_delete_ttl='14400',
+        num_workers=8,
+        worker_machine_type='n1-standard-8',
+        aws_conn_id=read_aws_conn_id,
+        gcp_conn_id=gcp_conn_id)
+)
+
+# TODO - change me later to remove the _2
```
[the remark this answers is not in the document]
haroldwoo: i was just using a different svc account for testing

dags/socorro_import.py
```diff
@@ -18,39 +46,153 @@
...
+bq_gcp_conn_id= 'google_cloud_derived_datasets_2'
+bq_connection = GoogleCloudBaseHook(gcp_conn_id=bq_gcp_conn_id)
+gke_location="us-central1-a"
```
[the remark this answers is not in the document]
haroldwoo: Yes, this cluster already exists there.

dags/socorro_import.py
```diff
@@ -18,39 +46,153 @@
...
+    env_vars={
+        "AWS_ACCESS_KEY_ID": aws_access_key,
+        "AWS_SECRET_ACCESS_KEY": aws_secret_key
+    },
+    dag=dag,
+)
+
+# Not using load_to_bigquery since our source data is on GCS.
+# We do use the parquet2bigquery container to load gcs parquet into bq though.
+# TODO - change the dataset also to telemetry_derived? This is what it is in original socorro job
```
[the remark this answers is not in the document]
haroldwoo: Yeah that's the plan. I was just using the tmp dataset for testing.

jobs/socorro_import_crash_data.py
haroldwoo: Yup it has been tested.

dags/socorro_import.py
```diff
@@ -18,39 +46,153 @@
...
+        aws_conn_id=read_aws_conn_id,
+        gcp_conn_id=gcp_conn_id)
+)
+
+# TODO - change me later to remove the _2
+bq_gcp_conn_id= 'google_cloud_derived_datasets_2'
+bq_connection = GoogleCloudBaseHook(gcp_conn_id=bq_gcp_conn_id)
+gke_location="us-central1-a"
+gke_cluster_name="bq-load-gke-1"
+# TODO - change this to just s3://telemetry-parquet when ready
```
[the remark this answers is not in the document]
haroldwoo: after one more test run, ill change the necessary variables to their prod versions

PR #595, opened 2019-09-06
Refactor glean_clients_daily to clients_daily in fenix datasets

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/337

### DIFF
--- a/dags/direct2parquet_bigquery_load.py
+++ b/dags/direct2parquet_bigquery_load.py
@@ -1,4 +1,5 @@
 from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
 from datetime import datetime, timedelta
 from airflow.operators.subdag_operator import SubDagOperator
 from utils.gcp import bigquery_etl_query, load_to_bigquery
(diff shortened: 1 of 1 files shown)


dags/direct2parquet_bigquery_load.py
```diff
@@ -110,22 +111,32 @@
...
-    glean_clients_daily = bigquery_etl_query(
-        task_id='glean_clients_daily',
-        destination_table='glean_clients_daily_v1',
-        dataset_id='telemetry',
-        start_date=datetime(2019, 7, 1),
+    wait_for_copy_deduplicate = ExternalTaskSensor(
+        task_id="wait_for_copy_deduplicate",
+        external_dag_id="copy_deduplicate",
+        external_task_id="copy_deduplicate_all",
+        execution_delta=timedelta(hours=-1),
```
relud: this should be positive or it won't be pointing at the correct `copy_deduplicate` job, because it's schedule is one hour before this task, and [per the docs](https://airflow.readthedocs.io/en/stable/_modules/airflow/sensors/external_task_sensor.html): `For yesterday, use [positive!] datetime.timedelta(days=1)`
```suggestion
        execution_delta=timedelta(hours=1),
```

jklukas: Do I need an execution delta at all? Reading the docs, it makes it sound like it will default to the job executed on the same day, and makes no mention of the specific timestamp needing to match up.

relud: it's consistent with other implementations to have one, but I honestly have no clue if it's actually necessary.

relud: also, it may only be necessary for same-day deltas where the dependency starts *after* the sensor

PR #607, opened 2019-09-16
Add Glam dag.

emtwo:


emtwo:
> This looks good. Have you been able to run it locally?

Unfortunately I couldn't run the full thing locally because it says I don't have permissions to access the prod data. But I did confirm that jobs are correctly being triggered as expected.

dags/glam.py
```diff
@@ -0,0 +1,148 @@
...
+    external_task_id="copy_deduplicate_main_ping",
+    execution_delta=timedelta(hours=-1),
+    check_existence=True,
+    dag=dag,
+)
+
+clients_daily_scalar_aggregates = bigquery_etl_query(
+    task_id="clients_daily_scalar_aggregates",
+    destination_table="clients_daily_scalar_aggregates_v1",
+    dataset_id=dataset_id,
```
jklukas:
```suggestion
    dataset_id=dataset_id,
    project_id='moz-fx-data-shared-prod',
```

All of the query tasks will need to have project set to `moz-fx-data-shared-prod` so that they write into that project rather than the default derived-datasets project.

dags/glam.py
```diff
@@ -0,0 +1,147 @@
...
+    owner="msamuel@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "msamuel@mozilla.com"],
+    depends_on_past=False,
+    dag=dag)
+
+sql_file_path = "sql/{}/{}/query.sql".format(dataset_id, "clients_daily_keyed_scalar_aggregates_v1")
+clients_daily_keyed_scalar_aggregates = bigquery_etl_query(
+    task_id="clients_daily_keyed_scalar_aggregates",
+    destination_table="clients_daily_scalar_aggregates_v1",
```
jklukas: It looks like the destination tables in a few of these queries don't match the names, so that's probably an error.

Also, because Airflow now has privs to issue query jobs in shared-prod, it shouldn't be necessary to specify the sql_file_path explicitly. If you set project, dataset, and destination_table, it should be able to infer the correct sql_file_path.

emtwo: Actually this was intentional because there are 3 queries that end up into 1 table:
`clients_daily_scalar_aggregates_v1` + `clients_daily_keyed_scalar_aggregates_v1` + `clients_daily_keyed_boolean_aggregates_v1` -> `clients_daily_scalar_aggregates_v1`

And similarly 2 queries go into this other table:
`clients_daily_histogram_aggregates_v1` + `clients_daily_keyed_histogram_aggregates_v1` -> `clients_daily_histogram_aggregates_v1`

Now that I write this out, I realize the naming can be a tad misleading... 
but generally it's because keyed scalars and keyed booleans are processed in a slightly different way than regular scalars so they need their own query but the end result of the aggregation is about the same so it can be included in the same table. Same idea for histograms.

jklukas: Fascinating. And I see now that you set `--noreplace` so that these all append.

A concern I have with this approach is reprocessing. We'd have to manually delete a day before running this DAG again, or we'd end up with duplicate results. It's a nice feature of the default parameters that we generally have ETL queries replace partitions rather than appending, so that reprocessing is always "safe".

Have you considered moving these three queries into a single query file and using `UNION ALL` between them? Or, we could have these be three separate tables with identical schemas and have a single view on top of them that does the UNION. Do either of those sound like reasonable options?

emtwo: Hm. I think this may not be an issue because `clients_daily_scalar_aggregates_v1` runs first and it doesn't have a `--no-replace` tag so it replaces the data in the partition. The other queries append afterwards. Same for histograms. That is my understanding of how this dag will run.

If I'm missing something about how this works, I think I'd rather go with the second option of 3 tables with a view on top.

jklukas: > Hm. I think this may not be an issue because `clients_daily_scalar_aggregates_v1` runs first and it doesn't have a `--no-replace` tag so it replaces the data in the partition. The other queries append afterwards. Same for histograms. That is my understanding of how this dag will run.

That sounds fine. Thanks for running through that. It's probably worth adding a comment above this section explaining the approach, since it may be difficult to infer it from reading through the individual task configs.


> If I'm missing something about how this works, I think I'd rather go with the second option of 3 tables with a view on top because BigQuery tends to give me `complex query` errors, at least in console, when there are too many parts to a query...

That is indeed a very good reason to keep these separated into different queries 👍

dags/glam.py
```diff
@@ -0,0 +1,157 @@
...
+    depends_on_past=True,
+    date_partition_parameter=None,
+    arguments=('--append_table','--noreplace',),
+    dag=dag)
+
+
+wait_for_main_ping >> clients_daily_scalar_aggregates
+clients_daily_scalar_aggregates >> clients_daily_keyed_scalar_aggregates
+clients_daily_keyed_scalar_aggregates >> clients_daily_keyed_boolean_aggregates
+clients_daily_keyed_boolean_aggregates >> clients_scalar_aggregates
```
jklukas: Can these three last steps all run in parallel? If so, you can define them all as dependent on `clients_daily_scalar_aggregates` instead of a chain. That will allow it to run a bit faster.

emtwo: I think we can do:

```
clients_daily_scalar_aggregates >> clients_daily_keyed_scalar_aggregates
clients_daily_scalar_aggregates >> clients_daily_keyed_boolean_aggregates
clients_daily_keyed_boolean_aggregates >> clients_scalar_aggregates
clients_daily_keyed_scalar_aggregates >> clients_scalar_aggregates
```

`clients_daily_scalar_aggregates` should run first to replace a partition. `clients_daily_keyed_scalar_aggregates` and `clients_daily_keyed_boolean_aggregates` can run in parallel afterwards, and `clients_scalar_aggregates` needs to wait for all the data to get into the table its querying from so it needs to wait on those two tasks.

emtwo:
@jklukas updated the PR, let me know if there is anything else.

PR #609, opened 2019-09-16
Kombu 4.6.4 breaks CeleryExecutor.

crankycoder:
Pin the version of kombu to 4.6.3 for py2

See: https://issues.apache.org/jira/browse/AIRFLOW-5240


crankycoder:
@haroldwoo ^

fbertsch:
Added in an alternate PR!

PR #612, opened 2019-09-17
Add remaining D2P datasets for historical import

whd:
These are the remaining datasets outlined in the [import document](https://docs.google.com/document/d/1i578JFxXmCCxxURO5UeWkUchv3uq0MSweHlreh3P6BI/edit#heading=h.l9i8jik4xe4x). They are all very small.

We're not importing these for users to use but rather as the first step in the planned migration of the data into the `_stable` tables.

### DIFF
--- a/dags/direct2parquet_bigquery_load.py
+++ b/dags/direct2parquet_bigquery_load.py
@@ -62,6 +62,21 @@ with DAG(
             'dataset_version': 'v1',
             'bigquery_dataset': 'telemetry',
             },
+        'coverage-coverage-parquet': {
+            'dataset_version': 'v1',
+            },
+        'firefox-launcher-process-launcher-process-failure-parquet': {
+            'dataset_version': 'v1',
+            },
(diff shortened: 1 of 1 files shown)


dags/direct2parquet_bigquery_load.py
```diff
@@ -62,6 +62,21 @@
...
+        'firefox-launcher-process-launcher-process-failure-parquet': {
+            'dataset_version': 'v1',
+            },
+        'pocket-fire-tv-events-parquet': {
+            'dataset_version': 'v1',
+            },
+        'webpagetest-webpagetest-run-parquet': {
+            'dataset_version': 'v1',
+            },
+        'telemetry-ip-privacy-parquet': {
```
whd: This is a static dataset but I've added it here for convenience. In fact, we could potentially just run this script manually once for all these datasets (we only want historical data), but I don't know have experience doing this.

whd:
We're going with a non-airflow solution for this, since the import only needs to happen once.

PR #613, opened 2019-09-18
Bug 1577460 - add BigQuery-based update orphaning ETL script

akkomar:
This adds update-orphaning ETL job adapted from the original notebook[1].
Core logic remains unchanged, there is a BigQuery shim code
(`longitudinal_shim_aggregate` and `longitudinal_shim_transform`) added to
maintain compatibility with legacy Longitudinal schema.

I have verified this job's output [2] against original one scheduled from ATMO [3] - there's pretty good match, differences in calculated metrics do not exceed 0.2%.

[1] https://github.com/mozilla/mozilla-reports/blob/80c9bd6468877e1df1f48208c7c99266bedb9896/projects/app_update_out_of_date.kp/orig_src/app_update_out_of_date.ipynb
[2] s3://telemetry-public-analysis-2/app-update-test/data/out-of-date/20191006.json
[3] s3://telemetry-public-analysis-2/app-update/data/out-of-date/20191006.json

### DIFF
--- a/dags/utils/dataproc.py
+++ b/dags/utils/dataproc.py
@@ -29,6 +29,7 @@ class DataProcHelper:
                  num_preemptible_workers=0,
                  service_account=None,
                  init_actions_uris=None,
+                 additional_metadata=None,
                  optional_components=['ANACONDA'],
                  install_component_gateway=True,
                  aws_conn_id=None,
@@ -82,6 +83,13 @@ class DataProcHelper:
                     elif key == "secret.key":
                         properties["core:fs.s3.awsSecretAccessKey"] = value
 
+        base_metadata = {
+                'gcs-connector-version': '1.9.16',
+                'bigquery-connector-version': '0.13.6'
+            }
+        additional_metadata = {} if additional_metadata is None
+        metadata = {**base_metadata, **additional_metadata}
+
         return DataprocClusterCreateOperator(
             task_id='create_dataproc_cluster',
             cluster_name=self.cluster_name,
@@ -101,10 +109,7 @@ class DataProcHelper:
             optional_components = self.optional_components,
             install_component_gateway = self.install_component_gateway,
             init_actions_uris=self.init_actions_uris,
-            metadata={
-                'gcs-connector-version': '1.9.16',
-                'bigquery-connector-version': '0.13.6'
-            },
+            metadata=metadata,
         )
 
     def delete_cluster(self):
@@ -132,6 +137,7 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
                                 num_preemptible_workers=0,
                                 service_account=None,
                                 init_actions_uris=None,
+                                additional_metadata=None,
                                 optional_components=['ANACONDA'],
                                 install_component_gateway=True,
                                 python_driver_code=None,
@@ -190,6 +196,8 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
                                           account needs the following permissions:
                                           roles/logging.logWriter and roles/storage.objectAdmin.
     :param list init_actions_uris:        List of GCS uri's containing dataproc init scripts.
+    :param dict additional_metadata       Custom metadata keys and values, might be used to
+                                          configure initialization actions.
     :param str job_name:                  Name of the spark job to run.
 
     :param str aws_conn_id:               Airflow connection id for S3 access (if needed).
@@ -221,6 +229,7 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
                                      service_account=service_account,
                                      init_actions_uris=init_actions_uris,
                                      optional_components=optional_components,
+                                     additional_metadata=additional_metadata,
                                      install_component_gateway=install_component_gateway,
                                      aws_conn_id=aws_conn_id,
                                      gcp_conn_id=gcp_conn_id)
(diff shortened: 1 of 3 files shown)


dags/app_update_out_of_date.py
```diff
@@ -0,0 +1,66 @@
...
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from datetime import datetime, timedelta
+
+from utils.dataproc import moz_dataproc_pyspark_runner
+
+"""
+
+The following WTMO connections are needed in order for this job to run:
+conn - google_cloud_airflow_dataproc
+conn - aws_dev_iam_s3
```
akkomar: @haroldwoo is this the correct connection for writing to `s3://telemetry-public-analysis-2`? Or we should rather create one specifically for this job?

dags/utils/dataproc.py
```diff
@@ -132,6 +137,7 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
                                 num_preemptible_workers=0,
                                 service_account=None,
                                 init_actions_uris=None,
+                                additional_metadata=None,
```
akkomar: I have added this to enable defining custom pip dependencies per job.

dags/app_update_out_of_date.py
haroldwoo: Just checked and that aws user only allows rw to telemetry-parquet and a few other buckets. The bucket you mentioned is not one of them.

haroldwoo: Can you create a bug and assign to me for the access required by this job? Also can we name this dag something more descriptive than app_update_out_of_date.py?

akkomar: > Can you create a bug and assign to me for the AWS access required by this job?

https://bugzilla.mozilla.org/show_bug.cgi?id=1590189

> Also can we name this dag something more descriptive than app_update_out_of_date.py?

I'll change this to update_orphaning_dashboard_etl, that should be less cryptic.

akkomar: Connection has been updated.
DAG has been renamed.

jobs/update_orphaning_dashboard_etl.py
```diff
@@ -0,0 +1,1236 @@
...
+
+        res = {}
+        n_hist = len(histograms)
+        for key in keys:
+            res[key] = [0]*n_hist
+            for i, histogram_struct in enumerate(histograms):
+                histogram_array = histogram_struct[histogram_name]
+                if histogram_array:
+                    for hist_key, count_histogram_string in histogram_array:
+                        if hist_key == key:
```
fbertsch: Why not move this to the earlier iterator? Here we're iterating over all histograms to get all the keys; then for each key, iterating over all the histogram's keys again. Seems we could do it in one pass, and this would be clearer.

jobs/update_orphaning_dashboard_etl.py
```diff
@@ -0,0 +1,1236 @@
...
+
+    def merge_enumerated_histograms(histograms, histogram_name, n_values):
+        res = []
+        all_null = True # needed to maintain compatibility with Longitudinal
+        for histogram_string_struct in histograms:
+            compacted_histogram = [0]*(int(n_values)+1)
+            histogram_string = histogram_string_struct[histogram_name]
+            if histogram_string:
+                all_null = False
+                values = json.loads(histogram_string)['values']
```
fbertsch: This strikes me as a likely place of failure for a histogram that is missing `values`. Consider using `.get`

jobs/update_orphaning_dashboard_etl.py
```diff
@@ -0,0 +1,1236 @@
...
+            payload.histograms.update_state_code_complete_startup,
+            payload.histograms.update_state_code_unknown_startup,
+            payload.histograms.update_status_error_code_complete_startup,
+            payload.histograms.update_status_error_code_partial_startup,
+            payload.histograms.update_status_error_code_unknown_startup,
+            payload.histograms.update_status_error_code_complete_stage,
+            payload.histograms.update_status_error_code_partial_stage,
+            payload.histograms.update_status_error_code_unknown_stage
+        FROM
+            `moz-fx-data-shared-prod.static.main_1pct_backfill`
```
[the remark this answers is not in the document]
akkomar: We need to look at both to get the full 6 month view (1% sample table has data from '2019-01-01' to '2019-09-01', full table starts on '2019-09-01'). After backfill is completed we can simplify that to just use full table.

I'll add a comment about that.

jobs/update_orphaning_dashboard_etl.py
```diff
@@ -0,0 +1,1236 @@
...
+min_update_ping_count = 4
+min_subsession_hours = 2
+min_subsession_seconds = min_subsession_hours * 60 * 60
+
+
+# Get the settings based on dates.
+if today_str is None:
+    today_str = start_time.strftime("%Y%m%d")
+
+assert (today_str is not None), "The date environment parameter is missing."
```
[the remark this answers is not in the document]
akkomar: Yeah, these are some leftovers from the original notebook.

akkomar:
> A few comments - have you tested the output of this shim against Longitudinal? For e.g. the length of associated column's arrays, number of clients, etc.

Yes. I'm also pretty confident this is correct because of the match between this job's and original one output. I had some bugs on the way and found out this job was quite sensitive to errors related to (among others) array sizes or default values.

PR #615, opened 2019-09-19
Add execution_delta to wait for copy_deduplicate

jklukas:
Currently, these wait tasks aren't completing. I believe it's due to the 1 hour
difference between the DAG schedules.

See https://github.com/mozilla/telemetry-airflow/pull/595#discussion_r322336248


jklukas:
Went ahead and merged since we previously discussed this change.

PR #617, opened 2019-09-21
Split main_v4 copy_deduplicate into hourly parts

relud:
requires https://github.com/mozilla/bigquery-etl/pull/370


dags/utils/gcp.py
```diff
@@ -472,6 +473,7 @@ def bigquery_etl_copy_deduplicate(
     :param Tuple[str] only_tables:   Only process tables matching the given globs of form 'telemetry_live.main_v*'
     :param Tuple[str] except_tables: Process all tables except those matching the given globs
     :param int parallelism:          Maximum number of queries to execute concurrently
+    :param bool houlry:              Deduplicate one hour at a time, rather than for whole days at once
```
jklukas:
```suggestion
    :param bool hourly:              Deduplicate one hour at a time, rather than for whole days at once
```

PR #546, opened 2019-07-16
WIP - test dataproc on airflow

haroldwoo:
Some things to test:

- [] dataproc cluster create/delete
- [] dataproc can execute aws emr script-runner jar
- [] dataproc clusters with jupyter, component-gateway, etc.
- [] wrap dataprocspark operator with our own dataproc operator similar to how we use our custom emrsparkoperator for easier user migrations
- [] create custom dataproc image for dataproc to use

PR #618, opened 2019-09-23
Run main_summary on dataproc

relud:


dags/utils/gcp.py
```diff
@@ -528,3 +531,152 @@ def normalize_table_id(table_name):
...
+            cluster_name=cluster_name,
+            gcp_conn_id=gcp_conn_id,
+        )
+
+        create_dataproc_cluster >> run_dataproc_spark >> delete_dataproc_cluster
+
+        return dag
+
+@contextmanager
+def dataproc_cluster(
```
haroldwoo: Similar creation/deletion/run logic already exists in dags/utils/dataproc.py (https://github.com/mozilla/telemetry-airflow/pull/546), as well as dags/prio/dataproc.py (https://github.com/mozilla/telemetry-airflow/pull/561). Since we were working on these in parallel they haven't been consolidated.

Its a PITA to retest everything after a refactor, but this PR will add a 3rd dataproc wrapper helper.

relud: thanks for noticing, i'll see about reusing it from one of those

dags/utils/dataproc.py
```diff
@@ -66,6 +66,12 @@ def create_cluster(self):
                 if value is not None:
                     properties["core:fs.s3a." + key] = value
 
+        init_actions_uris = []
+        if self.artifact_bucket:
+            init_actions_uris.append(
+                'gs://{}/bootstrap/dataproc_init.sh'.format(self.artifact_bucket)
+            )
```
relud: r? @haroldwoo

dags/utils/gcp.py
```diff
@@ -135,17 +135,20 @@ def load_to_bigquery(parent_dag_name=None,
...
+                gcs_bucket=gcs_buckets['transfer'],
+                description=_objects_prefix,
+                aws_conn_id=aws_conn_id,
+                gcp_conn_id=gcp_conn_id,
+                project_id=connection.project_id,
+                object_conditions=gcstj_object_conditions,
+                transfer_options=gcstj_transfer_options,
             )
+        else:
+            s3_to_gcs = DummyOperator(task_id='no_s3_to_gcs')
```
relud: r? @jasonthomas

dags/main_summary.py
```diff
@@ -43,6 +45,79 @@
...
+main_summary_dataproc_bigquery_load = SubDagOperator(
+    subdag=load_to_bigquery(
+        parent_dag_name=dag.dag_id,
+        dag_name="main_summary_dataproc_bigquery_load",
+        default_args=default_args,
+        dataset_gcs_bucket="moz-fx-data-derived-datasets-parquet-tmp",
+        dataset="main_summary",
+        dataset_version="v4",
+        gke_cluster_name="bq-load-gke-1",
+        bigquery_dataset="backfill",
```
relud: @jasonthomas is there a better dataset I can use here until this is ready to replace the existing bigquery import?

dags/main_summary.py
```diff
@@ -43,6 +45,79 @@
...
+        optional_components=[],
+        install_component_gateway=False,
+        jar_urls=[
+            "s3a://net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla/telemetry-batch-view/master/telemetry-batch-view.jar"
+        ],
+        main_class="com.mozilla.telemetry.views.MainSummaryView",
+        jar_args=[
+            "--from={{ds_nodash}}",
+            "--to={{ds_nodash}}"
+            "--bucket=gs://moz-fx-data-derived-datasets-parquet-tmp",
```
relud: @jasonthomas is there a different bucket I should be using here instead of `gs://moz-fx-data-derived-datasets-parquet-tmp` that won't interference with the production import of main summary in `gs://moz-fx-data-derived-datasets-parquet`?

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+        optional_components=[],
+        install_component_gateway=False,
+        jar_urls=[
+            "s3a://net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla/telemetry-batch-view/master/telemetry-batch-view.jar",
+        ],
+        main_class="com.mozilla.telemetry.views.MainSummaryView",
+        jar_args=[
+            "--from={{ds_nodash}}",
+            "--to={{ds_nodash}}",
+            "--bucket=" + main_summary_dataproc_bucket,
```
relud: @jasonthomas is there a different bucket I should be using here instead of `gs://moz-fx-data-derived-datasets-parquet-tmp` that won't interference with the production import of main summary in `gs://moz-fx-data-derived-datasets-parquet`?

dags/utils/gcp.py
```diff
@@ -528,3 +531,55 @@ def normalize_table_id(table_name):
...
+    :param str gke_location:       GKE cluster location
+    :param str gke_cluster_name:   GKE cluster name
+    :param str gke_namespace:      GKE cluster namespace
+    :param str image_pull_policy:  Kubernetes policy for when to pull
+                                   docker_image
+    :param Dict[str, Any] kwargs:  Additional keyword arguments for
+                                   GKEPodOperator
+
+    :return: GKEPodOperator
+    """
```
relud: r? @jasonthomas

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+main_summary_dataproc_bucket = "gs://moz-fx-data-derived-datasets-parquet-tmp"
+main_ping_bigquery_export_prefix = main_summary_dataproc_bucket + "/export"
+main_ping_bigquery_export_dest = main_ping_bigquery_export_prefix + "/submission_date={{ds}}/document_namespace=telemetry/document_type=main/document_version=4/*.avro"  # noqa
+main_ping_bigquery_export = gke_command(
+    task_id="main_ping_bigquery_extract",
+    command=[
+        "bq",
+        "extract",
+        "--destination_format=AVRO",
+        "moz-fx-data-shared-prod.payload_bytes_decoded.telemetry_telemetry__main_v4${{ds_no_dash}}",
```
haroldwoo: shouldn't this be {{ ds_nodash }}

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+main_summary_dataproc = SubDagOperator(
+    subdag=moz_dataproc_jar_runner(
+        parent_dag_name="main_summary",
+        dag_name="main_summary_dataproc",
+        default_args=default_args,
+        cluster_name="main-summary-{{ds}}",
+        image_version="1.3",
+        idle_delete_ttl="600",
+        worker_machine_type="n1-standard-8",
+        num_preemptible_workers=40,
```
haroldwoo: How'd you come up with this number?

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+        optional_components=[],
+        install_component_gateway=False,
+        jar_urls=[
+            "s3a://net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla/telemetry-batch-view/master/telemetry-batch-view.jar",
```
haroldwoo: I'm assuming this is s3 bucket is public, since I don't see any aws credentials configured

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+    docker_image=gcloud_docker_image,
+)
+
+main_summary_dataproc_s3_copy = gke_command(
+    task_id="main_summary_dataproc_s3_copy",
+    command=[
+        "gsutil",
+        "-m",
+        "rsync",
+        main_summary_dataproc_bucket + "/main_summary/v4/submission_date_s3={{ds_no_dash}}",
```
haroldwoo: does this work? or should it be {{ ds_nodash }}

dags/utils/dataproc.py
haroldwoo: I am addressing this in a PR i have for review soon: https://github.com/mozilla/telemetry-airflow/pull/620/files.

By adding init_action_uris as a parameter rather than relying on self.artifact_bucket being set

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+    docker_image=gcloud_docker_image,
+)
+
+main_summary_dataproc = SubDagOperator(
+    subdag=moz_dataproc_jar_runner(
+        parent_dag_name="main_summary",
+        dag_name="main_summary_dataproc",
+        default_args=default_args,
+        cluster_name="main-summary-{{ds}}",
+        image_version="1.3",
```
haroldwoo: I'm assuming this is because you need spark 2.3?

dags/main_summary.py
haroldwoo: At some point we'll need to move the CI and location to GCS?

dags/main_summary.py
relud: the databricks version of main summary uses `max_instance_count=40`

dags/main_summary.py
relud: yep. the default of `image_version="1.4"` doesn't work because telemetry-batch-view jars are compiled against spark 2.3. Compiling them against spark 2.4 also fixes the issue, but I don't want to block this on upgrading spark for t-b-v.

dags/main_summary.py
relud: >_< should be `ds_nodash`, fixing...

dags/main_summary.py
relud: > I'm assuming this is s3 bucket is public, since I don't see any aws credentials configured

yes

> At some point we'll need to move the CI and location to GCS?

at some point, yes

dags/main_summary.py

dags/main_summary.py
akkomar: That's fine for now, we should check if this is optimal later.

dags/main_summary.py
```diff
@@ -43,6 +45,94 @@
...
+            "s3a://net-mozaws-data-us-west-2-ops-ci-artifacts/mozilla/telemetry-batch-view/master/telemetry-batch-view.jar",
+        ],
+        main_class="com.mozilla.telemetry.views.MainSummaryView",
+        jar_args=[
+            "--from={{ds_nodash}}",
+            "--to={{ds_nodash}}",
+            "--bucket=" + main_summary_dataproc_bucket,
+            "--export-path=" + main_ping_bigquery_export_prefix,
+        ],
+        artifact_bucket=None,
```
haroldwoo: Did you make this change because the dataproc init script on master breaks your spark job?

relud: i made this change because I would like to be able to run without an init script.

dags/utils/dataproc.py
haroldwoo: This is fine, when I commit my PR ill make sure that your code changes to use an empty array for init_action_urls

dags/main_summary.py
jasonthomas: In what sense do mean by better? The `backfill` dataset is a temporary staging area that I've used in the past but if we prefer to use a different dataset I am not against it, we would just need to provision it.

dags/main_summary.py
jasonthomas: We can create another bucket just to make sure if we don't interfere here.

dags/main_summary.py
relud: just checking that you think it's an acceptable dataset to use as a temporary staging area for this.

dags/utils/gcp.py
jasonthomas: lgtm

dags/utils/gcp.py
jasonthomas: lgtm

dags/main_summary.py
jasonthomas: We discussed on slack and we will continue this bucket since we won't interfere with existing workloads.

relud:
> There are some questions about which big query dataset and bucket to use and if those change we will need to provision them.

Let's use the ones currently configured here.

> We will also need to make sure that the dataproc clusters (since they live in another project) have the correct access to read/write into the gcs locations. We also need to verify that the `aws_dev_iam_s3` connection has write access to the `s3://telemetry-parquet` bucket. I believe in the past we've restricted it by prefix.

In that case I would like to merge this as-is, and we can figure out permissions without blocking this PR. For that I need you to leave an github r+

jasonthomas:
I've updated `aws_dev_iam_s3` connection id aws crededntials to include `s3://telemetry-parquet/main_summary_dataproc/*`

PR #620, opened 2019-09-24
Migrate Fx usage report from emr to dataproc

haroldwoo:
Notes:
This job originally read main_summary from s3://telemetry-parquet in the AWS Dev account and wrote data to s3://net-mozaws-prod-us-west-2-data-public in the AWS Prod account which is hosted by cloudfront.

This job now reads parquet data from GCS and still writes to the same location in s3.  When the spark/bigquery main summary access has been resolved, we can modify this job to read from bq.

Currently a few YAUs are slightly higher for dataproc on gcs parquet data. This is not a KPI so we will leave it at that.

TODO - 
- [x] Figure out why dataproc and emr results have slightly different results for YAU. The full parquet data (back to 2016-03-12 didn't exist in GCP yet. After adding all the data back, there are still 30 diffs in the result json (~15876 lines), which means < 15 YAU differences since they are paired with slight avg_intensity and avg_daily_usage(hours) diffs. I've noticed about 350-480 greater YAU in the dataproc runs, using GCS parquet data that we use to load into bq.
- [x] All todos in code - change start date, and maybe followup PR for CI of custom bootstrap script
- [x]  rebase with master after https://github.com/mozilla/telemetry-airflow/pull/618/files merges, and make sure to change main_summary.py where artifact_bucket=None to set init_action_urls = [] instead.
- [x] Add aws_prod_fx_usage_report connection to wtmo

### DIFF
--- a/dags/fx_usage_report.py
+++ b/dags/fx_usage_report.py
@@ -1,13 +1,18 @@
 from airflow import DAG
 from airflow.operators.sensors import ExternalTaskSensor
 from datetime import datetime, timedelta
-from operators.emr_spark_operator import EMRSparkOperator
+
+from airflow.operators.subdag_operator import SubDagOperator
+from airflow.contrib.hooks.aws_hook import AwsHook
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+
 from utils.constants import DS_WEEKLY
+from utils.dataproc import moz_dataproc_scriptrunner
 
 default_args = {
     'owner': 'frank@mozilla.com',
     'depends_on_past': True,
-    'start_date': datetime(2018, 11, 25),
+    'start_date': datetime(2019, 9, 30),
     'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
     'email_on_failure': True,
     'email_on_retry': True,
@@ -17,6 +22,7 @@ default_args = {
 
 dag = DAG('fx_usage_report', default_args=default_args, schedule_interval='@weekly')
 
+
 wait_for_main_summary = ExternalTaskSensor(
     task_id='wait_for_main_summary',
     external_dag_id='main_summary',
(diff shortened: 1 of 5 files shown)


dags/fx_usage_report.py
```diff
@@ -17,24 +22,51 @@
...
+             "AWS_ACCESS_KEY_ID": aws_access_key,
+             "AWS_SECRET_ACCESS_KEY": aws_secret_key
+        },
+        gcp_conn_id=gcp_conn_id,
+        # This should is used to set the s3a configs for read/write to s3 for non boto calls
+        aws_conn_id=aws_conn_id,
+        num_workers=9,
+        worker_machine_type='n1-standard-16',
+        image_version='1.3',
+        init_actions_uris=['gs://moz-fx-data-prod-airflow-dataproc-artifacts/bootstrap/fx_usage_init.sh'],
```
whd: This seems to be a bit of a leaky interface i.e. we're already exposing `artifact_bucket` as a config parameter and setting the default bootstrap to a path within that. Maybe it would be better to have this config option be e.g. `bootstrap_actions` and by convention make these values object paths within the artifact bucket (perhaps also prefixed by `bootstrap/`). As-is this is of course more general, and appears to map to the upstream version.

dags/main_summary.py
```diff
@@ -85,7 +85,7 @@
             "--export-path=" + main_ping_bigquery_export_prefix,
         ],
         job_name="main_summary_view_{{ds_nodash}}",
-        artifact_bucket=None,
+        init_action_urls=[],
```
whd:
```suggestion
        init_actions_uris=[],
```

dags/utils/dataproc.py
```diff
@@ -29,6 +29,7 @@ def __init__(self,
                  num_preemptible_workers=0,
                  service_account=None,
                  artifact_bucket='moz-fx-data-prod-airflow-dataproc-artifacts',
+                 init_actions_uris=None,
```
whd: Semantically this is a bit confusing but I suppose the purpose is to differentiate between setting a default value based on `artifact_bucket` and explicitly setting no init actions via `[]`. It also appears to match the upstream.

dags/utils/dataproc.py
```diff
@@ -45,6 +46,12 @@ def __init__(self,
         self.num_preemptible_workers = num_preemptible_workers
         self.service_account = service_account
         self.artifact_bucket = artifact_bucket
+
+        if init_actions_uris is None:
+            self.init_actions_uris=['gs://{}/bootstrap/dataproc_init.sh'.format(self.artifact_bucket)]
```
whd: See earlier comment on leakiness.

dags/utils/dataproc.py
```diff
@@ -56,21 +63,24 @@ def create_cluster(self):
         """
         Returns a DataprocClusterCreateOperator
         """
-        # Set hadoop properties to access s3 from dataproc
         properties = {}
+
+        # Google cloud storage requires object.create permission when reading from pyspark
+        properties["core:fs.gs.implicit.dir.repair.enable"] = "false"
```
whd: I have no familiarity with this and I'm going to assume it's correct.

dags/main_summary.py
haroldwoo: woops thanks

haroldwoo:
The TODOs were mostly variables set to values for nonprod testing, that I am reminding myself to change back before merging

dataproc_bootstrap/fx_usage_init.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -exo pipefail
+
+# Logs will be available on the dataproc nodes at /var/log/dataproc-initialization-script-X.log
+# or via the GCP Dataproc UI
+
+/opt/conda/default/bin/pip install --upgrade pip
+
+/opt/conda/default/bin/pip install arrow==0.10.0
```
fbertsch: Can we use the requirements files from the repo instead? Then we're guaranteed to have the same dependencies between dev and prod.

dags/utils/dataproc.py
fbertsch: I agree, this is confusing that there is a default bootstrap when you don't explicitly set one (and this default to `None`).

dataproc_bootstrap/fx_usage_init.sh
haroldwoo: The repo's dockerfile and the actual libraries on the emr hosts that the job runs on are different. I went with the later

haroldwoo: also there were some changes when test iterating on dataproc. So the init script here is version i got to work

dags/utils/dataproc.py
haroldwoo: currently nobody is setting the artifact_bucket parameter. Would you prefer I remove it as a parameter entirely? And if init_action_uris is None, set it to the default init script(within that bucket). Otherwise they can set it to [].

haroldwoo: The only other place it is currently used is to point to the scriptrunner jar and init scripts that mimic the custom emrsparkoperator functionality

haroldwoo: removed artifact_bucket as a param

haroldwoo: the confusion is likely because a previous pr was using artifact_bucket=None to set no init actions rather than setting init actions itself (which didnt exist as a param before)

dataproc_bootstrap/fx_usage_init.sh
fbertsch: We should go with the former (and it looks like most of these are the former?). It makes it easier when testing and making changes to match dev.

haroldwoo: I disagree. We should go with the actual working instance on prod. Plus the dev container is way faster to iterate on than actually running the workload on dataproc

haroldwoo:
Closing out. A follow up is to modify the Fx usage report repo's docker container to match dataproc's environment as closely as possible. This was out of the scope of this PR since the dev container is not ran anywhere but tests, and currently didn't match EMR's env anyway.

PR #628, opened 2019-10-03
Features/gcp taar amodump

crankycoder:
These are patches to port the taar_amodump DAG to use GCP.

The first few jobs do not require Spark:

  * taar_amodump
  * taar_amo_whitelist
  * taar_update_whitelist

The last job requires PySpark:

  * taar_lite 

Of note - 3 Airflow connections:

GCP connections: `google_cloud_derived_datasets_3` and `google_cloud_airflow_dataproc`.

AWS connection: `airflow_taar_rw_s3`

This PR requires a patch to https://github.com/mozilla/taar_gcp_etl/ so that the taar_lite_guidguid.py script is pushed into a GCS bucket on a tagged build.

### DIFF
--- a/dags/taar_amodump.py
+++ b/dags/taar_amodump.py
@@ -108,8 +108,8 @@ taar_lite = SubDagOperator(
         default_args=default_args,
         cluster_name=taarlite_cluster_name,
         job_name="TAAR_Lite_GUID_GUID",
-        # python_driver_code="gs://moz-fx-data-prod-airflow-dataproc-artifacts/jobs/taar_lite_guidguid.py",
-        python_driver_code="gs://temp-hwoo-removemelater/taar_lite_guidguid.py",
+        python_driver_code="gs://moz-fx-data-prod-airflow-dataproc-artifacts/jobs/taar_lite_guidguid.py",
+        # python_driver_code="gs://temp-hwoo-removemelater/taar_lite_guidguid.py",
         num_workers=8,
         py_args=[
             "--date",
--- a/dags/taar_amodump.py
+++ b/dags/taar_amodump.py
@@ -1,83 +1,128 @@
-from airflow import DAG
 from datetime import datetime, timedelta
 
-from operators.emr_spark_operator import EMRSparkOperator
+from airflow import DAG
+from airflow.contrib.hooks.aws_hook import AwsHook
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.operators.subdag_operator import SubDagOperator
 
-from utils.mozetl import mozetl_envvar
+from operators.gcp_container_operator import GKEPodOperator  # noqa
+from utils.dataproc import moz_dataproc_pyspark_runner
+
+gke_cluster_name = "bq-load-gke-1"
+
+taarlite_cluster_name = "dataproc-taarlite-guidguid"
+
+# Defined in Airflow's UI -> Admin -> Connections
+gcp_conn_id = "google_cloud_derived_datasets"
+connection = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id)
+
+# Dataproc connection to GCP
+gcpdataproc_conn_id = "google_cloud_airflow_dataproc"
+
+
+aws_conn_id = "airflow_taar_rw_s3"
+aws_access_key, aws_secret_key, session = AwsHook(aws_conn_id).get_credentials()
 
 default_args = {
-    'owner': 'amiyaguchi@mozilla.com',
-    'depends_on_past': False,
-    'start_date': datetime(2018, 11, 26),
-    'email': ['telemetry-alerts@mozilla.com', 'amiyaguchi@mozilla.com'],
-    'email_on_failure': True,
-    'email_on_retry': True,
-    'retries': 3,
-    'retry_delay': timedelta(minutes=30),
+    "owner": "vng@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2019, 10, 7),
+    "email": ["telemetry-alerts@mozilla.com", "amiyaguchi@mozilla.com"],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
 }
 
-dag = DAG('taar_amodump', default_args=default_args, schedule_interval='@daily')
+dag = DAG("taar_amodump", default_args=default_args, schedule_interval="@daily")
 
-amodump = EMRSparkOperator(
+amodump = GKEPodOperator(
     task_id="taar_amodump",
-    job_name="Dump AMO JSON blobs with oldest creation date",
-    execution_timeout=timedelta(hours=1),
-    instance_count=1,
+    gcp_conn_id=gcp_conn_id,
+    project_id=connection.project_id,
+    location="us-central1-a",
+    cluster_name=gke_cluster_name,
+    name="taar-amodump",
+    namespace="default",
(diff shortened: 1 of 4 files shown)


dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
-from operators.emr_spark_operator import EMRSparkOperator
+from airflow import DAG
+from airflow.contrib.hooks.aws_hook import AwsHook
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.operators.subdag_operator import SubDagOperator
+
+from operators.gcp_container_operator import GKEPodOperator  # noqa
+from utils.dataproc import moz_dataproc_pyspark_runner
+
+# Unsalted cluster name so subsequent runs fail if the cluster name exists
```
haroldwoo: Rm comment. This is a long lived cluster

dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
+cluster_name = "bq-load-gke-1"
+
+taarlite_cluster_name = "taarlite-guidguid"
```
haroldwoo: Maybe add dataproc to the name

dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
 
-from utils.mozetl import mozetl_envvar
+# Defined in Airflow's UI -> Admin -> Connections
+gcp_conn_id = "google_cloud_derived_datasets_3"
```
haroldwoo: Rm the _3

dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
-    job_name="Dump AMO JSON blobs with oldest creation date",
-    execution_timeout=timedelta(hours=1),
-    instance_count=1,
+    gcp_conn_id=gcp_conn_id,
+    project_id=connection.project_id,
+    location="us-central1-a",
+    cluster_name="bq-load-gke-1",
+    name="taar-amodump",
+    namespace="default",
+    image="gcr.io/moz-fx-data-airflow-prod-88e0/taar_gcp_etl:0.1",
```
haroldwoo: Comment to link to external github repo here would be nice

dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
     owner="vng@mozilla.com",
-    email=["mlopatka@mozilla.com", "vng@mozilla.com"],
-    env=mozetl_envvar("taar_amowhitelist",
-                      {},
-                      {'MOZETL_SUBMISSION_METHOD': 'spark'}),
-    uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-submit.sh",
-    output_visibility="private",
-    dag=dag
+    email=["mlopatka@mozilla.com", "vng@mozilla.com", "hwoo@mozilla.com"],
+    arguments=["-m", "taar_etl.taar_amowhitelist"],
```
haroldwoo: No date req?

dags/taar_amodump.py
```diff
@@ -1,83 +1,122 @@
...
-    uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-submit.sh",
-    output_visibility="private",
-    dag=dag
+    subdag=moz_dataproc_pyspark_runner(
+        parent_dag_name="taar_amodump",
+        dag_name="taar_lite",
+        default_args=default_args,
+        cluster_name=taarlite_cluster_name,
+        job_name="TAAR_Lite_GUID_GUID",
+        python_driver_code="gs://temp-hwoo-removemelater/taar_lite_guidguid.py",
```
haroldwoo: This will need to change prior to merge

dags/taar_amodump.py
crankycoder: Comment deleted.

dags/taar_amodump.py
crankycoder: renamed to `dataproc-taarlite-guidguid`

dags/taar_amodump.py
crankycoder: dropped the `_3`

dags/taar_amodump.py
crankycoder: I've added comments to each of the jobs to point to the repo now

dags/taar_amodump.py
crankycoder: Nope. But added comments about it.  

The job just crawls the AMO API site to build a list of available addons.  Adding a date parameter wouldn't give us a real date scoped list of addons as AMO is always going to give us the latest available addon list.

dags/taar_amodump.py
crankycoder: @haroldwoo Can you help me get this integrated into CircleCI so that tagged builds automatically push to a GCS bucket?

haroldwoo: as long as the script doesnt have any secrets, you can put a copy of the script in this repo under the jobs folder, and itll be sync'd to gs://moz-fx-data-prod-airflow-dataproc-artifacts/jobs after this PR merges.

https://github.com/mozilla/telemetry-airflow/blob/master/.circleci/config.yml#L74

crankycoder: Thanks - I've put a copy of the script into the data directory in 09d9487

crankycoder:
@haroldwoo r?  I've put the taar_lite_guidguid script into the jobs directory per your last suggestion.

dags/taar_amodump.py
```diff
@@ -1,83 +1,127 @@
...
-    output_visibility="private",
-    dag=dag
+    subdag=moz_dataproc_pyspark_runner(
+        parent_dag_name="taar_amodump",
+        dag_name="taar_lite",
+        default_args=default_args,
+        cluster_name=taarlite_cluster_name,
+        job_name="TAAR_Lite_GUID_GUID",
+        python_driver_code="gs://moz-fx-data-prod-airflow-dataproc-artifacts/jobs/taar_lite_guidguid.py",
+        num_workers=15,
```
crankycoder: @haroldwoo @mreid-moz is this ok?  I bumped the number of workers up to 15 so that the job completes in ~30 minutes.  Is there a cost optimization that I should think about?  This job needs to run daily.

dags/taar_amodump.py
```diff
@@ -1,83 +1,127 @@
...
-from operators.emr_spark_operator import EMRSparkOperator
+from airflow import DAG
+from airflow.contrib.hooks.aws_hook import AwsHook
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.operators.subdag_operator import SubDagOperator
+
+from operators.gcp_container_operator import GKEPodOperator  # noqa
+from utils.dataproc import moz_dataproc_pyspark_runner
+
+cluster_name = "bq-load-gke-1"
```
haroldwoo: I would rename this variable to something like gke_cluster_name to avoid confusion. Then use the variable on lines 44, 64, 86

dags/taar_amodump.py
haroldwoo: Should be fine. Does this job take roughly twice the time with twice as many workers?

crankycoder:
This DAG runs successfully for me on a local airflow instance.

Airflow_-_DAGs

I've squashed and rebased to clarify the changes.  I've left the switch from test GCS and production GCS file paths in df629ae and 24f572e as I can't test the production GCS path on my local airflow instance.

crankycoder:
@haroldwoo 33d99b0a explicitly removes the `google_cloud_derived_datasets` connection as it causes surprising failures during test as the connection is defined with empty values for all connection parameters.

PR #630, opened 2019-10-08
Add fxa_users_services queries

jklukas:
Schedules new queries introduced in https://github.com/mozilla/bigquery-etl/pull/396

### DIFF
--- a/dags/fxa_events.py
+++ b/dags/fxa_events.py
@@ -96,3 +96,46 @@ with models.DAG(
     )
 
     fxa_users_last_seen >> smoot_usage_fxa_v2
+
+
+    # Per-user-per-service tables.
+
+    fxa_users_services_daily = bigquery_etl_query(
+        task_id='fxa_users_services_daily',
+        destination_table='moz-fx-data-shared-prod:telemetry_derived.fxa_users_services_daily_v1',
+        sql_file_path='sql/telemetry_derived/fxa_users_services_daily_v1/query.sql',
+        dataset_id='telemetry_derived',
+    )
+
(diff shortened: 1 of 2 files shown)


dags/fxa_events.py
```diff
@@ -96,3 +96,46 @@
...
+
+    fxa_users_services_daily << fxa_auth_events
+    fxa_users_services_daily << fxa_auth_bounce_events
+    fxa_users_services_daily << fxa_content_events
+    fxa_users_services_daily << fxa_oauth_events
+
+    fxa_users_services_first_seen = bigquery_etl_query(
+        task_id='fxa_users_services_first_seen',
+        destination_table='moz-fx-data-shared-prod:telemetry_derived.fxa_users_services_first_seen_v1',
+        sql_file_path='sql/telemetry_derived/fxa_users_services_first_seen_v1/query.sql',
```
relud: `init.sql` uses DDL that creates the table, and i'm not sure how that will behave in conjunction with `destination_table`, but `query.sql` doesn't exist
```suggestion
        sql_file_path='sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql',
```

jklukas: I'm pretty sure a query will fail if it contains DDL and is configured with a destination table, so I will need to either switch init.sql to not use DDL or will need some new machinery here.

relud: confirmed, it will fail

jklukas:
Updated the machinery to allow DDL queries. Re-requesting review.

PR #636, opened 2019-10-14
Add leanplum export job

fbertsch:
- Fennec nightly export job (will add the new vars myself)
- Leanplum export job (for use with dockerhub)

This depends on bug 1588174.


dags/leanplum_export.py
```diff
@@ -0,0 +1,33 @@
...
+         schedule_interval='@daily') as dag:
+
+    app_id = Variable.get('LEANPLUM_FENNEC_NIGHTLY_APP_ID')
+    client_key = Variable.get('LEANPLUM_FENNEC_NIGHTLY_CLIENT_KEY')
+
+    fennec_nightly_export = leanplum.export(
+        task_id='fennec_nightly_export',
+        leanplum_app_id=app_id,
+        leanplum_client_key=client_key,
+        gcs_bucket='moz-fx-data-leanplum-export',
```
fbertsch: This bucket doesn't exist. Which bucket it is doesn't matter, we could add a general "external datasets" bucket and have the prefix be leanplum-specific.

dags/leanplum_export.py
```diff
@@ -0,0 +1,33 @@
...
+        gcs_prefix='firefox-android-dev',
+        bq_dataset_id='firefox_android_dev_leanplum',
```
fbertsch: This dataset should be specific for a single Leanplum application. It will need to be created before running this job.

dags/utils/leanplum.py
```diff
@@ -0,0 +1,67 @@
...
+    return GKEPodOperator(
+        task_id=task_id,
+        gcp_conn_id=gcp_conn_id,
+        project_id=project_id,
+        location=gke_location,
+        cluster_name=gke_cluster_name,
+        namespace=gke_namespace,
+        image=docker_image,
+        image_pull_policy=image_pull_policy,
+        arguments=["leanplum-data-export"]
```
haroldwoo: why are we appending lists as opposed to using one?

dags/leanplum_export.py
```diff
@@ -0,0 +1,33 @@
...
+    'retries': 2,
+    'retry_delay': timedelta(minutes=30),
+}
+
+
+with DAG('leanplum_export',
+         default_args=default_args,
+         schedule_interval='@daily') as dag:
+
+    app_id = Variable.get('LEANPLUM_FENNEC_NIGHTLY_APP_ID')
```
haroldwoo: Can you add these to wtmo?

haroldwoo: Also, I'd need these to add to our secret hiera sops repo

bin/run
```diff
@@ -50,6 +50,22 @@ init_connections() {
     airflow connections --add \
         --conn_id amplitude_s3_conn \
         --conn_type s3
+
+    airflow connections --add \
```
haroldwoo: I'd prefer not to add these blank connections. When testing locally and not overriding these, the error messages were not obvious. E.g. I saw "TypeError: execv() arg 2 must contain only strings" when this conn wasn't overwritten

dags/utils/leanplum.py
fbertsch: No particular reason, I'm amenable to a single list

dags/leanplum_export.py
fbertsch: I'll add them to WTMO and send them to you

bin/run
fbertsch: It's a bit alarming to run dev and have a bunch of errors, plus IMHO it makes it easier to change them if they are already present. However that error is indeed not clear, any idea what it came from? I haven't seen anything like that.

haroldwoo: It comes when trying to run GKEPodOperator using the blank connection defined here

dags/leanplum_export.py
haroldwoo: is there a bug to create the gcs bucket and bq dataset? We would need to know which project, and potentially give the derived dataset svc account access to that bucket if it lives in another project

fbertsch: There is not. I don't care necessarily where they are located so long as it is consistent.

fbertsch: Bug has been filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1588654

bin/run
fbertsch: Yeah, that's definitely not great. It doesn't look like we can add non-empty `google_cloud_platform` connections via the CLI. I'll open an issue in Airflow.

haroldwoo: or add dummy records?

fbertsch: Yeah that's what I was trying to do, but there's no CLI spec for `json-key` as one of the connection variables.

PR #645, opened 2019-10-21
Leanplum project

fbertsch:
It tries to create the tables in `derived-datasets` otherwise. Depends on https://github.com/mozilla/leanplum-data-export/pull/6.


dags/utils/leanplum.py
```diff
@@ -54,7 +56,8 @@ def export(
             "--client-key", leanplum_client_key,
             "--date", "{{ ds_nodash }}",
             "--bucket", gcs_bucket,
-            "--bq-dataset", bq_dataset_id]
+            "--bq-dataset", bq_dataset_id
```
akkomar: Missing comma:
```suggestion
            "--bq-dataset", bq_dataset_id,
```

PR #652, opened 2019-10-23
decommission desktop_active_dau

relud:
this is actively used by one query https://sql.telemetry.mozilla.org/queries/52720/source which needs to be updated to pick up the changes at https://github.com/mozilla/stmo_core_product_metrics/pull/21


akkomar:
> this is actively used by one query https://sql.telemetry.mozilla.org/queries/52045/source#138916 

You mean this one: https://sql.telemetry.mozilla.org/queries/52720/source ? @mreid-moz - do you need this query?

relud:
oops, my bad. yes, that one.

akkomar:
> > this is actively used by one query https://sql.telemetry.mozilla.org/queries/52045/source#138916
> 
> You mean this one: https://sql.telemetry.mozilla.org/queries/52720/source ? @mreid-moz - do you need this query?

This query is used in the [dashboard](https://sql.telemetry.mozilla.org/dashboard/mdc-required-container-dashboard-for-dau) linked from MDC. @mreid-moz emailed MDC team about this today.

mreid-moz:
I think it should be safe to switch 52720 to use the more modern "exact mau" tables for this query, though I would prefer to drop it altogether if I can verify that they're no longer needed in MDC.

mreid-moz:
I checked, and this query _is_ linked to from MDC, so we should update it to used the `exact_*` tables.

relud:
I double-checked the results of the query before and after using exact active mau, and the graphs look almost identical, exact is just smoother, so i went ahead and updated the query. this PR should now be safe to merge.

PR #653, opened 2019-10-23
Dataproc operator fix

akkomar:
This fixes a bug in DataprocClusterCreateOperator resulting in `image_version` property being cleared during nested dictionary update.

I also silenced email notifications from update orphaning job until everything is stable.

### DIFF
--- a/dags/operators/moz_dataproc_operator.py
+++ b/dags/operators/moz_dataproc_operator.py
@@ -283,7 +283,10 @@ class DataprocClusterCreateOperator(DataprocOperationBaseOperator):
 
         # Fetch current nested dict and add nested keys
         cluster_config_new = cluster_data['config']
-        cluster_config_new.update({'softwareConfig' : {'optionalComponents': self.optional_components }})
+
+        software_config_new = cluster_config_new['softwareConfig']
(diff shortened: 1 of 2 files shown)


dags/operators/moz_dataproc_operator.py
```diff
@@ -283,7 +283,10 @@ def _build_gce_cluster_config(self, cluster_data):
 
         # Fetch current nested dict and add nested keys
         cluster_config_new = cluster_data['config']
-        cluster_config_new.update({'softwareConfig' : {'optionalComponents': self.optional_components }})
+
+        software_config_new = cluster_config_new['softwareConfig']
```
haroldwoo: What if this key does not exist?

akkomar: It will exist, whole dictionary is created here: https://github.com/mozilla/telemetry-airflow/blob/d2257dcf52b8ca5f847b53463b4e96c94e3015b7/dags/operators/moz_dataproc_operator.py#L387

Granted, this isn't ideal but it's consistent with the rest of this file.

akkomar:
> lgtm, I'm assuming you've tested this?

Yes, I tested it on update orphaning dag and confirmed that the cluster is created with the correct version.

PR #655, opened 2019-10-24
Set default service account for dataproc clusters

haroldwoo:


dags/utils/dataproc.py
```diff
@@ -44,7 +44,12 @@ def __init__(self,
         self.master_machine_type = master_machine_type
         self.worker_machine_type = worker_machine_type
         self.num_preemptible_workers = num_preemptible_workers
-        self.service_account = service_account
+
+        if service_account is None:
+            self.service_account = 'dataproc-runner-prod@airflow-dataproc.iam.gserviceaccount.com'
```
jasonthomas: Why not just change the default value of the `service_account` argument from None to this service account?

haroldwoo: Because the spark runners that actually initialize this helper will pass None as svc account, rather than omitting the parameter

haroldwoo: I could add it as a default to the runners instead, e.g. https://github.com/mozilla/telemetry-airflow/blob/23a2b44713ee91e08070e797e6736057917114eb/dags/utils/dataproc.py#L147

jasonthomas: Okay I understand. I think modifying it at the runner instead would be better option if for whatever reason we decide to create a new runner that uses the helper.

PR #657, opened 2019-10-24
Delete pod from GKE after completion

jasonthomas:
When attempting to run some GKEPodOperator tasks I noticed that the tasks were timing out and were not scheduled on `bq-load-gke-1`. It looks like there is a 1K deployment limit on GKE and once I cleaned up old pod deployments the new deployments worked. I can't find documentation around the actual limit but I think going forward we should just clean them up since logs are already available in Airflow. I've manually cleaned up old pod deployments for now.

Enabling `is_delete_operator_pod` will delete the pod when when the pod reaches its final state, or the execution is interrupted.

r?

### DIFF
--- a/dags/adi_dim_backfill.py
+++ b/dags/adi_dim_backfill.py
@@ -61,6 +61,7 @@ load_bq_to_tmp_tbl = GKEPodOperator(
     name='bq-load-tmp-tbl',
     namespace='default',
     image='google/cloud-sdk:242.0.0-alpine',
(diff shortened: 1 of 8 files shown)


wlach:
We should probably go ahead and merge this, no?

haroldwoo:
Closing out. Should'nt be needed now that https://github.com/mozilla/telemetry-airflow/pull/817 will clean out old pods

PR #663, opened 2019-10-29
Fix probe scraper to trigger GKE autoscaling for highmem node pool

haroldwoo:


dags/probe_scraper.py
```diff
@@ -40,6 +40,10 @@
         '--env', 'prod'
     ]
 
+    # Cluster autoscaling works on pod resource requests, instead of usage
+    resources = Resources(request_memory='13312Mi', request_cpu=None,
+                         limit_memory='20480Mi', limit_cpu=None)
```
fbertsch: Why `20480Mi`?

haroldwoo: no particular reason, allowing for growth over time. And the highmem node has 26gb allocatable

haroldwoo: Oh also, if not set, it gives the namespace's default which is probably 512, so we need to set it

PR #669, opened 2019-10-31
Add bq-based events jobs

sunahsuh:
Schedules the bigquery-based events queries for eventual replacement of the current spark jobs


jklukas:
Looks like this ran successfully last night, so Jason's permissions changes in Airflow are working!

PR #672, opened 2019-11-01
Run queries in shared-prod for destination tables in shared-prod

jklukas:
This is now possible due to updates to permissions such that Airflow can
now run BQ jobs in shared-prod.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1582791

Thank you, @jasonthomas. This is going to avoid some significant hassles esp. for the GLAM queries that are coming up.

### DIFF
--- a/dags/fxa_events.py
+++ b/dags/fxa_events.py
@@ -9,7 +9,7 @@ default_args = {
     'email': ['telemetry-alerts@mozilla.com', 'jklukas@mozilla.com'],
     'email_on_failure': True,
     'email_on_retry': True,
-    'depends_on_past': True,
+    'depends_on_past': False,
     # If a task fails, retry it once after waiting at least 5 minutes
     'retries': 1,
     'retry_delay': datetime.timedelta(minutes=10),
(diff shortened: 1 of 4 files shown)


jklukas:
I will be prepared to trigger some manual runs of the queries here once this gets merged to make sure I haven't missed something. I'll file a quick revert PR is that goes poorly.

dags/fxa_events.py
```diff
@@ -106,6 +106,7 @@
 
     fxa_users_services_first_seen = bigquery_etl_query(
         task_id='fxa_users_services_first_seen',
+        project_id='moz-fx-data-shared-prod',
         sql_file_path='sql/telemetry_derived/fxa_users_services_first_seen_v1/init.sql',
```
jasonthomas: Do we want to replace `sql_file_path` here too?

```suggestion
        destination_table='fxa_users_services_first_seen_v1',
```

jklukas: No, this is a weird case. We recreate this table in its entirety every day rather writing just a new partition. We have to specify the full path to `init.sql` even with this change.

PR #676, opened 2019-11-05
Migrate events to amplitude off of EMR

haroldwoo:
- [ ] Create aws keys that have access to the CI artifacts and amplitude keys needed
- [ ] Create aws keys that have access to whatever the spark jar jobs need to do e.g. rw to s3 buckets
- [ ] Add aws connections to wtmo and sops secrets
- [ ] Refactor jar if necessary? Ask frank about how amplitude should be exported
- [ ] Test
- [ ] Tweak sizes of clusters
- [ ] Todos before merging

### DIFF
--- a/dags/events_to_amplitude.py
+++ b/dags/events_to_amplitude.py
@@ -19,8 +19,8 @@ FIRE_TV_INSTANCES = 10
 VCPUS_PER_INSTANCE = 16
 
 # TODO - change this to 'prod' before merging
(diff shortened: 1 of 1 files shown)


haroldwoo:
Closing out, this task will be handled by someone on the data eng team who knows how the data should look.

Events_to_amplitude will no longer run a spark jar, and will use the bq “view” approach instead.

PR #685, opened 2019-11-07
Support BigQuery export to parquet via avro

relud:
and use it to export main summary

requires https://github.com/mozilla/bigquery-etl/pull/492


relud:
`clients_last_seen` takes on the order of 10K seconds to export via bigquery storage api, and is 1/10th the size of `main_summary`, so I believe this will be necessary in order to achieve acceptable performance when exporting `main_summary`

PR #686, opened 2019-11-07
Bug 1582110 - Add hardware report job with BigQuery support

akkomar:
This moves hardware report job from python_mozetl, adding an option to read from BigQuery.

It turned out it's pretty hard to run mozetl on Dataproc due to dependency conflicts. This is a temporary solution to unblock Longitudinal decommisioning. As a follow up we should clean up python_mozetl, repurpose it for Dataproc and move the job there [2].

While running on Dataproc, `main` ping table is used as data source.
Longitudinal-specific code is now extracted to `get_data_longitudinal` method, BigQuery shim was added to `get_data_bigquery`. Both methods return a dataframe with the same schema.

I did a test run, which:
* passed the validation step executed by the end of the job
* produced output[1] that looks reasonable and similar to the previous runs

[1] s3://telemetry-public-analysis-2/game-hardware-survey-test/data/hwsurvey-weekly-20190711.json
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1594845


jobs/hardware_report.py
```diff
@@ -0,0 +1,1009 @@
+import json
```
acmiyaguchi: Can you refer to the original location of the script from the mozetl repository at the top of the file, where the tests are also located?

akkomar: I put a link on line 27.

PR #688, opened 2019-11-07
Fix import path for operator from utils

relud:


relud:
as seen at https://github.com/mozilla/telemetry-airflow/blob/master/dags/utils/amplitude.py#L7

PR #694, opened 2019-11-08
Merge telemetry aggregates job into a single file

acmiyaguchi:
This makes it easier to find and modify all of the telemetry aggregates jobs.


acmiyaguchi:
cc @fbertsch

dags/telemetry_aggregates.py
```diff
@@ -0,0 +1,180 @@
...
+
+common = {
+    "depends_on_past": True,
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
+}
+
+default_args = common.copy()
```
akkomar: How about renaming `default_args` to `default_args_prerelease` here and similarly below?

PR #701, opened 2019-11-09
Fix export_to_parquet gcs_to_s3 step

relud:
Disable XCOM in GoogleCloudStorageToS3Operator to prevent too large
value error.

Include static partitions in export_prefix so that we don't copy all
history every day.

r=me


relud:
cc @akkomar

PR #702, opened 2019-11-09
Use thread pool to copy from gcs to s3

relud:
This forks the upstream gcs to s3 operator to add support for distributing tasks to a `ThreadPool`.

Without this it took 48 minutes to copy 12.8 GiB for clients_daily.

r=me

cc @akkomar

### DIFF
--- a/dags/operators/gcs_to_s3.py
+++ b/dags/operators/gcs_to_s3.py
@@ -1,8 +1,0 @@
-"""Disable XCOM in GoogleCloudStorageToS3Operator."""
-
-from airflow.contrib.operators import gcs_to_s3
-
-class GoogleCloudStorageToS3Operator(gcs_to_s3.GoogleCloudStorageToS3Operator):
-    def execute(self, context):
-        # don't return files moved because the list may be too long for XCOM.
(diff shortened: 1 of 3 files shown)


akkomar:
@relud it could be helpful in the future if the commit hash this for forked from was mentioned somewhere.

also cc @haroldwoo

relud:
@akkomar good point, I totally forgot to include that. This was insufficient to make main summary successfully copy to s3 and I will probably be removing it in favor of a different solution, so I won't make a PR to add that reference.

haroldwoo:
you could try using GKEPodOperator with the gcloud base image performing a gsutil cp -m E.g. https://github.com/mozilla/telemetry-airflow/blob/master/dags/socorro_import.py#L125-L132

relud:
> you could try using GKEPodOperator with the gcloud base image performing a gsutil cp -m

This dag used to do that, but I updated it because the of concerns about the aws credentials being exposed in the airflow UI.

This is now using DataProcHadoopOperator (patched to hide the aws credentials from the UI) to run hadoop's distcp to do a distributed copy to s3, because the subdag already had a dataproc cluster available.

PR #716, opened 2019-11-12
Point search datasets to shared prod

Ben-Wu:
Depends on https://github.com/mozilla/bigquery-etl/pull/500


dags/main_summary.py
```diff
@@ -629,19 +629,19 @@
...
     owner="bewu@mozilla.com",
     email=["telemetry-alerts@mozilla.com", "bewu@mozilla.com"],
-    sql_file_path='sql/search/search_clients_daily_v8/query.sql',
     dag=dag)
 
 search_aggregates_bigquery = bigquery_etl_query(
     task_id="search_aggregates_bigquery",
     destination_table="search_aggregates_v8",
-    dataset_id="search_staging",
+    dataset_id="search",
```
relud: this should write to a `search_derived` dataset, and `search.search_aggregates` and `search.search_clients_daily` should be views.

PR #717, opened 2019-11-12
Use hadoop to copy from gcs to s3

relud:
This copy method is running at ~145MB/s in testing on main summary for 2019-11-11 with ~300GB of parquet files, which should be fast enough to complete in a reasonable amount of time.

This is to work around issues with `GoogleCloudStorageToS3Operator` failing on large amounts of data.


dags/utils/gcp.py
```diff
@@ -439,11 +442,12 @@ def export_to_parquet(
                 bucket_name=gcs_output_bucket,
                 prefix=avro_prefix,
                 gcp_conn_id=gcp_conn_id,
+                trigger_rule=trigger_rule.TriggerRule.ALL_DONE,
```
relud: this makes sure we don't leave files around in GCS after errors.

dags/operators/dataproc_hadoop_with_aws.py
```diff
@@ -0,0 +1,46 @@
...
+            for key, value in zip(
+                ("access.key", "secret.key", "session.token"),
+                AwsHook(aws_conn_id=self.aws_conn_id).get_credentials(),
+            )
+            if value is not None
+        }
+        if self.dataproc_properties is not None:
+            properties.update(self.dataproc_properties)
+
+        # fork super().execute to reference combined properties. forked from:
```
acmiyaguchi: Would this work if `self.dataproc_properties` were set via `self.dataproc_properties.update(properties)`, followed by `super().execute()`?

dags/utils/gcp.py
```diff
@@ -439,11 +442,12 @@ def export_to_parquet(
...
             )
             avro_export >> run_dataproc_pyspark >> avro_delete
 
-        create_dataproc_cluster >> run_dataproc_pyspark >> delete_dataproc_cluster
-        run_dataproc_pyspark >> gcs_to_s3
+        create_dataproc_cluster >> run_dataproc_pyspark >> gcs_to_s3
+        gcs_to_s3 >> delete_dataproc_cluster
```
acmiyaguchi: This can be merged into a single line:

```suggestion
        create_dataproc_cluster >> run_dataproc_pyspark >> gcs_to_s3 >> delete_dataproc_cluster
```

dags/operators/dataproc_hadoop_with_aws.py
relud: I expect `self.dataproc_properties.update(properties)` would expose the credentials in the airflow UI, but I'll check.

dags/utils/gcp.py
relud: it can, but it's longer than 88 characters

dags/operators/dataproc_hadoop_with_aws.py
acmiyaguchi: I see, so the class/instance variables associated with the operator are displayed in the Airflow UI, so keeping the properties as a local variable is the workaround?

relud:
> Can you also comment in some detail about how you tested this job too, for future reference?

To test this I ran `make up` and checked in the UI that:

- dags don't have any errors compiling
- credentials are not exposed
- template variables look correct

and then I created a cluster with the worker-type, machine-type, image-version and num-workers used by main_summary. I then copied the values in the UI for main_summary on 2019-11-11, replaced the gcs bucket name with one in my testing project, and submitted a hadoop job with those arguments via `gcloud` to my cluster.

After I deleted the target directory in s3 the job succeeded in 42 minutes.

The reason I needed to delete first was that `GoogleCloudStorageToS3Operator` creates 0-byte objects for directories, and hadoop interpreted this to mean that I was copying a directory to a file. Re-running the job succeeded in 3.5 minutes (because it only had to check timestamps and file sizes), so retries should be fine as long as `GoogleCloudStorageToS3Operator` wasn't previously used.

dags/operators/dataproc_hadoop_with_aws.py
relud: yes

relud: also, it appears that the UI is reading the values from class instances before `execute` gets called, so that it doesn't show the credentials, however I don't want to rely on that behavior.

PR #722, opened 2019-11-13
Update upstream for ssl_ratios

relud:
now reads `main` (populated by copy_dedup) rather than `main_summary` as of https://github.com/mozilla/bigquery-etl/pull/502


dags/ssl_ratios.py
```diff
@@ -19,9 +19,9 @@
 with DAG("ssl_ratios", default_args=default_args, schedule_interval="@daily") as dag:
     # most downstream dependency is search_clients_daily
     wait_for_main_summary = ExternalTaskSensor(
-        task_id="wait_for_main_summary",
+        task_id="wait_for_copy_deduplicate",
```
jklukas:
```suggestion
        task_id="wait_for_copy_deduplicate_main_ping",
```

Just to be explicit.

PR #723, opened 2019-11-13
Bug 1596233 - Unschedule telemetry_aggregate_parquet_view

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1596233)

As far as I know, this dataset isn't being used anywhere. The underlying code has been successfully run on dataproc using bigquery, but it may not be worth the effort to schedule this on a regular basis. The data lake (`telemetry-parquet`), will be going away by the end of the year.


robhudson:
As far as I know also, these aren't being used. We were going to announce the creation of these tables but never did, so I think this is pretty safe to do.

PR #725, opened 2019-11-14
Replace spark clients_daily with sql version

relud:
co-dependent on https://github.com/mozilla/bigquery-etl/pull/501


dags/main_summary.py
```diff
@@ -75,7 +75,6 @@
 main_summary_export = SubDagOperator(
     subdag=export_to_parquet(
         table="moz-fx-data-shared-prod:telemetry_derived.main_summary_v4${{ds_nodash}}",
-        destination_table="main_summary_v4",
```
relud: removed because it's redundant

dags/utils/gcp.py
```diff
@@ -352,7 +352,7 @@ def export_to_parquet(
     connection = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id)
 
     if destination_table is None:
-        destination_table = table.rsplit(".", 1).pop()
+        destination_table = unqualified_table
```
relud: fixes issue in `clients_last_seen_v1_export` and `exact_mau_by_dimensions_export` where the partition qualifier (e.g. `$20191109`) was being included in the s3 path

dags/main_summary.py
```diff
@@ -336,45 +336,12 @@
...
-        replace=["SAFE_CAST(sample_id AS INT64) AS sample_id"],
-        ),
-    task_id="clients_daily_v6_bigquery_load",
-    dag=dag)
+register_status(clients_daily, "Clients Daily", "A view of main pings with one row per client per day.")
 
 clients_last_seen = bigquery_etl_query(
     task_id="clients_last_seen",
     destination_table="clients_last_seen_v1",
+    project_id="moz-fx-data-shared-prod",
```
relud: when this is merged the most recent day of `clients_last_seen_v1` will need to be copied from `moz-fx-data-derived-datasets` to `moz-fx-data-shared-prod` so that it can referenced in the next run.

dags/main_summary.py
```diff
@@ -382,6 +380,7 @@
             "scalar_parent_aushelper_websense_reg_version",
             "scalar_parent_dom_contentprocess_troubled_due_to_memory_sum",
             "--replace",
+            "DATE_FORMAT(profile_creation_date, 'yyyy-MM-dd 00:00:00') AS profile_creation_date",
```
jklukas: Is this necessary? This is already a string in the raw table, isn't it?

jklukas:
r+wc

dags/main_summary.py
relud: discussed in slack, this will be a string in the raw table

PR #727, opened 2019-11-14
Bug 1572115 - Add dataproc version aggregates job that writes to a dev database

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1572115)

This writes aggregates for the day into the dev database in AWS. Once this has been validated, we should swap the dataproc version for the databricks version. 

First commit in a series of commits. I've tried to test this locally, but I run into a permissions issue when spinning up the dataproc cluster:

```
[2019-11-14 00:32:25,514] {base_task_runner.py:101} INFO - Job 16: Subtask create_dataproc_cluster googleapiclient.errors.HttpError: <HttpError 400 when requesting https://dataproc.googleapis.com/v1beta2/projects/amiyaguchi-dev/regions/global/clusters?alt=json&requestId=fee90338-8182-42e3-99f9-070e2bb14a76 returned "791317032024@cloudservices.gserviceaccount.com does not have storage.buckets.get access to moz-fx-data-prod-dataproc-scratch.">
```

I'm using a service account created in my sandbox project `amiyaguchi-dev` under ` dataproc-worker@amiyaguchi-dev.iam.gserviceaccount.com `. @haroldwoo, do you have a suggestion on how to run the job via local airflow with a service account created in the sandbox project? I think getting access to the scratch bucket might be enough. I can also conditionally use a different bucket when running locally if there's a nonprod scratch bucket.


dags/mozaggregator_prerelease.py
```diff
@@ -44,3 +54,51 @@
...
+keyfile = json.loads(gcp_conn.extras["extra__google_cloud_platform__keyfile_dict"])
+client_email = keyfile["client_email"]
+
+prerelease_telemetry_aggregate_view_dataproc = SubDagOperator(
+    task_id=task_id,
+    dag=dag,
+    subdag=moz_dataproc_pyspark_runner(
+        parent_dag_name=dag.dag_id,
+        dag_name=task_id,
+        job_name="prerelease aggregates",
```
akkomar: This will fail due to whitespace. See https://github.com/mozilla/telemetry-airflow/commit/96f6a96d3586414402f34a703537bf9043e26af2

I'm wondering if this could be checked during tests, maybe as some assertion in the operator code?

jobs/mozaggregator_runner.py
```diff
@@ -0,0 +1,10 @@
+# This runner has been auto-generated from mozilla/python_mozaggregator/bin/dataproc.sh.
+# Any changes made to the runner file will be over-written on subsequent runs.
+from mozaggregator import cli
+
+try:
+    cli.entry_point(auto_envvar_prefix="MOZETL")
+except SystemExit:
+    # avoid calling sys.exit() in databricks
```
akkomar: Nit: this probably doesn't hurt, but I think we no longer need this since this is targeting Dataproc.

dags/mozaggregator_prerelease.py
```diff
@@ -44,3 +54,51 @@
...
+            "--num-partitions",
+            10 * 32,
+            "--source",
+            "bigquery",
+            "--project-id",
+            "moz-fx-data-shared-prod",
+        ],
+        aws_conn_id="aws_dev_iam_s3",
+        gcp_conn_id=gcp_conn.gcp_conn_id,
+        service_account=client_email,
```
akkomar: Why do we need `service_account` here?

I'm asking because I haven't been providing it in update_orphaning and hardware report jobs. cc @haroldwoo

haroldwoo: The gcp_conn_id is used by airflow for permissions to create/delete dataproc clusters, and run dataproc workloads that only touch services in the same airflow gcp project. The cluster is created with the project's default service account, which is different from the gcp_conn_id's service account, unless we specify service_account here.

Specifying a service account here instantiates the dataproc cluster with a custom service account, and we usually do this when we need cross project permissions. Its just confusing because we often re-use the same service account for the gcp_conn_id and service_account params here.

haroldwoo:
"@haroldwoo, do you have a suggestion on how to run the job via local airflow with a service account created in the sandbox project? I think getting access to the scratch bucket might be enough. I can also conditionally use a different bucket when running locally if there's a nonprod scratch bucket."

This is a good question. Does this need access to other GCP project resources outside your sandbox? (e.g. bq). If not, then you could create a scratch bucket in your sandbox gcp project and use that.

I also had a concern about writing to the AWS postgres db and idempotency/retention. Is this handled in the dag? Does this data need to live in the same AWS postgres db or postgres at all (as opposed to bq, other)?

dags/mozaggregator_prerelease.py
acmiyaguchi: I'm putting this here in particular because I'm trying to test locally. I've updated the `google_cloud_airflow_dataproc` connection to insert my keyfile for `amiyaguchi-dev`. This will default to the service account associated with the production account otherwise.

```bash
function format_extras() {
KEYFILE="$1" python3 - <<END
import os
import json

with open(os.environ["KEYFILE"]) as f:
    data = json.load(f)

extra = {
    "extra__google_cloud_platform__project": data["project_id"],
    "extra__google_cloud_platform__keyfile_dict": json.dumps(data)
}

print(json.dumps(extra))
END
}

function update() {
    local conn_id=$1
    local keyfile=$2

    docker exec $container_id \
        airflow connections -d --conn_id $conn_id
    
    docker exec $container_id \
        airflow connections -a \
            --conn_id $conn_id \
            --conn_type google_cloud_platform \
            --conn_extra "$(format_extras $keyfile)"
}

container_id=$(docker ps | grep web | cut -d' ' -f1)
update google_cloud_airflow_dataproc <path to keyfile>
```

acmiyaguchi:
> Does this need access to other GCP project resources outside your sandbox?

It doesn't, so it looks like I can modify the code path to include a custom scratch bucket for testing.

> I also had a concern about writing to the AWS postgres db and idempotency/retention. Is this handled in the dag? Does this data need to live in the same AWS postgres db or postgres at all (as opposed to bq, other)?

This job is not idempotent because it's incremental and will increase the size of the tables with each consecutive run. This is writing to the dev database though, so I'm focused on that the job doesn't fail as opposed to correctness.

haroldwoo:
The database we use in dev is considered production though. If this job is non-idempotent, is there a clear plan on how this job should be re-ran in case of failure?

acmiyaguchi:
`aggregator_dev_database_envvars.json` was explicitly created for testing the job, it is not referenced by any production services. For the dataproc job, we can just retry until it runs successfully, it's likely that it fails on bootstrap.

For the production job that touches `aggregator_database_envvars.json`, if any data has been written during a failure, I believe we may need to rollback the database from backup and try the job again (although I'm not confident on this piece, @fbertsch would be able to comment in more depth).

dags/mozaggregator_prerelease.py
acmiyaguchi: I checked this with manual testing, and found a slew of other bugs. This should be fine now.

acmiyaguchi:
The dummy connections and vars are necessary to make the DAGs available locally, though. Being able to test this job locally was very useful for development, and I think having the code around to bootstrap the job for a sandbox project will be useful in the future.

It's possible that it actually is idempotent, I haven't been able to characterize the behavior because it's difficult for me to actually navigate in the postgres database. This section in the actual job may prevent loading into the same build_id/submission_date combo via an upsert: 
https://github.com/mozilla/python_mozaggregator/blob/35bb1c0a0d1051e0662d6169eaa5d2387995e278/mozaggregator/db.py#L223-L230

PR #729, opened 2019-11-14
Restore int columns in main_summary_export

relud:
requires https://github.com/mozilla/bigquery-etl/pull/506


dags/main_summary.py
```diff
@@ -81,6 +81,44 @@
             "--partition-by=sample_id",
             "--replace='{{ds_nodash}}' AS submission_date",
             "--maps-from-entries",
+            "--bigint-columns",
+            "search_counts.count",
```
jklukas: Perhaps add a comment here about how we arrived at this list and under what circumstances we might need to update it.

PR #737, opened 2019-11-15
Bugfix for clients_daily_export

jklukas:
Should fix error:

> org.apache.spark.sql.AnalysisException: Partition column `submission_date` not found in schema

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -367,6 +367,8 @@ clients_daily_export = SubDagOperator(
         arguments=[
             # restore legacy schema
             "--maps-from-entries",
+            "--partition-by",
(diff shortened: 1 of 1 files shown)


dags/main_summary.py
```diff
@@ -367,6 +367,8 @@
         arguments=[
             # restore legacy schema
             "--maps-from-entries",
+            "--partition-by",
+            "submission_date_s3",
```
jklukas: This field is already specified above in `static_partitions="submission_date_s3={{ds_nodash}}",`, but it also needs to appear here in order to defeat the default partitioning field (`submission_date`).

PR #738, opened 2019-11-15
Bump mobile_metrics dataset to v3 with submission_date based partitioning

acmiyaguchi:
This is necessary to deploy https://github.com/mozilla/python_mozaggregator/pull/201. Having this partitioning scheme will make it easier to validate the data and keep it around.

For reference, the command line arguments are as follows:
https://github.com/mozilla/python_mozaggregator/blob/35bb1c0a0d1051e0662d6169eaa5d2387995e278/mozaggregator/cli.py#L125-L143


acmiyaguchi:
Just a note that the `channel` argument doesnt exist for the mobile job, it was likely copied and pasted from one of the other mozaggregator jobs.

PR #739, opened 2019-11-15
Bug 1596917 - Read credentials from airflow variables

acmiyaguchi:
This reads credentials from airflow variables and relies on https://github.com/mozilla/python_mozaggregator/pull/202.

This approach leaks the credentials into the arguments list in the dataproc cluster view. The job may also read credentials from the environment, but there isn't an appropriate way to load environment variables into a dataproc cluster at submission time.

This PR was created because the mozaggregator job relies on credentials to the postgres database in aws. However, the `aws_dev_iam_s3` role does not have access to the `telemetry-spark-emr-2` bucket. We also do not have an equivalent gcs bucket to store credentials in.


acmiyaguchi:
The following section needs to be set because the dependency on mozaggregator hasn't been merged:

```python
        additional_metadata={
            "PIP_PACKAGES": "git+https://github.com/mozilla/python_mozaggregator.git@environment"
        },
```

dags/mozaggregator_prerelease.py
```diff
@@ -138,18 +138,23 @@ def copy_artifacts_dev(dag, project_id, artifact_bucket, storage_bucket):
             "{{ ds_nodash }}",
             "--channels",
             "nightly,aurora,beta",
-            "--credentials-bucket",
-            "telemetry-spark-emr-2",
-            "--credentials-prefix",
-            "aggregator_dev_database_envvars.json",
+            "--postgres-db",
```
haroldwoo: We dont really need variables for the db and username.

PR #746, opened 2019-11-19
Move nondesktop KPI queries to stable table DAG

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/519


dags/copy_deduplicate.py
```diff
@@ -41,3 +44,69 @@
...
+
+    (copy_deduplicate_all >>
+     core_clients_daily >>
+     core_clients_last_seen >>
+     [firefox_nondesktop_exact_mau28, smoot_usage_nondesktop_v2])
+
+    (copy_deduplicate_all >>
+     fenix_clients_daily >>
+     fenix_clients_last_seen >>
+     [firefox_nondesktop_exact_mau28, smoot_usage_nondesktop_v2])
```
jklukas: We use a few different styles for dependency chaining, and this is introducing an additional style. But this feels like a less redundant style that also better shows the full chain in one place. Open to pushback on it.

relud: I like it

PR #747, opened 2019-11-20
drop --output flag to AddonRecommender

relud:
flag removed in https://github.com/mozilla/telemetry-batch-view/pull/555/files#diff-ebfbc196fc5b4a7f5a210a2363aa5ae1L42


relud:
r=me

PR #749, opened 2019-11-20
Run AddonRecommender on dataproc

relud:
I still need to test this


dags/main_summary.py
```diff
@@ -566,18 +567,42 @@
...
-    email=["telemetry-alerts@mozilla.com", "mlopatka@mozilla.com", "vng@mozilla.com"],
-    env={"date": "{{ ds_nodash }}",
-         "privateBucket": "{{ task.__class__.private_output_bucket }}",
-         "publicBucket": "{{ task.__class__.public_output_bucket }}"},
-    uri="https://raw.githubusercontent.com/mozilla/telemetry-airflow/master/jobs/addon_recommender.sh",
-    dag=dag)
+    subdag=moz_dataproc_jar_runner(
+        parent_dag_name=dag.dag_id,
+        dag_name="addon_recommender",
+        job_name="Train the Collaborative Addon Recommender",
```
akkomar: This will fail because of whitespaces (we're not sanitizing job names in current operator implementation).

dags/main_summary.py
```diff
@@ -566,18 +567,42 @@
...
+          "--runDate={{ds_nodash}}",
+          "--inputTable=gs://moz-fx-data-derived-datasets-parquet/clients_daily/v6",
+          "--privateBucket=telemetry-parquet",
+          "--publicBucket=telemetry-public-analysis-2",
+        ],
+        cluster_name="addon-recommender-{{ds_nodash}}",
+        image_version="1.3",
+        worker_machine_type="n1-standard-8",
+        num_workers=20,
+        optional_components=[],
```
akkomar: nit:
```suggestion
```

dags/main_summary.py
```diff
@@ -566,18 +567,42 @@
...
+        install_component_gateway=False,
+        init_actions_uris=[],
```
akkomar: nit:
```suggestion
```

dags/main_summary.py
relud: without this dataproc will be configured to install anaconda on startup, which we don't need.

dags/main_summary.py
relud: without this dataproc will be configured to run `gs://moz-fx-data-prod-airflow-dataproc-artifacts/bootstrap/dataproc_init.sh` on startup, which we don't need.

dags/main_summary.py
relud:
```suggestion
        job_name="Train_the_Collaborative_Addon_Recommender",
```

PR #750, opened 2019-11-21
Bug 1572115 - Unschedule prerelease telemetry aggregates from databricks

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1572115)

I've done validation on the data using 2019-11-18 and 2019-11-19. We'll likely be okay swapping data sources to BigQuery.

# Comparison
Test is data that has been written into the dev aggregates database being written into by dataproc. The reference (ref) is the production aggregates database.

Error is defined to be `|ref - test|/ref * 100`.

Source: https://gist.github.com/acmiyaguchi/a2a405545147404d232dc80a2f5da4a0

### Task duration

image

The BQ source/Dataproc job takes 30-45 minutes longer than it did in Databricks.

### Count of rows
```
  submission_date     test      ref       err
0        20191119  3693149  3754045  1.622144
1        20191118  3645259  3699431  1.464333
```

The number of rows is larger when sourcing from BQ. 

### Sum of all histogram sums

```
               sum_test               sum_ref                         sum_err
0  20186565809991403910  20321078377354477065  0.6619361672900787665812792448
1  19764958396836226602  19789360689297368440  0.1233101606679960620068926903
```

The histogram values are likewise larger. 

### Sum of all histogram counts

```
   count_test   count_ref                      count_err
0  4709045032  4791465719  1.720156040628034805313818421
1  4688191083  4748249808  1.264860262802752689544256598
```

### plot of aggregated `GC_MS`

image

There is a small, but negligible difference, between the distributions of GC_MS.


acmiyaguchi:
The credentials will need to be changed within the airflow connections page with this PR.

PR #754, opened 2019-11-21
Add dag for error aggregates query

Ben-Wu:


dags/error_aggregates.py
```diff
@@ -0,0 +1,29 @@
+import datetime
+
+from airflow import models
+from utils.gcp import bigquery_etl_query
+
+default_args = {
+    'owner': 'bewu@mozilla.com',
+    'start_date': datetime.datetime(2019, 11, 1),
+    'email': ['telemetry-alerts@mozilla.com', 'bewu@mozilla.com', 'wlach@mozilla.com'],
```
wlach:
```suggestion
    'email': ['telemetry-alerts@mozilla.com', 'bewu@mozilla.com', 'wlachance@mozilla.com'],
```

wlach: `wlach@mozilla.com` should work, but may as well use my non-aliased email to be a bit more confident it gets through

PR #756, opened 2019-11-25
Switch addons and addon_aggregates to being generated in sql

relud:
dags have no errors locally, and running queries by copy-pasting rendered operators worked without issue

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -135,34 +135,28 @@ main_summary_export = SubDagOperator(
 
 register_status(main_summary, "Main Summary", "A summary view of main pings.")
 
-addons = EMRSparkOperator(
+addons = bigquery_etl_query(
     task_id="addons",
-    job_name="Addons View",
-    execution_timeout=timedelta(hours=4),
-    instance_count=3,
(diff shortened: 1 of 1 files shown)


dags/main_summary.py
```diff
@@ -135,34 +135,28 @@
...
-    subdag=load_to_bigquery(
+addons_export = SubDagOperator(
+    subdag=export_to_parquet(
+        table="moz-fx-data-derived-datasets:telemetry_derived.addons_v2${{ds_nodash}}",
+        static_partitions="submission_date_s3={{ds_nodash}}",
+        arguments=[
+            "--partition-by=sample_id",
+            "--drop=submission_date",
+            # no bigint columns, convert them all to int for backward compatibility
+            "--bigint-columns=",
```
acmiyaguchi: Is there a particular reason that this option is here, but not in addon_aggregates_v2? This is also missing a value, does this default to casting all bigint columns into int if the value is None?

relud: > Is there a particular reason that this option is here, but not in addon_aggregates_v2?

in `addon_aggregates_v2` there are no int columns, only bigint, so we don't need this.

> This is also missing a value

it's an empty list of values, indicating that all bigint columns should be cast to int.

PR #759, opened 2019-11-26
Bug 1574490 - Send iOS events to Amplitude from BigQuery view

akkomar:
This should be merged after https://github.com/mozilla/bigquery-etl/pull/510 is ready and test load to dev iOS project in Amplitude is validated.

### DIFF
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -21,15 +21,30 @@ with models.DAG(
         dag_name,
         default_args=default_args) as dag:
 
-    task_id = 'fenix_amplitude_export'
+    fenix_task_id = 'fenix_amplitude_export'
     SubDagOperator(
         subdag=export_to_amplitude(
-            dag_name=task_id,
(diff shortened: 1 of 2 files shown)


dags/bq_events_to_amplitude.py
```diff
@@ -21,15 +21,30 @@
...
             dataset='telemetry',
             table_or_view='fenix_events_v1',
             s3_prefix='fenix',
         ),
-        task_id=task_id
+        task_id=fenix_task_id
+    )
+
+    fennec_ios_task_id = 'fennec_ios_amplitude_export'
+    fennec_ios_args = default_args.copy()
```
fbertsch: This could be done in one line with `.copy().update(..)`

akkomar: Unfortunately `update()` does not return a dictionary instance (just updates it in place).

akkomar:
I'm going to close this and include iOS view in https://github.com/mozilla/telemetry-airflow/pull/779

PR #763, opened 2019-11-26
Run main_summary_experiments on dataproc

relud:
requires https://github.com/mozilla/telemetry-batch-view/pull/557

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -215,23 +215,41 @@ addon_aggregates_export = SubDagOperator(
     executor=GetDefaultExecutor(),
     dag=dag)
 
-main_summary_experiments = MozDatabricksSubmitRunOperator(
+main_summary_experiments = SubDagOperator(
     task_id="main_summary_experiments",
-    job_name="Experiments Main Summary View",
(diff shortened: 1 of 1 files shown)


dags/main_summary.py
```diff
@@ -215,23 +215,41 @@
...
+        jar_args=[
+            "--from={{ds_nodash}}",
+            "--to={{ds_nodash}}",
+            "--inbucket=gs://moz-fx-data-derived-datasets-parquet",
+            "--bucket=s3a://telemetry-parquet",
+        ],
+        cluster_name="main-experiments-{{ds_nodash}}",
+        image_version="1.3",
+        worker_machine_type="n2-highmem-8",
+        num_workers=40,
```
sunahsuh: Are we enabling autoscaling on dataproc jobs? The load for this job tends to vary based on what experiments are running at any given time so autoscaling makes a lot of sense for it.

relud: the dataproc stuff we have right now doesn't support autoscaling. i'll file a bug to attempt to add that if i can't migrate it to sql soon

PR #770, opened 2019-11-27
Export bq-based event_events_v1 to aws (and write to test bucket for …

sunahsuh:
…now)

Includes one minor change to the `export_to_parquet` function, which is to make `static_partitions` a list instead of a string so we can pass in multiple arguments (which the export_parquet script supports)

I was able to get the DAGs to load locally but haven't been able to run the actual task since gcp permissions are non-trivial to set up correctly locally, so I'm going to run at least one day of this as a test writing `telemetry-test-bucket`.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -1,7 +1,13 @@
 import datetime
 
 from airflow import models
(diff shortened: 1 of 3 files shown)


sunahsuh:
Tagging in Arkadiusz as well since it's late on a day before a US holiday, and this is removes a moztelemetry.Dataset dependency

PR #772, opened 2019-11-28
Bug 1596205 - Turn off AWS bgbb (active_profiles) jobs

acmiyaguchi:
This turns off the the aws jobs and modifies the bigquery load job to skip loading from aws. This can be merged in once there has been a successful airflow of the predict job.

### DIFF
--- a/dags/bgbb.py
+++ b/dags/bgbb.py
@@ -3,14 +3,12 @@ from datetime import datetime, timedelta
 from airflow import DAG
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
 from airflow.operators.dummy_operator import DummyOperator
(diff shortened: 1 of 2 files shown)


erindcole:
@acmiyaguchi  need some help?

acmiyaguchi:
@erindcole There's not all that much to do here, most of the migration work [has already been done](https://bugzilla.mozilla.org/show_bug.cgi?id=1596205), plus testing this job requires permissions within the data engineering org to access tables.

@akkomar I'm going to merge this PR, it looks like having both the databricks and dataproc job running at the same time is causing strange issues. https://github.com/wcbeard/bgbb_airflow/pull/15 should fix schema errors that were causing the load job to fail, as well as some of the pyarrow issues with Spark.

PR #775, opened 2019-12-02
Output experiment_enrollment_aggregates_live view query as json for x…

sunahsuh:
…com_push option

See mozilla/bigquery-etl#558 for the corresponding change


dags/copy_deduplicate.py
```diff
@@ -97,7 +97,8 @@
             "python",
             "templates/telemetry_derived/experiment_enrollment_aggregates_live/view.sql.py",
             "--submission-date",
-            "{ds}"
+            "{{ ds }}",
+            "--json-output"
```
relud:
```suggestion
            "--json-output",
```

PR #777, opened 2019-12-02
Use doc_type as partitioning column instead

sunahsuh:
Okay, I was able to test this config with --dry_run on the xport script and the output is what I had actually expected


sunahsuh:
Again, merging in advance of review to try to run ahead of the migration

PR #780, opened 2019-12-03
Write event ping events export to telemetry-parquet and deschedule sp…

sunahsuh:
…ark job

Last one for events!

### DIFF
--- a/dags/event_ping_events.py
+++ b/dags/event_ping_events.py
@@ -12,6 +12,7 @@ default_args = {
     'owner': 'ssuh@mozilla.com',
     'depends_on_past': False,
     'start_date': datetime(2018, 11, 26),
+    'end_date': datetime(2019, 12, 02),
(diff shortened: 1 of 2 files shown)


dags/event_ping_events.py
```diff
@@ -12,6 +12,7 @@
     'owner': 'ssuh@mozilla.com',
     'depends_on_past': False,
     'start_date': datetime(2018, 11, 26),
+    'end_date': datetime(2019, 12, 02),
```
relud: `SyntaxError: leading zeros in decimal integer literals are not permitted`
```suggestion
    'end_date': datetime(2019, 12, 2),
```

sunahsuh: ...What is the airflow CI even doing if it's not catching syntax errors?

relud: 😭

PR #781, opened 2019-12-03
Run sql experiments_v1 alongside databricks

relud:


dags/main_summary.py
```diff
@@ -219,6 +223,59 @@
...
+    subdag=export_to_parquet(
+        table="moz-fx-data-shared-prod:telemetry_derived.experiments_v1${{ds_nodash}}",
+        destination_table="sql_experiments_v1",  # TODO remove this line when removing the databricks job
+        arguments=[
+            "--partition-by",
+            "experiment_id",
+            "submission_date_s3",
+            "--select",
+            "*",
+            "'{{ds_nodash}}' AS submission_date_s3",
```
sunahsuh: Should this have a `--replace` flag before this line?

dags/main_summary.py
```diff
@@ -219,6 +223,59 @@
...
+            "--replace",
+            "'{{ds_nodash}}' AS submission_date",
+            "SAFE_CAST(sample_id AS INT64) AS sample_id",
```
sunahsuh: Same here re: `--replace`?

relud: `--replace` uses `nargs="+"`, so we specify the flag once, and 1 or more values

dags/main_summary.py
relud: this is an arg for `--select`, which also uses `nargs="+"`

PR #783, opened 2019-12-03
Patch GKEPodOperator to preserve XCOM

relud:


dags/operators/gcp_container_operator.py
```diff
@@ -52,7 +54,9 @@ def execute(self, context):
 
             # Tell `KubernetesPodOperator` where the config file is located
             self.config_file = os.environ[KUBE_CONFIG_ENV_VAR]
-            super(GKEPodOperator, self).execute(context)
+            result = super(UpstreamGKEPodOperator, self).execute(context)
```
relud: @jasonthomas without the alias to `UpstreamGKEPodOperator`, this was calling the overridden method `airflow.contrib.operators.gcp_container_operator.GKEPodOperator.execute`, instead of the method from `KubernetesOperator`

PR #785, opened 2019-12-04
Set XCOM_PUSH envvar to True for tasks using it

sunahsuh:
The other half of the fix we landed on for mozilla/bigquery-etl#553


dags/utils/gcp.py
```diff
@@ -711,14 +713,18 @@ def gke_command(
...
             key: value
             for key, value in zip(
                 ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"),
                 AwsHook(aws_conn_id).get_credentials() if aws_conn_id else (),
             )
-            if value is not None
-        },
+            if value is not None},
+            **{
+                "XCOM_PUSH": xcom_push
```
relud: we need to lower case this, json seems like the most succinct way to do that
```suggestion
                "XCOM_PUSH": json.dumps(xcom_push),
```

relud: i see now that it's checked uppercased, please ignore

dags/utils/gcp.py
```diff
@@ -711,14 +713,18 @@ def gke_command(
...
             key: value
             for key, value in zip(
                 ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN"),
                 AwsHook(aws_conn_id).get_credentials() if aws_conn_id else (),
             )
-            if value is not None
-        },
+            if value is not None},
+            **{
+                "XCOM_PUSH": json.dumps(xcom_push),
```
relud: this needs an `import json` at the top of the file, if you leave it in.

PR #791, opened 2019-12-05
Bug 1601139 - Add dag for materializing sample of documents

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1601139)

This schedules the query added in mozilla/bigquery-etl#570 on a daily basis. This will get used where `sanitized-landfill-v3` was used, for example integration testing of schemas.


dags/document_sample.py
```diff
@@ -0,0 +1,25 @@
...
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+dag = DAG("document_sample", default_args=default_args, schedule_interval="0 1 * * *")
+
+document_sample_nonprod_v1 = bigquery_etl_query(
+    task_id="document_sample_nonprod_v1",
```
relud: `task_id` defaults to `destination_table`
```suggestion
```

PR #793, opened 2019-12-09
Use only main ping view in update orphaning ETL

akkomar:
This removes usage of `main_1pct_backfill` table which is no longer
needed after main ping table has been backfilled.

This also adds a filter for `build.version` being not null. Such pings
should not be included in this analysis (we had a single ping having
this field empty in the last year).

### DIFF
--- a/jobs/update_orphaning_dashboard_etl.py
+++ b/jobs/update_orphaning_dashboard_etl.py
@@ -146,47 +146,10 @@ def longitudinal_shim_aggregate(date_from, date_to, destination_project, destina
             payload.histograms.update_status_error_code_partial_stage,
             payload.histograms.update_status_error_code_unknown_stage
         FROM
-            `moz-fx-data-shared-prod.static.main_1pct_backfill`
-        WHERE
-            sample_id = 42
-            AND DATE(submission_timestamp)>='2019-01-01'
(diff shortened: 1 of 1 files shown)


jobs/update_orphaning_dashboard_etl.py
```diff
@@ -186,7 +149,7 @@ def longitudinal_shim_aggregate(date_from, date_to, destination_project, destina
             `moz-fx-data-shared-prod.telemetry.main`
         WHERE
             sample_id = 42
-            AND DATE(submission_timestamp)>='2019-09-01' )
+            AND environment.build.version IS NOT NULL )
```
fbertsch: This is new, what's the rationale?

fbertsch: Also, will this run without a partition filter on `telemetry.main`?

akkomar: We had a single ping with NULLed version recently which caused this break. Such pings shouldn't be included here since we're specifically looking at various versions in this analysis.

akkomar: Yes, this line is a CTE, there is a partition filter couple o lines below.

PR #794, opened 2019-12-09
Add percentile and user count tasks to glam dag

emtwo:


dags/glam.py
```diff
@@ -135,6 +146,28 @@
...
+    project_id="moz-fx-data-shared-prod",
+    owner="msamuel@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "msamuel@mozilla.com"],
+    date_partition_parameter=None,
+    arguments=('--replace',),
+    dag=dag)
+
+user_counts = bigquery_etl_query(
+    task_id="user_counts",
+    destination_table="user_counts",
```
fbertsch: I think we should make this name clearer, and version it: `glam_user_counts_v1`?

dags/glam.py
```diff
@@ -135,6 +146,28 @@
     arguments=('--replace',),
     dag=dag)
 
+historgam_percentiles = bigquery_etl_query(
+    task_id="historgam_percentiles",
```
fbertsch: s/historgram/histogram

fbertsch:
It looks like percentile queries rely on `client_probe_counts_v1`, right?

emtwo:
> It looks like percentile queries rely on `client_probe_counts_v1`, right?

It's confusing, but scalar percentiles depend on `client_scalar_aggregates_v1` and histogram percentiles depend on `client_probe_counts_v1`

PR #796, opened 2019-12-10
Add tasks for running simpleprophet forecasts

jklukas:
Depends on https://github.com/mozilla/forecasting/pull/10

We aren't going to start relying on these "productionized" forecasts until beginning of 2020, so for now nothing depends on the forecast tasks and we leave the kpi_dashboard dag alone.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -3,6 +3,7 @@ import datetime
 from airflow import models
 from airflow.executors import GetDefaultExecutor
 from airflow.operators.subdag_operator import SubDagOperator
+from utils.forecasting import simpleprophet_forecast
 from utils.gcp import (bigquery_etl_copy_deduplicate,
                        bigquery_etl_query,
                        gke_command,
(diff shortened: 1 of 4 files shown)


dags/copy_deduplicate.py
```diff
@@ -175,7 +176,31 @@
...
+        task_id="simpleprophet_forecasts_mobile",
+        datasource="mobile",
+        project_id='moz-fx-data-shared-prod',
+        dataset_id='telemetry_derived',
+        table_id='simpleprophet_forecasts_mobile_v1',
+        owner="jklukas@mozilla.com",
+        email=["telemetry-alerts@mozilla.com", "jklukas@mozilla.com"],
+    )
+
+    simpleprophet_forecasts_mobile << firefox_nondesktop_exact_mau28
```
acmiyaguchi: Keeping the chain moving to the right might be more intuitive.

dags/utils/forecasting.py
```diff
@@ -0,0 +1,56 @@
...
+    gcp_conn_id="google_cloud_derived_datasets",
+    gke_location="us-central1-a",
+    gke_cluster_name="bq-load-gke-1",
+    gke_namespace="default",
+    docker_image="gcr.io/moz-fx-data-forecasting/simpleprophet:latest",
+    image_pull_policy="Always",
+    **kwargs
+):
+    """ Copy a day's data from live ping tables to stable ping tables,
+    deduplicating on document_id.
```
acmiyaguchi: This comment looks like it suffers from copy-paste.

PR #797, opened 2019-12-11
Schedule the exact_mau by client count dims queries

mreid-moz:
These are added as helpers for rewriting old queries that used
the HLL-based "client_count" datasets. Using this data makes
them faster and cheaper than using the "last_seen" tables
directly.

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -398,6 +398,14 @@ exact_mau_by_dimensions_export = SubDagOperator(
     executor=GetDefaultExecutor(),
     dag=dag)
 
+exact_mau_by_client_count_dimensions = bigquery_etl_query(
+    task_id="exact_mau_by_client_count_dimensions",
+    destination_table="firefox_desktop_exact_mau28_by_client_count_dimensions_v1",
+    dataset_id="telemetry_derived",
+    owner="mreid@mozilla.com",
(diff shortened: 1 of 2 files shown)


mreid-moz:
Data is backfilled up to `2019-12-10`

dags/main_summary.py
```diff
@@ -398,6 +398,14 @@
     executor=GetDefaultExecutor(),
     dag=dag)
 
+exact_mau_by_client_count_dimensions = bigquery_etl_query(
+    task_id="exact_mau_by_client_count_dimensions",
+    destination_table="firefox_desktop_exact_mau28_by_client_count_dimensions_v1",
+    dataset_id="telemetry_derived",
+    owner="mreid@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "mreid@mozilla.com"],
```
jklukas: I'd encourage you to make me or relud owner on these, as they're so similar to the existing exact_mau tables and I'd argue you should avoid dataset ownership in preference to management duties. I volunteer to be listed as owner.

jklukas:
Merging, as I have another PR to merge that's going to conflict with this one.

dags/main_summary.py
mreid-moz: I was going to ask this very question. I will update it in a follow-up.

PR #798, opened 2019-12-11
Fixups for wrong project ID in exact_mau28 by client count

jklukas:
Follow-up to #797


jklukas:
> D'oh, sorry about that!

Sorry for not catching it when I reviewed!

Issue #145, opened 2017-07-07
Run automated tests for basic python soundness


mreid-moz:
In order to avoid things like #144 we can have an automated test [run each dag as a python script](https://airflow.incubator.apache.org/tutorial.html#running-the-script) and fail if there are any errors.

acmiyaguchi:
Fixed by #804

PR #804, opened 2019-12-13
Add testing script to parse dags in CI

### LINKED ISSUE #789: Fail CI on non-parseable DAGs
We may have an issue to capture this already, but I merged in a PR today that broke a bunch of dags because of a syntax error (use of a python 3 feature.) We should at the very least be able to catch these syntax errors, and ideally run a smoke test task from the command line as part of CI

acmiyaguchi:
This fixes #789 and #145 by running the `bin/test-parse` script. This script sets up the locally hosted version of airflow and parses the logs for errors on `airflow list_dags`. CircleCI has been set up to run this script under a job named `test-environment`.

In addition to catching ill-specified Python, this will catch cases where dummy variables have not been added to the initialization scripts. 

This script can be run locally, in lieu of opening the airflow admin UI. This has been tested by adding a broken dag to the commit (ce87783) and letting CI fail. Removing this file fixes the job.

The job takes about 10 minutes in total (4 minutes for initial build, then 6 minutes for built + test).


.circleci/config.yml
```diff
@@ -21,6 +21,14 @@ jobs:
...
           printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' "$CIRCLE_SHA1" "$CIRCLE_TAG" "$CIRCLE_PROJECT_USERNAME" "$CIRCLE_PROJECT_REPONAME" "$CIRCLE_BUILD_URL" > version.json
       - run: docker build -t app:build .
+      - run:
+          name: Save image into cache
+          command: |
+            docker save -o /tmp/dev.tar "app:build"
+      - save_cache:
+          key: v1-{{ .Branch }}-{{ epoch }}
+          paths:
+            - /tmp/dev.tar
```
sunahsuh: If we use DLC we can remove this

.circleci/config.yml
```diff
@@ -33,6 +41,30 @@ jobs:
       - run: find . -name *.pyc -delete
       - run: tox -e py27
 
+  test-environment:
+    machine:
+      image: default
```
sunahsuh: As it appears you noticed, the default image is from 2017, it appears. This ubuntu image, according to https://circleci.com/docs/2.0/configuration-reference/#available-machine-images appears to be the most recent option
```suggestion
      image: ubuntu-1604:201903-01
      docker_layer_caching: true
```

.circleci/config.yml
```diff
@@ -33,6 +41,30 @@ jobs:
...
+    working_directory: ~/mozilla/telemetry-airflow
+    steps:
+      - checkout
+      - restore_cache:
+          key: v1-{{ .Branch }}
+      - run:
+          name: Restore cache
+          command: |
+            docker load -i /tmp/dev.tar
```
sunahsuh: We can remove these steps with DLC

.circleci/config.yml
```diff
@@ -33,6 +41,30 @@ jobs:
...
+      - run:
+          command: |
+            pip install --upgrade pip
+            pip install --upgrade docker-compose
```
sunahsuh: Don't need these with the ubuntu image, I think

.circleci/config.yml
```diff
@@ -88,6 +120,13 @@ workflows:
             tags:
               only: /.*/
 
+      - test-environment:
+          requires:
+            - build
```
sunahsuh: Don't need this anymore with DLC

acmiyaguchi:
Excellent, thanks for optimizing the CI process. The patch is more concise now and speedier to boot now,

PR #808, opened 2019-12-17
Bug 1583947: Add initial attitudes daily dag

sunahsuh:
I've already added these variables to the prod Airflow instance so this should be good to go – the dag will get filled out with a few more steps when those queries are ready

### DIFF
--- a/dags/attitudes_daily.py
+++ b/dags/attitudes_daily.py
@@ -38,6 +38,4 @@ with models.DAG(
             "--destination_table",
             "moz-fx-data-shared-prod.telemetry_derived.survey_gizmo_daily_attitudes"
         ],
-        docker_image="mozilla/bigquery-etl:latest",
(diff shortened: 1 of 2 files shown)


dags/attitudes_daily.py
```diff
@@ -0,0 +1,43 @@
...
+            "--sg_api_token",
+            Variable.get("surveygizmo_api_token"),
+            "--sg_api_secret",
+            Variable.get("surveygizmo_api_secret"),
+            "--destination_table",
+            "moz-fx-data-shared-prod.telemetry_derived.survey_gizmo_daily_attitudes"
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        owner="ssuh@mozilla.com",
+        email=["telemetry-alerts@mozilla.com", "ssuh@mozilla.com"])
```
acmiyaguchi: The owner and email may be redundant.

PR #809, opened 2019-12-18
Bug 1604934 - Add copy of prerelease aggregates dag for cloudsql

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1604934)

This copies the prerelease aggregates DAG and sets the start_date to 2019-12-01 and updates the variables for credentials with prefix `mozaggregator_cloudsql_*`.


haroldwoo:
this lgtm at a quick glance. I need to complete pr https://github.com/mozilla/telemetry-airflow/pull/811 first, so we can have the sparkrunners use NAT'd clusters (so the cluster can talk to postgres in gcp). will lyk once that is done and tested

dags/mozaggregator_prerelease_cloudsql.py
```diff
@@ -0,0 +1,92 @@
...
+# process via dataproc and write to cloudsql
+prerelease_telemetry_aggregate_view_dataproc_cloudsql = SubDagOperator(
+    task_id=task_id,
+    dag=dag,
+    subdag=moz_dataproc_pyspark_runner(
+        parent_dag_name=dag.dag_id,
+        dag_name=task_id,
+        job_name="prerelease_aggregates_cloudsql",
+        cluster_name="prerelease-telemetry-aggregates-cloudsql-{{ ds_nodash }}",
+        idle_delete_ttl="600",
```
haroldwoo: For NAT'd cluster with access to the gcp postgres, use these args here (once pr 811 lands)

      	zone='us-west2-a',
	subnetwork_uri='default',
        internal_ip_only=True

acmiyaguchi:
I'm assuming there's little tuning to be done in this case, so it'll be easier to turn on in production and watch from there. I've added this as a separate DAG so it can easily be turned off if things go awry.

PR #812, opened 2019-12-19
Remove gcs_to_s3 step from parquet export jobs

relud:


relud:
tested locally to confirm this doesn't break dags

PR #815, opened 2019-12-21
GCP port of taar-dynamo

crankycoder:
* added options for master and worker disk type and size
* dataproc should really be operated with only pd-ssd disk types and >
  1TB drives.
* added explicit 10 retry configuration for boto3+dynamodb
* added custom exponential backoff around the batch write
  as boto3 does not seem to backoff quite enough

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -300,20 +300,36 @@ devtools_panel_usage = bigquery_etl_query(
     start_date=datetime(2019, 11, 25),
     dag=dag)
 
-taar_dynamo = EMRSparkOperator(
-    task_id="taar_dynamo",
-    job_name="TAAR DynamoDB loader",
-    execution_timeout=timedelta(hours=14),
-    instance_count=6,
-    disable_on_dev=True,
-    owner="vng@mozilla.com",
-    email=["mlopatka@mozilla.com", "vng@mozilla.com", "sbird@mozilla.com"],
-    env=mozetl_envvar("taar_dynamo", {
-        "date": "{{ ds_nodash }}"
-    }),
-    uri="https://raw.githubusercontent.com/mozilla/python_mozetl/master/bin/mozetl-submit.sh",
-    output_visibility="private",
-    dag=dag)
+
+taar_dynamo_job = SubDagOperator(
+    task_id="taar_dynamo_job",
+    subdag=moz_dataproc_pyspark_runner(
+        parent_dag_name=dag.dag_id,
+        dag_name="taar_dynamo_job",
+        default_args=default_args,
+        master_machine_type='n1-standard-32',
+        worker_machine_type='n1-standard-32',
+        cluster_name=taar_dynamo_cluster_name,
+        job_name="TAAR_Dynamo",
+        #python_driver_code="gs://moz-fx-data-prod-airflow-dataproc-artifacts/jobs/taar_dynamo.py",
+        python_driver_code="gs://temp-hwoo-removemelater/taar_dynamo.py",
+        num_workers=12,
+        py_args=[
+            "--date",
+            "{{ ds_nodash }}",
+            "--aws_access_key_id",
+            taar_aws_access_key,
+            "--aws_secret_access_key",
+            taar_aws_secret_key,
+        ],
+        aws_conn_id=taar_aws_conn_id,
+        gcp_conn_id=taar_gcpdataproc_conn_id,
+        master_disk_type='pd-ssd',
+        worker_disk_type='pd-ssd',
+    ),
+    dag=dag,
+)
+
 
 
 taar_locale_job = SubDagOperator(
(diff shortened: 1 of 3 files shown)


crankycoder:
When resampling down to 0.001 - the job completes successfully so I'm pretty sure we've got the permissions issues sorted out at the very least.  

Here's the end of the job run showing write completed to AWS DynamoDB:

TAAR_Dynamo_440d8fa6_-_airflow-dataproc-prod_-_Google_Cloud_Platform_and___dev_telemetry-airflow_dags

jobs/taar_dynamo.py
```diff
@@ -0,0 +1,499 @@
...
+        )
+
+        table = conn.Table(self._table_name)
+
+        retry_exceptions = (
+            "ProvisionedThroughputExceededException",
+            "ThrottlingException",
+        )
+
+        retries = 0
```
crankycoder: @robotblake The `_write_batch` function was extracted so that it's more obvious that there is exponential backoff added around the boto3 write operation.

dags/utils/dataproc.py
```diff
@@ -44,6 +44,10 @@ def __init__(self,
                  gcp_conn_id='google_cloud_airflow_dataproc',
                  artifact_bucket='moz-fx-data-prod-airflow-dataproc-artifacts',
                  storage_bucket='moz-fx-data-prod-dataproc-scratch',
+                 master_disk_type='pd-standard',
+                 master_disk_size=1024,
```
haroldwoo: is this necessary as a default for all dataproc clusters launched by airflow? I would rather see a default of None here and set them if provided, that way other current job behavior doesn't change

dags/main_summary.py
```diff
@@ -29,6 +28,7 @@
 taarlite_cluster_name = "dataproc-taarlite-guidguid"
 taar_locale_cluster_name = "dataproc-taar-locale"
 taar_gcpdataproc_conn_id = "google_cloud_airflow_dataproc"
+taar_dynamo_cluster_name = "dataproc-taar-dynamo"
```
haroldwoo: Can we give these taar clusters more descriptive names? There are 3 now, and I cant remember what they all do

dags/utils/dataproc.py
```diff
@@ -178,6 +191,10 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
                                 gcp_conn_id='google_cloud_airflow_dataproc',
                                 artifact_bucket='moz-fx-data-prod-airflow-dataproc-artifacts',
                                 storage_bucket='moz-fx-data-prod-dataproc-scratch',
+                                master_disk_type='pd-standard',
```
haroldwoo: Add descriptions of the new parameters to the comments starting on line 233.  Also you may want to add these parameters to the other 2 runners (jar runner, script runner)

jobs/taar_dynamo.py
```diff
@@ -0,0 +1,499 @@
...
+
+PATCH_DAYS = timedelta(days=2)
+
+# We use the os and threading modules to generate a spark worker
+# specific identity:w
+
+MAX_RECORDS = 200
+EMPTY_TUPLE = (0, 0, [], [])
+
+BOTO_CFG = botocore.config.Config(retries={"max_attempts": 16})
```
haroldwoo: make max_attempts configurable?

dags/utils/dataproc.py
crankycoder: That seems reasonable - defaults are defined in the `DataprocClusterCreateOperator` already - I'm just duplicating those settings from :  https://github.com/mozilla/telemetry-airflow/blob/master/dags/operators/moz_dataproc_operator.py#L197

I've made the patch in 24dc037

dags/main_summary.py
crankycoder: How do you want to name these?  I'm just using the literal job name and prefixing it with `dataproc-` for cluster names.

haroldwoo: hmm if this cluster is loading dynamo then maybe dataproc-taar-dynamo-loader. Just some descriptor of what the cluster is responsible for doing

dags/utils/dataproc.py
haroldwoo: ah this was the default, my mistake.

jobs/taar_dynamo.py
crankycoder: I don't think there's any advantage to making this configurable.  

The primary problem we see in practice is cross cloud network throttling.  To work around that - I've put in an exponential back-off over here: https://github.com/mozilla/telemetry-airflow/pull/815/files#diff-3f86d29de1b5bb05b05e589750cae922R221

crankycoder:
@robotblake can you follow up with jason?  I don't know what this CircleCI test-environment failure is about - it doesn't look like my code.

crankycoder:
CircleCI test-environment breakage appears to be: https://github.com/mozilla/telemetry-airflow/pull/815

PR #819, opened 2019-12-30
 Bug 1606361 - Add alternative code path for reading from avro

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1606361)

This adds a code-path for dumping data to avro and using this within the `python_mozaggregator` job.


acmiyaguchi:
This is a good catch, I remember having tested this operator before, I'll modify the DAG to take advantage of the operator instead.

acmiyaguchi:
I took a closer look at the BigQueryToCloudStorageOperator. While this operator works well when we're dumping a partition from a table, some of the jobs specify a channel clause to filter out data to prerelease. This would require reimplementing most of the logic

akkomar:
Yeah, then you'd need to do something similar to Amplitude jobs: https://github.com/mozilla/telemetry-airflow/blob/0bf6891fb6250b79577c2c54f84497fb953d3388/dags/utils/amplitude.py#L66-L83

acmiyaguchi:
It's also a stop-gap solution while the Storage API issues on the `payload_bytes_decoded` data-set are being investigated, which doesn't have the associated 10TB export limit associated with it.

PR #826, opened 2020-01-08
Add saved_session export for mozaggregator_prerelease

acmiyaguchi:
This PR exports the saved_session ping for prerelease aggregates. This also adds an additional namespace to the path so the mobile aggregates and prerelease aggregates can run without interference of each other.

### DIFF
--- a/dags/mozaggregator_mobile.py
+++ b/dags/mozaggregator_mobile.py
@@ -116,7 +116,7 @@ if EXPORT_TO_AVRO:
             "bin/export-avro.sh",
             "moz-fx-data-shared-prod",
             "moz-fx-data-shared-prod:analysis",
-            "gs://moz-fx-data-derived-datasets-parquet-tmp/avro/mozaggregator",
+            "gs://moz-fx-data-derived-datasets-parquet-tmp/avro/mozaggregator/mobile",
             "mobile_metrics_v1",
             '""',
             "{{ ds }}",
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
Logs for the job that this PR would fix: https://workflow.telemetry.mozilla.org/log?task_id=run_dataproc_pyspark&dag_id=prerelease_telemetry_aggregates.prerelease_telemetry_aggregate_view_dataproc&execution_date=2020-01-07T00%3A00%3A00%2B00%3A00

dags/mozaggregator_mobile.py
```diff
@@ -116,7 +116,7 @@
             "bin/export-avro.sh",
             "moz-fx-data-shared-prod",
             "moz-fx-data-shared-prod:analysis",
-            "gs://moz-fx-data-derived-datasets-parquet-tmp/avro/mozaggregator",
+            "gs://moz-fx-data-derived-datasets-parquet-tmp/avro/mozaggregator/mobile",
```
akkomar: Shouldn't this also be included in Dataproc task parameter? (https://github.com/mozilla/telemetry-airflow/blob/9ff864b68240bef41900b1c59114a26fc2c8b45f/dags/mozaggregator_mobile.py#L99)

acmiyaguchi: Yes, thank you for catching this.

PR #834, opened 2020-01-14
Add probe expiry alert task to probe scraper

Ben-Wu:


dags/probe_scraper.py
```diff
@@ -91,3 +91,22 @@
...
+        project_id=connection.project_id,
+        location=gke_location,
+        cluster_name=gke_cluster_name,
+        name="probe-expiry-alerts",
+        namespace='default',
+        image=probe_scraper_image,
+        arguments=["python3", "-m", "probe_scraper.probe_expiry_alert"],
+        email=["bewu@mozilla.com"],
+        env_vars={
+            "AWS_ACCESS_KEY_ID": aws_access_key,
```
fbertsch: We don't need these, do we?

Ben-Wu: Don't we need it for the emailer?

fbertsch: Indeed we do. Carry on 😄

PR #835, opened 2020-01-15
Remove unnecessary image_pull_policy params

fbertsch:
We are now default to "Always" for image_pull_policy,
so there's no reason to add that again to every util.

### DIFF
--- a/dags/probe_scraper.py
+++ b/dags/probe_scraper.py
@@ -81,7 +81,6 @@ with DAG('probe_scraper',
         name='schema-generator-1',
         namespace='default',
         image='mozilla/mozilla-schema-generator:latest',
-        image_pull_policy='Always',
         env_vars={
             "MPS_SSH_KEY_BASE64": "{{ var.value.mozilla_pipeline_schemas_secret_git_sshkey_b64 }}",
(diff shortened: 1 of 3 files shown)


fbertsch:
Note that this doesn't prevent people from adding `image_pull_policy="..."` to any of these utils or dags, since it would still propagate with `kwargs` to the GKE operator.

dags/probe_scraper.py
```diff
@@ -81,7 +81,6 @@
         name='schema-generator-1',
         namespace='default',
         image='mozilla/mozilla-schema-generator:latest',
-        image_pull_policy='Always',
```
haroldwoo: Just a suggestion that it may be helpful for this GkePodOperator to keep this param so that others writing new dags can reference it.

fbertsch: Hmm, can you clarify? The `GkePodOperator` does have this param now

haroldwoo: Just if anyone wanted to modify image_pull_policy, and were to git grep for image_pull_policy, they wouldn't find anything. But that's prob okay if they reference the upstream code for whatever deploy tag we're on

PR #837, opened 2020-01-16
taar_similarity ported out of databricks

crankycoder:
WIP

taar_similarity migration to run in GCP instead of Databricks

The DAG will need to be modified similar to this the one linked below.

boto requires the authentication credentials to be explicitly passed in or it will use the wrong set of AWS credentials to write to AWS+S3 for the final output.

https://github.com/mozilla/telemetry-airflow/blob/master/dags/main_summary.py#L334

### DIFF
--- a/jobs/taar_similarity.py
+++ b/jobs/taar_similarity.py
@@ -1,0 +1,514 @@
+"""
+Bug 1386274 - TAAR similarity-based add-on donor list
+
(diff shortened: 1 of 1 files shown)


akkomar:
Moving to https://github.com/mozilla/telemetry-airflow/pull/839

PR #841, opened 2020-01-22
Fix TAAR similarity for reading Parquet from GCS

akkomar:


jobs/taar_similarity.py
```diff
@@ -168,7 +168,6 @@ def get_samples(spark, date_from, clients_daily_gcs_bucket):
         .where("size(active_addons) < 100")
         .where("channel = 'release'")
         .where("app_name = 'Firefox'")
-        .where("submission_date_s3 >= {}".format(date_from))
```
akkomar: `submission_date_s3` is not present in this dataset because we are reading only a selected day using path prefix (`"gs://{}/clients_daily/v6/submission_date_s3={}"` above).

crankycoder:
lgtm

PR #843, opened 2020-01-22
Airflow 1.10.7 upgrade

haroldwoo:
TODO:
- [x] fix initial parse issues with all dags

Changelogs:
- [x] go over changelogs. Looks like these are of note, although we don't rely on most of these:
- add envFrom for Kubernetes Executor
- KubernetesPodOperator support for envFrom configMapRef/secretRef. Also some improvements with display of pod resources, etc.
- Dataflow hook pinned to py2. Ability to add labels.
- GoogleCloudStorageToBigQueryOperator will be renamed GCSToBigQueryOperator in the future (not in 1.10.7). Also had some fixes (autodetect, encoding param, etc.).
- Add support CSV files in MySqlToGoogleCloudStorageOperator
- Add run_as_user and fs_group options for Kubernetes
- [x] Changes to Google Cloud Transfer Service Operators - e.g. changes to s3_to_gcs_transfer_operator.py which we rely on (https://github.com/apache/airflow/pull/4792), timeout parameter added, moved source to airflow.contrib.operators.gcp_transfer_operator
- Add Google Cloud Translate operator (do we have a use case for this?)
- Operators for S3 object copying/deleting
- Add compose to GoogleCloudStorageHook (not used currently)
- make number of dataproc masters configurable
-  Add more metrics around the scheduler (airflow scheduler, although we don't scrape metrics in a prom fashion, and google doesnt either?)
- [x]  Use GOOGLE_APPLICATION_CREDENTIALS constant from library (https://github.com/apache/airflow/pull/5744) - We fork GKEPodOperator to set CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE to use service account key jsons to issue gcloud commands, plus some xcom changes
- [x] Fix Databricks hook leaks API secret in logs - We only use the token for local dev testing. bin/run reflects the change. Prod jobs use the WTMO connection which uses basic user/pass auth
- KubernetesPodOperator fixes and test
- Environment variables not correctly set in SparkSubmitHook 
- [x] Add ability to specify optional components in DataprocClusterCreateOperator (we were doing this via a fork, so eliminate duplicate code where possible) - https://github.com/apache/airflow/pull/5821/files

review code for the following between 1.10.2 and 1.10.7
- [x] dags/operators/gcp_container_operator (diff 1.10.2 and 1.10.7, copypasta the 1.10.7 one into our own fork and add our changes again (https://github.com/mozilla/telemetry-airflow/blame/master/dags/operators/gcp_container_operator.py). This functionality will migrate with newer versions as this is deprecated in master.
- [x] dags/utils/dataproc.py -> check bc we import the dataproc operators from 1.10.2 (i know that optional components is now supported for example). pysparkop changes look ok

- [x] dags/operators/gcs.py -> https://github.com/apache/airflow/blob/a87f3dfc0c513b05271cba7051c18e4e742ef245/airflow/gcp/operators/gcs.py#L321 + changes for 1.10.2 /py2 , might be able to delete
- [x] dags/operators/moz_dataproc_operator.py -> can we delete this now? -> no bc we use component gateway. updated all refs
- [x] git grep s3_to_gcs_transfer_operator - there are changes in (https://github.com/apache/airflow/pull/4792)
- [x] documentation  - update readme on how to troubleshoot test-parse environment circleci failures
- [x] rm dags/operators/dataproc_hadoop_with_aws.py - no longer needed

Test:
- [x] GkePodOperator - clean gke pods should suffice
- [x] DataprocOperator - verify component gateway installed on new cluster
- [x] DataprocOperator - test pyspark runner
- [x] DataprocOpertaor - test jar runner
- [x] DataprocOperator - test scriptrunner
- [x] dags/utils/dataproc.py -> Test dataproc runners without changing main_class to main_jar (this was a bug found in earlier dataproc testing, that may have been fixed) for jar and script runner.
- [x] test s3togcs transfer job operators
- [x] dags/utils/gcp.py -> also imports dataproc operators and gkepodoperator, also some bigquery operators.  git grep utils.gcp for a list of dags using pieces of this. Assuming the dags load, and diffs don't look bad, rely on devs to keep an eye on these after upgrade
- [x] utils/amplitude - some bqtocloudstoageoperators. Rely on devs to keep an eye on. Diffs are minimal also

Other
- Makefile clean target cleans out local logs
- Add timeout to s3togcs operators because it now defaults to 60s.
- Add common defaults to our GkePodOperator to simplify usage, and rm'd usage from dags
- Add pod name label (without the uuid) to pod
- Using upstream gcs_delete_operator and removing our version
- Random other changes to reflect upstream changes


Dockerfile
```diff
@@ -54,9 +55,10 @@ ENV PYTHONUNBUFFERED=1 \
     # DEPLOY_TAG = \
     # ARTIFACTS_BUCKET = \
     # DATABRICKS_DEFAULT_IAM \
-    PORT=8000
+
```
haroldwoo: moved env vars around to avoid empty continuation line warnings when building dockerfiles.

Dockerfile.dev
```diff
@@ -67,14 +67,15 @@ ENV AIRFLOW_HOME=/app \
     AIRFLOW_DATABASE_URL=mysql://root:secret@db:3306/airflow \
     AIRFLOW__CORE__FERNET_KEY="0000000000000000000000000000000000000000000=" \
     AIRFLOW_SECRET_KEY="000000000000000000000000000000000000000000=" \
-    # AIRFLOW_SMTP_HOST= \
-    # AIRFLOW_SMTP_USER= \
-    # AIRFLOW_SMTP_PASSWORD= \
     AIRFLOW_SMTP_FROM=telemetry-alerts@airflow.dev.mozaws.net \
     AIRFLOW__SCHEDULER__CATCHUP_BY_DEFAULT=False \
     AIRFLOW_EMAIL_BACKEND="airflow.macros.log_email_backend.log_email_backend" \
+    AIRFLOW__KUBERNETES__IN_CLUSTER=False\
```
haroldwoo: in_cluster previously defaulted to False in 1.10.2, but this changed with 1.10.7 (bugs), and I added a default in our GKEPodOperator as well as this env config because some code paths check env and others use the argument iirc

bin/run
```diff
@@ -74,7 +74,7 @@ init_connections() {
         --conn_id databricks_default \
         --conn_type databricks \
         --conn_host https://dbc-caf9527b-e073.cloud.databricks.com \
-        --conn_extra "{\"token\":\"${DB_TOKEN}\"}"
+        --conn_extra "{\"token\":\"${DB_TOKEN}\", \"host\": \"\"}"
```
haroldwoo: some databricks operator change

dags/adi_by_region.py
```diff
@@ -57,6 +54,7 @@
     project_id=connection.project_id,
     object_conditions=gcstj_object_conditions,
     transfer_options=gcstj_transfer_options,
+    timeout=720,
```
haroldwoo: class was moved and added a default timeout of 60s when set to None

dags/operators/gcp_container_operator.py
```diff
@@ -80,57 +111,13 @@ def _set_env_from_extras(self, extras):
         if not key_path and not keyfile_json_str:
             self.log.info('Using gcloud with application default credentials.')
         elif key_path:
-            os.environ[G_APP_CRED] = key_path
-            os.environ[GCLOUD_APP_CRED] = key_path
+            os.environ[CREDENTIALS] = key_path
+            return None
```
haroldwoo: we dont hit this code path, but i left this line here to preserve diffs with upstream

dags/prio/kubernetes.py
```diff
@@ -3,11 +3,9 @@
 
 from airflow import DAG
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
-from operators.gcp_container_operator import (
-    GKEClusterCreateOperator,
-    GKEClusterDeleteOperator,
-    GKEPodOperator,
-)
+from airflow.contrib.operators.gcp_container_operator import GKEClusterCreateOperator, GKEClusterDeleteOperator
```
haroldwoo: upstreams were fixed

bin/run
haroldwoo: Appears to be the case according to https://github.com/apache/airflow/pull/5635/files#diff-53a81d33b2423e1a6b28c14b7e5642d1

PR #854, opened 2020-02-12
Add tasks to extract GLAM data to CSVs

robhudson:


dags/glam.py
```diff
@@ -262,6 +267,34 @@
     arguments=('--append_table','--noreplace',),
     dag=dag)
 
+sql_file_path = "sql/{}/{}/query.sql".format(dataset_id, "glam_client_probe_counts_extract_v1")
```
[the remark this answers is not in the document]
robhudson: Ah, I thought that was the path so it knows where to find the SQL in bigquery-etl?

dags/glam.py
```diff
@@ -15,6 +19,7 @@
     'retries': 1,
     'retry_delay': timedelta(minutes=30),
 }
+glam_bucket = "gs://glam-dev-bespoke-nonprod-dataops-mozgcp-net",
```
acmiyaguchi: This is invalid syntax.

```suggestion
glam_bucket = "gs://glam-dev-bespoke-nonprod-dataops-mozgcp-net"
```

dags/glam.py
```diff
@@ -262,6 +267,33 @@
...
+    dataset_id=dataset_id,
+    project_id="moz-fx-data-shared-prod",
+    owner="robhudson@mozilla.com",
+    email=["telemetry-alerts@mozilla.com", "msamuel@mozilla.com", "robhudson@mozilla.com"],
+    date_partition_parameter=None,
+    arguments=('--replace',),
+    dag=dag)
+
+glam_extract_clear_bucket = GoogleCloudStorageDeleteOperator(
+    task_id="glam_gcs_delete_old_extracts",
```
emtwo: I think maybe `task_id` needs to match the variable for `GoogleCloudStorageDeleteOperator`.

So `task_id` would also be `glam_extract_clear_bucket` or vice versa.

PR #867, opened 2020-02-27
Add temporary job to backfill mozaggregator trimmed database

haroldwoo:


dags/mozaggregator_prerelease_new.py
```diff
@@ -0,0 +1,185 @@
...
+        gcp_conn_id=gcp_conn.gcp_conn_id,
+        service_account=client_email,
+        artifact_bucket=artifact_bucket,
+        storage_bucket=storage_bucket,
+        default_args=subdag_args,
+    ),
+)
+
+
+"""
```
haroldwoo: commented out the avro stuff

dags/mozaggregator_prerelease_new.py
```diff
@@ -0,0 +1,185 @@
...
+# and then once we're caught up, I simply plan to switch DNS to point the production mozaggregator to the new trimmed database
+# (will also redeploy mozagg prod to get rid of datadog)
+EXPORT_TO_AVRO = True
+
+default_args = {
+    "owner": "frank@mozilla.com",
+    "depends_on_past": True,
+    # TODO - double check? our last run before snapshot was for runid 2/19/2020 but started 2/20/2020 0 UTC
+    # or 2/19/2020 at 4pm PST
+    "start_date": datetime(2020, 02, 20),
```
haroldwoo: Anthony, was the last date you saw when checking the data 2/19?

acmiyaguchi: Yes, `20200219` is the last date seen.

dags/mozaggregator_prerelease_new.py
```diff
@@ -0,0 +1,185 @@
...
+from airflow import DAG
+from airflow.contrib.operators.gcs_delete_operator import GoogleCloudStorageDeleteOperator
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.operators.subdag_operator import SubDagOperator
+from utils.dataproc import moz_dataproc_pyspark_runner, copy_artifacts_dev
+from utils.gcp import gke_command
+
+# TODO - should i disable this? since this job will only backfill the lean database
+# and then once we're caught up, I simply plan to switch DNS to point the production mozaggregator to the new trimmed database
+# (will also redeploy mozagg prod to get rid of datadog)
```
acmiyaguchi: I think disabling it when it's caught up is fine.

dags/mozaggregator_prerelease_new.py
acmiyaguchi: The avro stuff is necessary, the job doesn't run via the Storage API with BigQuery.

dags/mozaggregator_prerelease_new.py
```diff
@@ -0,0 +1,181 @@
...
+
+# export to avro, if necessary
+if EXPORT_TO_AVRO:
+    gke_command(
+        task_id="export_main_avro",
+        cmds=["bash"],
+        command=[
+            "bin/export-avro.sh",
+            "moz-fx-data-shared-prod",
+            "moz-fx-data-shared-prod:analysis",
```
acmiyaguchi: I hope nothing weird happens here, because the same table gets reused for exports. This takes 10 minutes, so hopefully the new job and the old job dont run exactly in parallel.

dags/mozaggregator_prerelease_new.py
```diff
@@ -0,0 +1,181 @@
...
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
+}
+
+dag = DAG(
+    "prerelease_telemetry_aggregates_new",
+    default_args=default_args,
+    schedule_interval="@daily",
```
acmiyaguchi: Is it possible to offset this by an hour?

dags/mozaggregator_prerelease_new.py
acmiyaguchi: With respect to this comment, the way to avoid table re-use is to specify a different dataset, possibly `moz-fx-data-shared-prod:mozaggregator-backfill-tmp`, which does not overlap with `analysis`.

acmiyaguchi: See: https://github.com/mozilla/python_mozaggregator/blob/207068afed33c592f1d49904b16881910459b878/bin/export-avro.sh#L59-L61

PR #882, opened 2020-03-06
Factor out clients_daily_* as subdags and generate these queries daily.

emtwo:
This PR is mainly doing 2 things:

1. All clients daily processing now generates the sql queries on the fly that it needs before running them. This means that new probes added would be included daily (https://github.com/mozilla/bigquery-etl/issues/760).

2. Factors out `clients_daily_*` tasks into scalar and histogram subdags

### DIFF
--- a/dags/glam_subdags/daily_aggregates.py
+++ b/dags/glam_subdags/daily_aggregates.py
@@ -1,0 +1,114 @@
+from airflow import DAG
+
+from utils.gcp import gke_command, bigquery_xcom_query
+
+dags = {}
+
+
+def generate_tasks_for_clients_daily(
+    metric_type, agg_type, dataset_id, dag, arguments=()
+):
+    gen_query_task_id = "clients_daily_{}_aggregates_generate_query".format(agg_type)
+
+    # setting xcom_push to True outputs this query to an xcom
+    generate_query = gke_command(
+        task_id=gen_query_task_id,
+        command=[
+            "python",
+            "sql/telemetry_derived/clients_daily_{}_aggregates_v1.sql.py".format(
+                metric_type
+            ),
+            "--agg-type",
+            agg_type,
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        xcom_push=True,
+        owner="msamuel@mozilla.com",
+        email=[
+            "telemetry-alerts@mozilla.com",
+            "msamuel@mozilla.com",
+            "robhudson@mozilla.com",
+        ],
+        dag=dags[dag],
+    )
+
+    run_query = bigquery_xcom_query(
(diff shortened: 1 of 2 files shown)


emtwo:
Screen Shot 2020-03-06 at 11 50 10 AM
Screen Shot 2020-03-06 at 11 49 58 AM
Screen Shot 2020-03-06 at 11 49 42 AM

dags/glam_subdags/daily_aggregates.py
```diff
@@ -0,0 +1,114 @@
...
+        dag=dags[dag],
+    )
+    generate_query >> run_query
+    return run_query
+
+
+def daily_scalar_aggregates_subdag(
+    parent_dag_name, child_dag_name, default_args, schedule_interval, dataset_id
+):
+    GLAM_DAILY_SCALAR_AGGREGATES_SUBDAG = "%s.%s" % (parent_dag_name, child_dag_name)
```
robhudson: Or as an f-string:

`     GLAM_DAILY_SCALAR_AGGREGATES_SUBDAG = f"{parent_dag_name}.{child_dag_name}"

acmiyaguchi: Airflow is still on python 2.7, so no f-strings yet.

acmiyaguchi:
I'd recommend deploying this today, to avoid failures over the weekend.

emtwo:
> The use of xcom is new to me, but it looks like this should work as intended based on [the experiment aggregates query](https://github.com/mozilla/telemetry-airflow/commit/d0de0993a163dde0a93f9ed08d5f3a0dfe085fd1).

Indeed that is where I borrowed the idea from :) 

> It's probably best to have a single source of truth. The generated copies in the `sql` directory of the `bigquery_etl` project are certainly going to go stale since the schema is constantly evolving. On the other hand, it's useful to have a concrete query around, since the diffs can be used to review the changes between versions of the query. What are your thoughts on the checked-in queries?

I was planning on getting rid of the checked in queries. They would definitely be stale. I think if we want to look at diffs we can generate them when needed.

emtwo:
> I'd recommend deploying this today, to avoid failures over the weekend.

I'm not sure what you mean. Did you mean to *not* deploy them today in case they cause failures over the weekend?

acmiyaguchi:
> > I'd recommend deploying this today, to avoid failures over the weekend.
> 
> I'm not sure what you mean. Did you mean to _not_ deploy them today in case they cause failures over the weekend?

Yes, the edited comment reads "I'd recommend not deploying this today". I dropped an important word there.

emtwo:
> Yes, the edited comment reads "I'd recommend not deploying this today". I dropped an important word there.

lol ok cool. yes I agree.

PR #883, opened 2020-03-09
Add DAG for running pensieve

scholtzan:
To schedule pensieve in Airflow. https://github.com/mozilla/pensieve/issues/4

### DIFF
--- a/dags/pensieve.py
+++ b/dags/pensieve.py
@@ -4,41 +4,53 @@ from datetime import timedelta, datetime
 from operators.gcp_container_operator import GKEPodOperator
 
 default_args = {
-    'owner': ['ascholtz@mozilla.com', 'ssuh@mozilla.com', 'tdsmith@mozilla.com'],
-    'depends_on_past': False,
-    'start_date': datetime(2020, 3, 9),
-    'email_on_failure': True,
-    'email_on_retry': True,
-    'retries': 2,
-    'retry_delay': timedelta(minutes=30),
+    "owner": "ascholtz@mozilla.com",
+    "email": [
+        "ascholtz@mozilla.com",
+        "ssuh@mozilla.com",
+        "tdsmith@mozilla.com",
+    ],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 3, 12),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
 }
 
-with DAG('pensieve',
-         default_args=default_args,
-         schedule_interval='0 1 * * *') as dag:
+with DAG(
+    "pensieve", default_args=default_args, schedule_interval="0 1 * * *"
+) as dag:
 
     # Built from repo https://github.com/mozilla/pensieve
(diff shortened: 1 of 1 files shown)


dags/pensieve.py
```diff
@@ -0,0 +1,59 @@
...
+        task_id="wait_for_event_events",
+        external_dag_id="main_summary",
+        external_task_id="event_events",
+        dag=dag,
+    )
+
+    pensieve.set_upstream(wait_for_clients_daily_export)
+    pensieve.set_upstream(wait_for_main_summary_export)
+    pensieve.set_upstream(wait_for_search_clients_daily)
+    pensieve.set_upstream(wait_for_event_events)
```
scholtzan: I'm not completely sure if those are correct. I looked at the datasources that mozanalysis uses for its analysis: https://github.com/mozilla/mozanalysis/blob/561e1ad650d81a940562e7090b5ca5757f87c55d/src/mozanalysis/metrics/desktop.py

dags/pensieve.py
```diff
@@ -0,0 +1,60 @@
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+from datetime import timedelta, datetime
+from operators.gcp_container_operator import GKEPodOperator
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+
+default_args = {
+    'owner': ['ascholtz@mozilla.com', 'sunahsuh@mozilla.com', 'tdsmith@mozilla.com'],
```
sunahsuh:
```suggestion
    'owner': ['ascholtz@mozilla.com', 'ssuh@mozilla.com', 'tdsmith@mozilla.com'],
```

dags/pensieve.py
sunahsuh: At least for the metrics in that file it looks like we need clients daily, search clients daily, and events.

dags/pensieve.py
```diff
@@ -0,0 +1,60 @@
...
+        task_id="pensieve",
+        name='pensieve',
+        image=pensieve_image,
+        email=['ascholtz@mozilla.com', 'sunahsuh@mozilla.com', 'tdsmith@mozilla.com'],
+        dag=dag)
+
+    wait_for_clients_daily_export = ExternalTaskSensor(
+        task_id="wait_for_clients_daily_export",
+        external_dag_id="main_summary",
+        external_task_id="clients_daily_export",
```
sunahsuh: I think all of these jobs are actually scheduled to start at 01:00 UTC (using a schedule_interval of `0 1 * * *`) so these task sensors either need to include an execution delta like [this](https://github.com/mozilla/telemetry-airflow/blob/02ab6b7d489e7de545913c46938a5c85fb7e07db/dags/addons_daily.py#L31) or (and I'd recommend this just to keep things simple) we change the dag's schedule_interval from `@daily` to `0 1 * * *`

sunahsuh:
Hmm, let's figure out what's up with this test before merging too

dags/pensieve.py
scholtzan: oops

scholtzan:
I finally figured out why the test was failing. The DAG `owner` arg doesn't allow a list of owners

PR #884, opened 2020-03-09
Update extracts to extract per channel

robhudson:


dags/glam_subdags/extract.py
```diff
@@ -0,0 +1,69 @@
...
+from airflow.contrib.operators.gcs_delete_operator import (
+    GoogleCloudStorageDeleteOperator,
+)
+from airflow.models import DAG
+from utils.gcp import bigquery_etl_query
+
+
+gcp_conn = GoogleCloudBaseHook("google_cloud_airflow_dataproc")
+project_id = ("moz-fx-data-shared-prod",)
+glam_bucket = "glam-dev-bespoke-nonprod-dataops-mozgcp-net"
```
emtwo: You can remove this from `dags/glam.py` then?

dags/glam.py
```diff
@@ -306,7 +306,6 @@
 
 clients_scalar_aggregates >> glam_user_counts
 
-glam_user_counts >> glam_client_probe_counts_extract
-histogram_percentiles >> glam_client_probe_counts_extract
-glam_client_probe_counts_extract >> glam_gcs_delete_old_extracts
-glam_gcs_delete_old_extracts >> glam_extract_to_csv
+glam_user_counts >> extract_nightly
+histogram_percentiles >> extract_nightly
+extract_nightly >> extract_beta >> extract_release
```
emtwo: You can all 3 extracts in parallel right? It would be verbose to do that here because it would be 

```
glam_user_counts >> extract_nightly
histogram_percentiles >> extract_nightly
glam_user_counts >> extract_beta
histogram_percentiles >> extract_beta
glam_user_counts >> extract_release
histogram_percentiles >> extract_release
```

However, if these 3 are in that subdag we talked about over slack, the one subdag can happen after and run all 3 in parallel.

PR #889, opened 2020-03-11
Set GOOGLE_CLOUD_PROJECT for parquet2bigquery load operations

jasonthomas:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1612541#c12

r?

### DIFF
--- a/dags/main_summary.py
+++ b/dags/main_summary.py
@@ -355,6 +355,8 @@ bgbb_pred_bigquery_load = SubDagOperator(
         cluster_by=["sample_id"],
         rename={"submission_date_s3": "submission_date"},
         replace=["SAFE_CAST(sample_id AS INT64) AS sample_id"],
+        env_vars={"GOOGLE_CLOUD_PROJECT":
+                  "{{ var.value.gcp_shared_prod_project }}"},
         ),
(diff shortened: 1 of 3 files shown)


dags/main_summary.py
```diff
@@ -355,6 +355,8 @@
         cluster_by=["sample_id"],
         rename={"submission_date_s3": "submission_date"},
         replace=["SAFE_CAST(sample_id AS INT64) AS sample_id"],
+        env_vars={"GOOGLE_CLOUD_PROJECT":
+                  "{{ var.value.gcp_shared_prod_project }}"},
```
jklukas: Should we be using `{{ var.value.gcp_shared_prod_project }}` rather than `moz-fx-data-shared-prod` everywhere in the DAGs?

jasonthomas: Maybe. I think it might be a good pattern to adopt if we ever decide to make use of our stage environment for dag testing. It might be useful for local testing too.

I don't have a real strong opinion here and I could just change this to conform to what we have been doing.

relud: +1 for using `{{ var.value.gcp_shared_prod_project }}`

PR #894, opened 2020-03-11
Add Messaging System tasks

ncloudioj:
@jklukas r?

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -115,6 +115,74 @@ with models.DAG(
 
     copy_deduplicate_all >> impression_stats_flat
 
+    # Derived tables for messaging-system.
+
+    # CFR
+    messaging_system_cfr_users_daily = bigquery_etl_query(
+        task_id='messaging_system_cfr_users_daily',
+        project_id='moz-fx-data-shared-prod',
+        destination_table='cfr_users_daily_v1',
+        dataset_id='messaging_system_derived',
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -115,6 +115,71 @@
...
+        email=['najiang@mozilla.com'],
+    )
+
+    messaging_system_cfr_users_last_seen = bigquery_etl_query(
+        task_id='messaging_system_cfr_users_last_seen',
+        project_id='moz-fx-data-shared-prod',
+        destination_table='cfr_users_last_seen_v1',
+        dataset_id='messaging_system_derived',
+        email=['najiang@mozilla.com'],
+    )
```
jklukas:
```suggestion
        depends_on_past=True,
    )
```

The last_seen queries all need to be marked as `depends_on_past`.

ncloudioj: Ah, good catch. Thanks! Always wonder why it's not the default behavior.

jklukas: > Always wonder why it's not the default behavior.

Because it's a pain ;) When depends_on_past is False, backfills can proceed in parallel and a failure one day doesn't have any impact on subsequent days. When depends_on_past is true, there's a certain amount of handholding that needs to happen when things aren't running smoothly. So it's nice to avoid it, but in the last_seen case it's necessary.

PR #895, opened 2020-03-11
Fix #891 - Use snake_cased cluster definition to match protobuf

### LINKED ISSUE #891: prio-processor dag fails to start gke cluster
Logs:
```
[2020-03-10 23:15:39,120] {logging_mixin.py:112} INFO - Running %s on host %s <TaskInstance: prio_processor.bootstrap.create_gke_cluster 2020-02-09T17:29:33.628418+00:00 [running]> airflow-prod-airflow-app-1-worker-0.airflow-prod-airflow-app-1-worker.prod-airflow.svc.cluster.local
[2020-03-10 23:15:39,152] {taskinstance.py:1088} ERROR - 'dict' object has no attribute 'name'
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/airflow/models/taskinstance.py", line 955, in _run_raw_task
    result = task_copy.execute(context=context)
  File "/usr/local/lib/python2.7/site-packages/airflow/contrib/operators/gcp_container_operator.py", line 177, in execute
    self._check_input()
  File "/usr/local/lib/python2.7/site-packages/airflow/contrib/operators/gcp_container_operator.py", line 168, in _check_input
    elif self.body.name and self.body.initial_node_count:
AttributeError: 'dict' object has no attribute 'name'
```

The underlying airflow code has changed, causing failures in starting a cluster.

acmiyaguchi:
This fixes #891. When tested locally, the DAG now fails in a new location.

### DIFF
--- a/dags/utils/gke.py
+++ b/dags/utils/gke.py
@@ -49,7 +49,7 @@ def create_gke_config(
                     "service_account": service_account,
                     "labels": {"owner": owner_label, "team": team_label},
                     "preemptible": preemptible,
-                    "diskType": disk_type,
+                    "disk_type": disk_type,
                 },
                 "initial_node_count": 1,
                 "autoscaling": {
--- a/dags/utils/gke.py
+++ b/dags/utils/gke.py
@@ -13,27 +13,31 @@ def create_gke_config(
 ):
 
     """
-    Helper function to create gke cluster definition dict.
-    See https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster
+    Helper function to create gke cluster definition dict. All fields must match
+    their protobuf definitions.
 
-    owner and team labels  can contain only lowercase letters, numeric characters,
+    See:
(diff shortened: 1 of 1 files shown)


dags/utils/gke.py
```diff
@@ -13,27 +13,31 @@ def create_gke_config(
...
+        https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster
+        https://googleapis.dev/python/container/latest/gapic/v1/types.html#google.cloud.container_v1.types.Cluster
+    owner and team labels can contain only lowercase letters, numeric characters,
     underscores, and dashes. E.g. owner_label='hwoo', team_label='dataops'
 
     """
 
     cluster_def_dict = {
         "name": name,
+        "initial_node_count": None,
```
acmiyaguchi: The `GKEClusterCreateOperator` looks for this field, otherwise fails to startup. However, if it's set to an integer, then it will fail saying that it can't be set when `node_pools` are defined.

PR #897, opened 2020-03-12
Fix upstream dependencies for pensieve

scholtzan:
Currently, the dependencies look like this:

Screen Shot 2020-03-12 at 09 29 10

This PR fixes this by correctly setting multiple upstream tasks as dependencies


scholtzan:
I accidentally ran `black` on this, hence the formatting changes

jklukas:
_Does_ it change the tree view? The image you pasted looks like what I would expect to see in the tree viewer.

I don't think this PR does any harm, but I'm interested to see if it has any effect on visualization.

PR #900, opened 2020-03-12
Schedule shredder to run every 28 days

relud:
I am currently manually running this, but if I stop the manual runs then shredder will pick up where it left off, so long as it runs with `--start-date=2019-11-20` and `--end-date=2020-03-09`.

### DIFF
--- a/dags/shredder.py
+++ b/dags/shredder.py
@@ -1,0 +1,66 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from utils.gcp import gke_command
+
+default_args = {
+    "owner": "dthorn@mozilla.com",
+    "depends_on_past": True,
+    "start_date": datetime(2020, 3, 9),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "dthorn@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 5,
+    "retry_delay": timedelta(minutes=5),
+}
+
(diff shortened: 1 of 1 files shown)


dags/shredder.py
```diff
@@ -0,0 +1,66 @@
...
+    "retries": 2,
+    "retry_delay": timedelta(minutes=5),
+}
+
+dag = DAG("shredder", default_args=default_args, schedule_interval=timedelta(days=28))
+docker_image = "mozilla/bigquery-etl:latest"
+base_command = [
+    "script/shredder_delete",
+    "--state-table=moz-fx-data-shredder.shredder_state.shredder_state",
+    "--start-date=2019-11-20",
```
acmiyaguchi: Does it matter that the `start-date` is fixed for all runs in the future?

relud: This should be fine for now, but it will need to be updated in the future.

I talked with mreid about this, and we agreed that it should cover the last 2 cycles of deletion requests (i.e. run every 4 weeks, but delete 8 weeks of requests), but I think that's going to require adding an option like `--preceding-days`, rather than using `--start-date`.

relud:
> How does shredder pick up where it last left off?

it records state with `--state-table=moz-fx-data-shredder.shredder_state.shredder_state`

dags/shredder.py
relud: fixed this to use `2019-11-20` for `ds == '2020-03-09'` and otherwise use 8 weeks before `ds`

PR #902, opened 2020-03-13
Move dashboard and amplitude ETL to shared-prod tables

jklukas:
We should be able to merge these changes immediately. They are internal details that don't need to hit derived-datasets anymore.

### DIFF
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -27,7 +27,7 @@ with models.DAG(
             dag_name=fenix_task_id,
             parent_dag_name=dag_name,
             default_args=default_args,
-            project='moz-fx-data-derived-datasets',
+            project='moz-fx-data-shared-prod',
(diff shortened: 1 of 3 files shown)


dags/bq_events_to_amplitude.py
```diff
@@ -27,7 +27,7 @@
             dag_name=fenix_task_id,
             parent_dag_name=dag_name,
             default_args=default_args,
-            project='moz-fx-data-derived-datasets',
+            project='moz-fx-data-shared-prod',
```
jklukas: `telemetry.fenix_events_v1` is a view published to both projects, so this should have no effect.

jklukas:
From what I can tell, the hardware report and update orphaning dashboard jobs overwrite a table in the `analysis` dataset simply to stage data, and then they export that data to GCS. So the location of the table looks like it is an internal detail that doesn't affect any other jobs or use cases.

PR #903, opened 2020-03-13
Update bucket paths to include protocol

acmiyaguchi:
This should fix remaining issues with the currently broken prio processor dag. This will close https://bugzilla.mozilla.org/show_bug.cgi?id=1609548.


acmiyaguchi:
Closed in favor of #945

PR #904, opened 2020-03-16
Update addons_daily dag to call bigquery-etl query

benmiroglio:
The current addons_daily DAG has been turned off since the GCP migration. I've started work to migrate this query from spark to BQ in https://github.com/mozilla/bigquery-etl/pull/819.

I updated the DAG using the [bigquery-airflow cookbook](https://docs.telemetry.mozilla.org/cookbooks/bigquery-airflow.html) as a guide, however I'm not confident that everything looks as it should.

@jklukas Tapping you to take a look since you've seen the bigquery-etl query :).

### DIFF
--- a/dags/addons_daily.py
+++ b/dags/addons_daily.py
@@ -1,62 +1,42 @@
-from airflow import DAG
+from airflow import models
+from utils.gcp import bigquery_etl_query
 from airflow.operators.sensors import ExternalTaskSensor
-from datetime import datetime, timedelta
-from airflow.operators.moz_databricks import MozDatabricksSubmitRunOperator
-from utils.mozetl import mozetl_envvar
+import datetime
 
 default_args = {
     "owner": "bmiroglio@mozilla.com",
-    "depends_on_past": True,
-    "start_date": datetime(2019, 6, 10),
-    "email": [
-        "telemetry-alerts@mozilla.com",
-        "bmiroglio@mozilla.com",
-        "smelancon@mozilla.com",
-        "bwright@mozilla.com",
-        "dthorn@mozilla.com",
-    ],
+    "start_date": datetime.datetime(2020, 3, 16),
+    "email": ["telemetry-alerts@mozilla.com", "bmiroglio@mozilla.com"],
     "email_on_failure": True,
     "email_on_retry": True,
-    "retries": 2,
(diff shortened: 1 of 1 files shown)


jklukas:
This needs to depend on clients_last_seen and search_clients_daily. It might make sense to put this query into the main_summary DAG so it's easier to express the dependency on those two queries. Otherwise, we'll need to add ExternalTaskSensor to wait for the two queries in the other DAG.

benmiroglio:
I added an `ExternalTaskSensor` for those datasets, following the pattern in the [attitudes_daily] DAG.

dags/addons_daily.py
```diff
@@ -1,62 +1,40 @@
...
-        "dthorn@mozilla.com",
-    ],
+    "start_date": datetime.datetime(2020, 3, 16),
+    "email": ["telemetry-alerts@mozilla.com", "bmiroglio@mozilla.com"],
     "email_on_failure": True,
     "email_on_retry": True,
-    "retries": 2,
-    "retry_delay": timedelta(minutes=30),
+    "depends_on_past": False,
+    # If a task fails, retry it once after waiting at least 5 minutes
```
jklukas:
```suggestion
    # If a task fails, retry it once after waiting at least 30 minutes
```

dags/addons_daily.py
```diff
@@ -1,62 +1,40 @@
...
-    "depends_on_past": True,
-    "start_date": datetime(2019, 6, 10),
-    "email": [
-        "telemetry-alerts@mozilla.com",
-        "bmiroglio@mozilla.com",
-        "smelancon@mozilla.com",
-        "bwright@mozilla.com",
-        "dthorn@mozilla.com",
-    ],
+    "start_date": datetime.datetime(2020, 3, 16),
```
jklukas: You're missing the import for this.

jklukas:
This looks to be in the right shape now. If you can fix the import and get the tests all passing, we should be ready to merge as soon as https://github.com/mozilla/bigquery-etl/pull/819 is merged.

jklukas:
Looks like we have rebase conflicts here. Can you handle that, @benmiroglio? Let me know if you need help.

benmiroglio:
@jklukas Sorry, I'm not sure where to start here.

PR #906, opened 2020-03-17
Fix upstream task IDs for pensieve DAG

scholtzan:
The naming of two upstream tasks pensieve depends on were incorrect causing the pensieve DAG to never get executed.


dags/pensieve.py
```diff
@@ -45,14 +45,14 @@
...
-        external_task_id="search_clients_daily",
+        external_task_id="search_clients_daily_bigquery",
         dag=dag,
     )
 
     wait_for_event_events = ExternalTaskSensor(
         task_id="wait_for_event_events",
         external_dag_id="main_summary",
-        external_task_id="event_events",
+        external_task_id="bq_main_events",
```
sunahsuh: You probably want to make this also dependent on event_events in the copy_deduplicate DAG https://github.com/mozilla/telemetry-airflow/blob/master/dags/copy_deduplicate.py#L44 (events used to be in the main ping but they were moved into the event ping a while back, but we still get ~150k events from old clients in the main ping per day). Re: experiments, they're most likely not running on the older clients, so it's more important to make it dependent on event_events than on main_events

scholtzan: Added this one as dependency

PR #907, opened 2020-03-17
Glam's clients_histogram_aggregates is partitioned by date.

emtwo:


dags/glam_subdags/general.py
```diff
@@ -23,12 +25,10 @@ def repeated_subdag(
         owner="msamuel@mozilla.com",
         email=["telemetry-alerts@mozilla.com", "msamuel@mozilla.com"],
         depends_on_past=True,
-        date_partition_parameter=None,
         parameters=(
             "min_sample_id:INT64:0",
             "max_sample_id:INT64:{}".format(PARTITION_SIZE - 1),
         ),
-        arguments=("--replace",),
```
acmiyaguchi: The first partition should still be replaced, in order to stay idempotent.

emtwo: I had incorrectly assumed the partition replacement was done by default but this arg was only needed for replacing a full table.

In this case, I added the replace arg to clients_daily_* as well

PR #911, opened 2020-03-19
Add operator for publishing query results as JSON

scholtzan:
This will run the bigquery-etl `publish_public_data_json` script after running a BigQuery query to check if the query results should be uploaded as JSON to a public bucket.

Depends on https://github.com/mozilla/bigquery-etl/pull/839

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -488,7 +488,7 @@ def bigquery_etl_query(
     if destination_table is not None and date_partition_parameter is not None:
         destination_table = destination_table + "${{ds_nodash}}"
         parameters += (date_partition_parameter + ":DATE:{{ds}}",)
-    return GKEPodOperator(
+    etl_operator = GKEPodOperator(
         gcp_conn_id=gcp_conn_id,
         project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
         location=gke_location,
@@ -504,6 +504,20 @@ def bigquery_etl_query(
         + [sql_file_path],
         **kwargs
     )
+
+    publish_json_operator = publish_json(
+        destination_table=kwargs["task_id"],
+        dataset_id=dataset_id,
+        parameters=parameters,
+        arguments=arguments,
+        project_id=project_id,
+        sql_file_path=sql_file_path,
+        **kwargs
+    )
+
+    publish_json_operator.set_upstream(etl_operator)
+
+    return etl_operator
 
 
 def bigquery_etl_copy_deduplicate(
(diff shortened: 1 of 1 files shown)


dags/utils/gcp.py
```diff
@@ -700,3 +714,65 @@ def gke_command(
...
+    :param Optional[str] date_partition_parameter: Parameter for indicating destination
+                                                   partition to generate, if None
+                                                   destination should be whole table
+                                                   rather than partition
+    :param Dict[str, Any] kwargs:                  Additional keyword arguments for
+                                                   GKEPodOperator
+    """
+
+    kwargs["task_id"] = "publish_json_" + destination_table
+    kwargs["name"] = kwargs["task_id"].replace("_", "-")[:63]
```
scholtzan: We need to ensure that name is <64 characters long. By prepending `publish_json_` some tests failed because the name was too long.

dags/utils/gcp.py
```diff
@@ -505,6 +505,20 @@ def bigquery_etl_query(
         **kwargs
     )
 
+    publish_json_operator = publish_json(
```
jklukas: Am I correct in understanding that this will add a publish_json operator for every query but in most cases it simply checks the metadata file, determines it's not public, and shuts down?

If so, that's going to add a lot of noise to the DAGs and potentially have performance implications for tracking all the extra tasks. For the time being, I think it would be preferable to add publish_json explicitly for each public table, and we defer having a fully-metadata-based solution until we're generating DAGs in bigquery-etl.

fbertsch: A simpler alternative is add an argument here for publishing json data.

dags/utils/gcp.py
fbertsch: Can you maybe add a comment here, and make `64` a constant?

dags/utils/gcp.py
scholtzan: Right, it would probably be more convenient and intuitive if users would not need to set the `public_json` flag in the metadata file and additionally change the task in Airflow to set the flag. So, I think the better way for now is to execute both scripts in the bigquery-etl entrypoint (at least until the DAG generation stuff is in place)

scholtzan: On second thought, if publishing json is part of the `bigquery_etl_query` tasks, then we'd need to use the highmem node pool for all of them. Those machines seem to be a bit more expensive. So, maybe in the meantime setting an argument for publishing json is the way to go. I suspect there won't be too many use cases for datasets to be published until the lower-friction scheduled queries get implemented (which should happen before/by the end of H1)

PR #914, opened 2020-03-20
 Bug 1623728 - Schedule LTV Task in Airflow for Dataset Creation

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1623728)

```bash
make build
make up
```

Create the following script:

```bash
#!/bin/bash

set -eou pipefail
set -x

function format_gcp() {
KEYFILE="$1" python3 - <<END
import os
import json

with open(os.environ["KEYFILE"]) as f:
    data = json.load(f)

extra = {
    "extra__google_cloud_platform__project": data["project_id"],
    "extra__google_cloud_platform__keyfile_dict": json.dumps(data)
}

print(json.dumps(extra))
END
}


function update_gcp() {
    local conn_id=$1
    local keyfile=$2

    container_id=$(docker ps | grep telemetry-airflow_web | cut -d' ' -f1)

    docker exec $container_id \
        airflow connections -d --conn_id $conn_id
    
    docker exec $container_id \
        airflow connections -a \
            --conn_id $conn_id \
            --conn_type google_cloud_platform \
            --conn_extra "$(format_gcp $keyfile)"
}

update_gcp google_cloud_airflow_dataproc <PATH TO YOUR KEYFILE>
```

Run it:

```bash
chmod +x update_dataproc.sh
```

I modified py_args in the task to be the following:

```python
        py_args=[
            "--submission-date",
            "{{ ds }}",
            "--project-id",
            "amiyaguchi-dev",
            "--source-qualified-table-id",
            "amiyaguchi-dev.analysis.test_search_rfm",
            "--dataset-id",
            "analysis",
            "--temporary-gcs-bucket",
            params.storage_bucket,
        ],
```

I created `amiyaguchi-dev.analysis.test_search_rfm` by running the following query and using it as the destination table:

```sql
SELECT
  * EXCEPT(client_id),
  GENERATE_UUID() AS client_id
FROM
  `moz-fx-data-shared-prod.search.search_rfm`
WHERE
  submission_date = date "2020-03-19"
LIMIT
  10000
```

I opened up the airflow UI at http://localhost:8000/admin/airflow/tree?dag_id=ltv_daily and enabled it.

The first job copies the `jobs` folder into a bucket in the service account project. This should be rerun anytime the job file changes i.e. `ltv_daily.py`. Then the actual task runs on dataproc. If the job fails, the cluster will shut down after 5 minutes. In the meanwhile, when debugging, the zoomed in subdag task for `run_dataproc_pyspark` can be manually cleared to rerun the job.

image

The `jobs/ltv_daily.py` can still be run via the original bash script, along with the new arguments.


dags/ltv.py
```diff
@@ -0,0 +1,89 @@
...
+            "moz-fx-data-shared-prod.search.search_rfm",
+            "--dataset-id",
+            "analysis",
+            "--intermediate-table-id",
+            "temporary_search_rfm_day",
+            "--model-input-table-id",
+            "ltv_daily_model_perf_script",
+            "--model-output-table-id",
+            "ltv_daily_script",
+            "--temporary-gcs-bucket",
```
acmiyaguchi: @benmiroglio Are these arguments fine for the scheduled job. These can be written to another dataset and/or tables if you'd like.

dags/ltv.py
```diff
@@ -0,0 +1,89 @@
...
+        "amiyaguchi@mozilla.com",
+        "bmiroglio@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
+}
+
+dag = DAG("ltv_daily", default_args=default_args, schedule_interval="@daily")
```
acmiyaguchi: @benmiroglio Are there any dependencies for this job that need to be taken into consideration?

jobs/ltv_daily.py
```diff
@@ -0,0 +1,206 @@
...
+    model_pred_data = None
+    pred_metrics = ["days_searched", "days_tagged_searched", "days_clicked_ads"]
+    search_rfm_ds = search_rfm_full.limit(TRAINING_SAMPLE).select(columns).toPandas()
+    for metric in pred_metrics:
+        print(metric)
+        # train and extract model performace
+        model_perf, model = train_metric(search_rfm_ds, metric, plot=False, penalty=0.8)
+        model_perf["pct"] = model_perf.Model / (model_perf.Actual + 1) - 1
+        model_perf["metric"] = metric
+        model_perf["date"] = submission_date
```
acmiyaguchi: @benmiroglio This job had `d` as the variable that was set here. Is `submission_date` the correct variable to use?

jobs/ltv_daily.py
```diff
@@ -0,0 +1,206 @@
...
+    parser.add_argument("--dataset-id", default="bmiroglio")
+    parser.add_argument("--intermediate-table-id", default="search_rfm_day")
+    parser.add_argument("--model-input-table-id", default="ltv_daily_model_perf_script")
+    parser.add_argument("--model-output-table-id", default="ltv_daily_script")
+    parser.add_argument(
+        "--temporary-gcs-bucket", default="moz-fx-data-bq-data-science-bmiroglio"
+    )
+    args = parser.parse_args()
+
+    main(**vars(args))
```
acmiyaguchi: @benmiroglio Do you might taking a copy of this script and running it locally, just to verify that I havent functionally changed anything in the course of writing this?

dags/ltv.py
benmiroglio: This lgtm, can the table just be called `ltv_daily`?

dags/ltv.py
benmiroglio: the only dependent dataset is`search.search_rfm`.

dags/ltv.py
acmiyaguchi: It looks like there are two tables that the script generates, you're referring to `ltv_daily_script`?

jobs/ltv_daily.py
benmiroglio: yes, this should be `submission_date`

jobs/ltv_daily.py
benmiroglio: I'll try to give a whirl now.

dags/ltv.py
acmiyaguchi: Do you know if that table is scheduled in airflow, so it can be set as a dependency? A quick cursory glance at the `main_summary` dag doesn't turn up anything.

dags/ltv.py
```diff
@@ -0,0 +1,102 @@
...
+
+wait_for_search_clients_last_seen = ExternalTaskSensor(
```
fbertsch: I think this should live after search_clients_last_seen directly. That dataset and this should live together.

jobs/ltv_daily.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # flatten list
+    columns = [item for sublist in columns for item in sublist]
+
+    model_perf_data = pd.DataFrame()
+    model_pred_data = None
+    pred_metrics = ["days_searched", "days_tagged_searched", "days_clicked_ads"]
+    search_rfm_ds = search_rfm_full.limit(training_sample).select(columns).toPandas()
+    for metric in pred_metrics:
+        print(metric)
```
fbertsch: Probably can drop this print statement

jobs/ltv_daily.py
```diff
@@ -0,0 +1,206 @@
...
+        .option("table", f"{project_id}.{dataset_id}.{model_output_table_id}")
+        .option("temporaryGcsBucket", temporary_gcs_bucket)
+        .mode("overwrite")
+        .save()
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(__doc__)
+    parser.add_argument("--submission-date", default="2020-03-03")
```
fbertsch: Removing a default for this argument seems wise.

PR #916, opened 2020-03-23
Use Auto Zone Placement for Dataproc clusters

akkomar:
On 2020-03-23 Hardware report job is consistently failing with `The zone 'projects/REDACTED/zones/us-west1-b' does not have enough resources available to fulfill the request.  '(resource type:compute)'.`.
We should be able to avoid this error by using Auto Zone Placement.

### DIFF
--- a/dags/utils/dataproc.py
+++ b/dags/utils/dataproc.py
@@ -27,7 +27,7 @@ class DataProcHelper:
                  cluster_name=None,
                  num_workers=2,
                  image_version='1.4',
-                 zone='us-west1-b',
+                 region='us-west1',
                  subnetwork_uri=None,
                  internal_ip_only=None,
                  idle_delete_ttl='14400',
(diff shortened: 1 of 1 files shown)


acmiyaguchi:
I tested this locally using the `mobile_aggregates.mobile_aggregates_view_dataproc` task. As-is, this breaks the dataproc workflow, since the actual task looks for a cluster in the default "global" region.

```
HttpError: <HttpError 404 when requesting https://dataproc.googleapis.com/v1beta2/projects/amiyaguchi-dev/regions/global/jobs?clusterName=mobile-metrics-aggregates-20200324&alt=json returned "Not found: Cluster projects/amiyaguchi-dev/regions/global/clusters/mobile-metrics-aggregates-20200324">
```

I've added the `region` field to the relevant `DataProcSparkOperator` calls.

akkomar:
Thanks @acmiyaguchi! I'm going to merge this since that was also suggested in support ticket.

PR #920, opened 2020-03-25
Remove destination table from migrated_clients

fbertsch:


dags/incline_dash.py
```diff
@@ -37,10 +37,11 @@
 
     migrated_clients = bigquery_etl_query(
         task_id="generate_migrated_clients",
-        destination_table=table,
+        destination_table=None,
         project_id=project,
         dataset_id=dataset,
         date_partition_parameter=None,
+        parameters=('--nouse_legacy_sql'),
```
jklukas: This should already be handled by the `.bigqueryrc` installed in the docker image.

fbertsch: Ah, that's why it didn't work locally.

PR #921, opened 2020-03-25
Update incline export with new changes

fbertsch:
After some discussion with Hamilton:
- Export entire table of data
- Use CSV instead of JSON
- Export to /latest and /ds
- Add .gz ending to files


fbertsch:
Closing this pending a new serving model. Will update with changes.

PR #922, opened 2020-03-27
Schedule copy deviations task

scholtzan:
This should run the query in https://github.com/mozilla/bigquery-etl/pull/859/files on a daily basis and overwrite the destination table.

cc @jmccrosky

### DIFF
--- a/dags/analysis.py
+++ b/dags/analysis.py
@@ -17,7 +17,7 @@ default_args = {
     "retry_delay": timedelta(minutes=30),
 }
 
-with DAG("public_analysis", default_args=default_args, schedule_interval="0 4 * * *") as dag:
+with DAG("public_analysis", default_args=default_args, schedule_interval="0 6 * * *") as dag:
     deviations = bigquery_etl_query(
         destination_table="deviations_v1",
         dataset_id="telemetry_derived",
--- a/dags/analysis.py
+++ b/dags/analysis.py
@@ -17,10 +17,9 @@ default_args = {
     "retry_delay": timedelta(minutes=30),
 }
 
-with DAG("analysis", default_args=default_args, schedule_interval="0 1 * * *") as dag:
+with DAG("public_analysis", default_args=default_args, schedule_interval="0 4 * * *") as dag:
     deviations = bigquery_etl_query(
         destination_table="deviations_v1",
-        dataset_id="analysis",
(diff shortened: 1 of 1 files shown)


dags/analysis.py
```diff
@@ -0,0 +1,26 @@
...
+        "telemetry-alerts@mozilla.com",
+        "jmccrosky@mozilla.com",
+        "ascholtz@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("analysis", default_args=default_args, schedule_interval="0 1 * * *") as dag:
```
scholtzan: @jmccrosky As I understand, the data is updated by a databricks notebook on a daily basis. So this task will copy the data from the "internal" to the public table at 1am every day. Does that timing look correct, or is the internal table updated at different times so that the most recent results might be missing?

jmccrosky: Do you know what time clients daily is generally up to date?  I guess I'd run it shortly after that.

jklukas: I think the ideal flow here would be that we have a task here that waits on clients_daily, then triggers the databricks notebook to run via the databricks API and waits for that job to complete before running the query task. I'm not aware of any prior art for that, though. It's likely out of scope here.

We can check in Airflow when clients_daily tends to be done. Looks like normally 03:00 UTC should be a safe time to start the notebook, although an extra hour of padding would be nice if we can allow it. I don't know how long the notebook generally takes to run.

jklukas: Can we give this DAG a more specific name? Perhaps it can be "public_analysis"?

dags/analysis.py
```diff
@@ -0,0 +1,25 @@
...
+        "telemetry-alerts@mozilla.com",
+        "jmccrosky@mozilla.com",
+        "ascholtz@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("public_analysis", default_args=default_args, schedule_interval="0 6 * * *") as dag:
```
scholtzan: The databricks notebook gets triggered at 3am UTC and runs 1-2 hours (if I understood correctly). Running this at 6am should be safe enough?

PR #924, opened 2020-03-29
Fixing Backfill ui plugin

haroldwoo:
Modified from: https://github.com/AnkurChoraywal/airflow-backfill-util

todo 
- [ ] after merges, create another release 1.10.7.2 with container changes and redeploy to stage/prod. then notify users
- [x] fix infinite loop when clearing tasks
- [x] reverse streaming logs
- [x] find out why certain output is not shown
- [x] add dry runs for clear/backfill

### DIFF
--- a/plugins/backfill/LICENSE.md
+++ b/plugins/backfill/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Ankur Choraywal (Email: ankur@choraywal.in)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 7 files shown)


plugins/backfill/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
```
haroldwoo: is this how im supposed to add his license file?

requirements.txt
```diff
@@ -18,3 +18,5 @@ websocket-client>=0.32.0,!=0.40.0,!=0.41.*,!=0.42.* # LGPLv2+
 # Pin to older version, newer version has issues
 JPype1==0.7.1
 shelljob==0.5.6
+popen
+subprocess.run
```
jasonthomas, line 22: Are these needed here? subprocess and popen (which is part of subprocess) should be part of the standard libraries.

haroldwoo: thanks. was testing fixes and forgot to rm.

PR #927, opened 2020-04-01
Fixup addon_names task

jklukas:


dags/main_summary.py
```diff
@@ -166,7 +166,7 @@
     # previous day's addon data in main pings, thus the odd combination of
     # parameters below.
     date_partition_parameter=None,
-    parameters=("submission_date:DATE:{{ds}}"),
+    parameters=["submission_date:DATE:{{ds}}"],
```
jklukas, line 169: This needs to be an iterable, and I was intending for it to be a tuple, but forgot that parentheses only make a tuple when at least one comma appears inside. Changing to a list.

PR #929, opened 2020-04-01
TAAR ensemble fix

crankycoder:
This fixes the TAAR ensemble job after the GCP migration.

The two major changes are:

* no more logging in recommenders as they are not serializable by python pickle until Python 3.8.  This caused crashes in PySpark workers
* added explicit AWS keys into a context object that is passed from the PySpark master to workers as environment variables are not properly configured to use the `airflow_taar_rw_s3` connection.

Both these changes are embedded in taar 0.4.12 which is imported by the DAG and are not visible in this PR.

I've added support for local SSDs to keep runtime down to ~30 minutes given 35 x n1-standard-8 nodes with 2 SSDs per node.

### DIFF
--- a/dags/operators/moz_dataproc_operator.py
+++ b/dags/operators/moz_dataproc_operator.py
@@ -91,7 +91,7 @@ class DataprocClusterCreateOperator(DataprocOperationBaseOperator):
     :type master_disk_type: str
     :param master_disk_size: Disk size for the master node
     :type master_disk_size: int
-    :param master_num_local_ssds : Number of local SSDs to mount
+    :param master_num_local_ssds : Number of local SSDs to mount.  Local SSDs are used for writing and reading Apache Hadoop and Apache Spark scratch files, such as shuffle outputs. Adding SSDs will improve Spark runtime performance.
         (default is 0)
     :type master_num_local_ssds : int
     :param worker_machine_type: Compute engine machine type to use for the worker nodes
@@ -103,7 +103,7 @@ class DataprocClusterCreateOperator(DataprocOperationBaseOperator):
     :type worker_disk_type: str
     :param worker_disk_size: Disk size for the worker nodes
     :type worker_disk_size: int
-    :param worker_num_local_ssds : Number of local SSDs to mount
+    :param worker_num_local_ssds : Number of local SSDs to mount.  Local SSDs are used for writing and reading Apache Hadoop and Apache Spark scratch files, such as shuffle outputs.  Adding SSDs will improve Spark runtime performance.
         (default is 0)
     :type worker_num_local_ssds : int
     :param num_preemptible_workers: The # of preemptible worker nodes to spin up
--- a/dags/operators/moz_dataproc_operator.py
+++ b/dags/operators/moz_dataproc_operator.py
@@ -91,6 +91,9 @@ class DataprocClusterCreateOperator(DataprocOperationBaseOperator):
     :type master_disk_type: str
     :param master_disk_size: Disk size for the master node
     :type master_disk_size: int
+    :param master_num_local_ssds : Number of local SSDs to mount
+        (default is 0)
+    :type master_num_local_ssds : int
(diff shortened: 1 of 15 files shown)


crankycoder:
Output from a job run on a local Airflow instance can be found at:  s3://telemetry-parquet/taar/ensemble/ensemble_weight_daily20200322.json

```
{
  "ensemble_weights": {
    "collaborative": 2.4237995141351956,
    "similarity": 1.0295617216484558,
    "locale": 1.726434159924938
  }
}
```

dags/operators/moz_dataproc_operator.py
```diff
@@ -91,6 +91,9 @@ class DataprocClusterCreateOperator(DataprocOperationBaseOperator):
     :type master_disk_type: str
     :param master_disk_size: Disk size for the master node
     :type master_disk_size: int
+    :param master_num_local_ssds : Number of local SSDs to mount
```
haroldwoo, line 94: Please add a more descriptive comment on why we would ever want to set these. And what a defualt of 0 means

dags/utils/dataproc.py
```diff
@@ -279,13 +288,19 @@ def moz_dataproc_pyspark_runner(parent_dag_name=None,
     :type master_disk_type: str
     :param master_disk_size:              Disk size for the master node
     :type master_disk_size: int
+    :param master_num_local_ssds : Number of local SSDs to mount
```
haroldwoo, line 291: Thanks for adding this to all 3 runner methods. Can we add more description as to why we'd ever want to set these, and what a default of 0 means

dags/taar_weekly.py
```diff
@@ -6,55 +6,77 @@
 from datetime import datetime, timedelta
 from airflow.operators.moz_databricks import MozDatabricksSubmitRunOperator
 from utils.mozetl import mozetl_envvar
+from airflow.operators.subdag_operator import SubDagOperator
```
haroldwoo, line 9: nit: imports arent grouped together or in any alphabetical ordering

dags/taar_weekly.py
```diff
@@ -6,55 +6,77 @@
...
+from utils.dataproc import moz_dataproc_pyspark_runner
+from airflow.contrib.hooks.aws_hook import AwsHook
+
+taar_aws_conn_id = "airflow_taar_rw_s3"
+taar_aws_access_key, taar_aws_secret_key, session = AwsHook(
```
haroldwoo, line 14: nit: this fits on one line?

dags/taar_weekly.py
```diff
@@ -6,55 +6,77 @@
...
-    external_dag_id="main_summary",
-    external_task_id="clients_daily_export",
-    execution_delta=timedelta(
-        days=-7, hours=-1
-    ),  # main_summary waits one hour, execution date is beginning of the week
-    dag=taar_weekly,
-)
-
-
-taar_ensemble = MozDatabricksSubmitRunOperator(
```
haroldwoo, line 37: rip db

dags/taar_weekly.py
crankycoder: Thx.  Fixed in 49c1c930

dags/operators/moz_dataproc_operator.py
crankycoder: Added better doc strings: e04d7375

dags/utils/dataproc.py
crankycoder: Added better doc strings: e04d7375

dags/taar_weekly.py
crankycoder: Fixed in 4363acb

crankycoder: Crap.  I had to revert in 4363acb7 as it broke the DAG definition in https://github.com/mozilla/telemetry-airflow/pull/929/files#diff-00957536695958afe22bc38f1597e7c4R71

PR #935, opened 2020-04-06
Update incline export with new changes

fbertsch:
- Export entire table of data
- Use CSV instead of JSON
- Export to /latest and /ds
- Add .gz ending to files
- Export entire table to non-partitioned tmp table
- Export table for each country

### DIFF
--- a/dags/incline_dash.py
+++ b/dags/incline_dash.py
@@ -24,7 +24,6 @@ with DAG('incline_dashboard',
          default_args=default_args,
          schedule_interval="0 1 * * *") as dag:
 
-
     wait_for_copy_deduplicate = ExternalTaskSensor(
         task_id="wait_for_copy_deduplicate",
         external_dag_id="copy_deduplicate",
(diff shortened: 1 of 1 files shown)


dags/incline_dash.py
```diff
@@ -59,49 +58,62 @@
...
+        create_table = BigQueryOperator(
+            task_id='create_temporary_table',
+            sql=sql,
+            destination_dataset_table=fully_qualified_tmp_table,
+            bigquery_conn_id=gcp_conn_id,
+            use_legacy_sql=False
+        )
+
+        gcs_bucket = 'moz-fx-data-prod-analysis'
+        incline_prefix = 'incline/executive_dash/{}/' + '{}.csv.gz'.format(country)
```
jklukas, line 88:
```suggestion
        incline_prefix = 'incline/executive_dash/{}/' + '{}.csv.gz'.format(country)
```

Can we call this `incline_prefix_template` to make it more obvious that a `.format` is going to be called on it later? I also think we should make it a named placeholder rather than positional. Maybe call it `date`?

dags/incline_dash.py
```diff
@@ -59,49 +58,62 @@
...
+        data_copy = GoogleCloudStorageToGoogleCloudStorageOperator(
+            task_id='gcs_copy_latest_to_date',
+            source_bucket=gcs_bucket,
+            source_object=incline_prefix.format('latest'),
+            destination_bucket=gcs_bucket,
+            destination_object=incline_prefix.format('{{ds}}')
+        )
+
+        table_drop = BigQueryOperator(
+            task_id='drop_temp_table',
```
jklukas, line 109: We're creating these tasks in a loop, but giving them all static task_ids. I expect Airflow won't like this. It will also make the DAG surprisingly large. This has become complex enough that I think we'd be better off putting all of this logic into a single script using the google python SDKs, so the DAG is just two steps: the exec_dash query and then the script.

PR #939, opened 2020-04-07
Switch incline_dash DAG to use script/export_incline_dash

jklukas:
Closes https://github.com/mozilla/telemetry-airflow/pull/935

Depends on https://github.com/mozilla/bigquery-etl/pull/885

### DIFF
--- a/dags/incline_dash.py
+++ b/dags/incline_dash.py
@@ -3,11 +3,9 @@ from datetime import datetime, timedelta
 
 from utils.gcp import bigquery_etl_query
 
-from operators.backport.bigquery_operator_1_10_2 import BigQueryOperator
-
 from airflow.operators.sensors import ExternalTaskSensor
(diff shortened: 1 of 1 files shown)


dags/incline_dash.py
```diff
@@ -41,7 +37,7 @@
         dataset_id=dataset,
         # We recreate this entire table from scratch every day because we are
         # taking the last seen migration ping over all time for each client.
-        destination_table=None,
+        destination_table="migrated_clients_v1",
```
jklukas, line 40: Mistake here. Will fix.

fbertsch: This isn't necessary, [the query creates the table itself.](https://github.com/mozilla/bigquery-etl/blob/master/sql/org_mozilla_firefox_derived/migrated_clients_v1/init.sql#L1)

PR #940, opened 2020-04-07
Add generic glean ETL

jklukas:
This will immediately replace the VR browser tables, but there will be follow-up steps to point all downstream Fenix ETL at this source as well, after which we can remove many deprecated tables.

Depends on https://github.com/mozilla/bigquery-etl/pull/887 and https://github.com/mozilla/bigquery-etl/pull/877

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -9,6 +9,9 @@ from utils.gcp import (bigquery_etl_copy_deduplicate,
                        gke_command,
                        bigquery_xcom_query)
 
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from operators.gcp_container_operator import GKEPodOperator
+
 default_args = {
     "owner": "jklukas@mozilla.com",
     "start_date": datetime.datetime(2019, 7, 25),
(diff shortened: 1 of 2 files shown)


dags/copy_deduplicate.py
```diff
@@ -229,10 +232,39 @@
...
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
+        location="us-central1-a",
+        cluster_name="bq-load-gke-1",
+        namespace="default",
+        image="mozilla/bigquery-etl:latest",
+    )
+    baseline_args = [
+        "--project-id=moz-fx-data-shared-prod",
+        "--date={{ ds }}",
+        "--only=*_stable.baseline_v1"
```
acmiyaguchi, line 249: Is this how the `--only` option works? Is this the `only_tables` parameter or the `table_filter` parameter in `list_baseline_tables`?

https://github.com/mozilla/bigquery-etl/blob/aeccb5c5039a2671d9fe9a40adf69b44b22f73a4/bigquery_etl/glean_usage/common.py#L38-L49

jklukas: `--only` affects both what is passed as `only_tables` and what is passed as `table_filter`. The logic is fairly complex and you have to look at the `standard_args` implementation to see what all happens.

PR #942, opened 2020-04-08
Add sampling parameters to glam desktop.

emtwo:
Goes with https://github.com/mozilla/bigquery-etl/pull/881


emtwo:
Screen Shot 2020-04-16 at 1 03 18 PM

* Now using `sample_size` to space out the sampling so partitions are evenly sized.
* `clients_histogram_probe_counts` is its own table now - this is better overall but also helpful for processing it over several partitions

PR #948, opened 2020-04-14
Add public_data DAG

scholtzan:
This schedules publishing metadata to GCS for publicly available datasets.

### DIFF
--- a/dags/public_data.py
+++ b/dags/public_data.py
@@ -1,0 +1,46 @@
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+from datetime import timedelta, datetime
+from utils.gcp import gke_command
+
+default_args = {
+    "owner": "ascholtz@mozilla.com",
+    "email": ["ascholtz@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 4, 14),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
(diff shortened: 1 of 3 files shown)


dags/public_data.py
```diff
@@ -0,0 +1,46 @@
...
+    "email": ["ascholtz@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 4, 14),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("public_data", default_args=default_args, schedule_interval="0 6 * * *") as dag:
```
scholtzan, line 17: `deviations` schedule_interval is `"0 6 * * *"`

dags/public_data.py
```diff
@@ -0,0 +1,46 @@
...
+        task_id="wait_for_ssl_ratios",
+        external_dag_id="ssl_ratios",
+        external_task_id="ssl_ratios",
+        dag=dag,
+    )
+
+    public_data_gcs_metadata.set_upstream(
+        [
+            wait_for_public_analysis,
+            wait_for_ssl_ratios,
```
scholtzan, line 44: Those are the only two public datasets for now. I think once the lower-friction scheduled queries stuff is implemented, we don't need this task anymore. Instead it should be possible to determine upstream dependencies automatically.

PR #951, opened 2020-04-17
Dataproc Documentation && LTV Update

fbertsch:
Ben and I decided we wanted the partitioning per-day for this dataset. We will end up limiting the history of the dataset, but during this iteration phase we need to see over-time performance and comparisons.

Documentation for Dataproc should enable others to get up-and-running without consulting either of us.

### DIFF
--- a/bin/add_gcp_creds
+++ b/bin/add_gcp_creds
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -eou pipefail
+
+keyfile_path=${1:-""}
+if [ -z $keyfile_path ] ; then
+    echo "Must specify keyfile path" >&2;
+    exit 1;
+fi
+
+connection=${2:-"google_cloud_derived_datasets"}
+
+# Wait for full display until after checks
+set -x
+
+function format_gcp() {
+KEYFILE="$1" python3 - <<END
+import os
+import json
+
+with open(os.environ["KEYFILE"]) as f:
+    data = json.load(f)
+
+extra = {
+    "extra__google_cloud_platform__project": data["project_id"],
+    "extra__google_cloud_platform__keyfile_dict": json.dumps(data)
+}
+
+print(json.dumps(extra))
+END
+}
+
+function update_gcp() {
+    local conn_id=$1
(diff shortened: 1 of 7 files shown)


dags/utils/dataproc.py
```diff
@@ -663,7 +665,7 @@ def copy_artifacts_dev(dag, project_id, artifact_bucket, storage_bucket):
     return BashOperator(
         task_id="copy_to_dev_artifacts",
         bash_command="""
-        gcloud auth activate-service-account --key-file ~/.credentials
+        gcloud auth activate-service-account --key-file ~/.credentials || cat ~/.credentials
```
fbertsch, line 668: This now helpfully prints the contents of `.add_credentials.md` if you haven't set them correctly.

dags/utils/dataproc.py
```diff
@@ -166,6 +167,7 @@ def delete_cluster(self):
         """
         return DataprocClusterDeleteOperator(
             task_id='delete_dataproc_cluster',
+            trigger_rule=TriggerRule.ALL_DONE,
```
fbertsch, line 170: This will affect more than LTV, so checking in with you @acmiyaguchi to ensure it won't break anything that expects the cluster to stick around if the upstream job fails.

bin/add_gcp_creds
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -eou pipefail
+
+keyfile_path=${1:-""}
+if [ -z $keyfile_path ] ; then
+    echo "Must specify keyfile path" >&2;
+    exit 1;
+fi
```
acmiyaguchi, line 9:
```suggestion
keyfile_path=${1?"Must specify keyfile path"}
```

dags/utils/dataproc.py
acmiyaguchi: I don't think there are any jobs that rely on this behavior.

dags/utils/dataproc.py
acmiyaguchi: Good addition

PR #954, opened 2020-04-21
Reschedule vrbrowser_clients_daily

jklukas:
These were descheduled in #940 in favor of generic Glean baseline ETL, but
VR Browser folks have existing dashboards which rely on some features here that
don't exist in the baseline-only ETL. We will continue to support this for now
until we develop a more robust strategy for incorporating per-product metrics
into the generic ETL.


jklukas:
cc @daoshengmu.

PR #958, opened 2020-04-24
Bug 1632970 - Stop sending FFTV and Focus events to Amplitude

fbertsch:
This is part of the work to cover the immediate overages we are
seeing in Amplitude events.


fbertsch:
Just to clarify what this will result in:
- Stop sending data to Amplitude today for these two products
- No deleting of any existing data in Amplitude
- No deleting of any Telemetry data
- Possibility of resending these events at a later point in time, if we wish to resume Amplitude usage for said products

See also: https://github.com/mozilla/bigquery-etl/pull/929

PR #960, opened 2020-04-28
Fix writing to partition in ltv

fbertsch:


fbertsch:
@acmiyaguchi I tested this in my dev project and indeed there are two partitions in the tables.

PR #962, opened 2020-04-29
Ltv updates

fbertsch:
This makes two big changes:
1. Adds days_seen to predictions. We will now predict the number
   of days we will see this user over the next 28 days.
2. Adds p_alive for all metrics. We can use this to determine
   if we think the user is alive or dead. If p_alive for
   days_seen is <.5, we may assume the user is dead.

Small changes:
- Code fixup
- Cluster on sample id and client id
- Allow adding new fields to output table

@benmiroglio I've run this on the full set of rfm data from yesterday. You can see the results in `frank-sandbox.analysis.ltv_daily`. I've added you as a viewer to that dataset so you should have access.

The results for `p_alive` look interesting and merit further investigation:
image

The x-axis there is `p_alive` for `days_seen`, and the y-axis is cumulative fraction of users.

### DIFF
--- a/jobs/ltv_daily.py
+++ b/jobs/ltv_daily.py
@@ -12,6 +12,7 @@ from pyspark.sql.types import DoubleType
 from lifetimes import BetaGeoFitter
 
 PRED_METRICS = [
+    "days_seen",
 	"days_searched", 
 	"days_tagged_searched",
 	"days_clicked_ads",
@@ -126,6 +127,9 @@ def main(
     # flatten list
     columns = [item for sublist in columns for item in sublist]
 
+    prediction_prefix = "prediction_"
+    p_alive_prefix = "p_alive_"
+
     model_perf_data = pd.DataFrame()
     model_pred_data = None
     search_rfm_ds = search_rfm_full.limit(training_sample).select(columns).toPandas()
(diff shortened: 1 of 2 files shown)


jobs/ltv_daily.py
```diff
@@ -183,10 +209,22 @@ def ltv_predict_metric(metric, model=model):
...
     (
-        model_pred_data.write.format("bigquery")
+        model_pred_data
+
+        # Add prediction columns as maps
+        .withColumn("predictions", predictions)
+        .withColumn("p_alive", p_alive)
+
+        # Drop top-level prediction columns
+        .drop(*list(chain(*[[prediction_prefix + n, p_alive_prefix + n] for n in PRED_METRICS])))
```
fbertsch, line 219: Not super attached to this. We can keep the top-level columns.

fbertsch:
Looks like I can't get review from Ben. Anthony, he verified r+ for this change in-channel.

jobs/ltv_daily.py
```diff
@@ -12,6 +12,7 @@
 from lifetimes import BetaGeoFitter
 
 PRED_METRICS = [
+    "days_seen",
```
acmiyaguchi, line 15: The indentation looks off here, can you run the results through black?

PR #963, opened 2020-04-30
fixed improper path to pip-install.sh

crankycoder:
taar-weekly is failing because the pip-install.sh script is not being referenced correctly.

This ran properly in the dev enviroment because the path to the shell script did not have the `jobs/` prefix, but failed in production as I didn't put the prefix back in.


crankycoder:
After applying the fix and running from a local instance of airflow :
```
(py36) ➜  ~ aws s3 ls s3://telemetry-parquet/taar/ensemble/
2020-04-27 18:24:47        149 ensemble.weight.json
2020-04-30 12:19:13        142 ensemble_weight.json
```

PR #965, opened 2020-04-30
 Bug 1632591 - Reflect operational state of job on success by default 

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1632591)

This addresses a few concerns in the linked bug. One of the concerns is that the dashboard is not up to date. The reason this has been the case was that the `register_status` helper would by default only set jobs to partial outage, instead of always pinging statuspage. This could lead to some wonky situations in the case of backfill, but in retrospect this allowed the dashboard to become far out of date. The burden was always for someone to respond and to set the page to operational afterwards. Setting the option `on_success` to `True` should automate this.

This PR also removes some of the old checks that were run on datasets based on the number of partitions in s3.


acmiyaguchi:
Yes, it's also possible for a backfill to cause a partial outage if it fails, but the current day is up to date. We also don't take advantage of incidents on statuspage, which is probably one of the higher value things of the tool. Not everything can be done automatically with the statuspage, but hopefully this will prevent it from actively being misleading.

PR #968, opened 2020-05-04
Fix shredder dag timing

relud:
This was misconfigured because I misunderstood how `schedule_interval` impacts `ds` relative to actual run date.

This should run on Tuesday mornings UTC every 28 days, with `--end-date` set to the day before the run actually started, and with `--start-date` set to two `schedule_interval`s before that.

### DIFF
--- a/dags/shredder.py
+++ b/dags/shredder.py
@@ -5,7 +5,7 @@ from utils.gcp import gke_command
 default_args = {
     "owner": "dthorn@mozilla.com",
     "depends_on_past": True,
-    "start_date": datetime(2020, 3, 9),
+    "start_date": datetime(2020, 4, 7),
     "email": [
         "telemetry-alerts@mozilla.com",
         "dthorn@mozilla.com",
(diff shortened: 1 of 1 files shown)


dags/shredder.py
```diff
@@ -21,8 +21,11 @@
 base_command = [
     "script/shredder_delete",
     "--state-table=moz-fx-data-shredder.shredder_state.shredder_state",
-    "--start-date={{'2019-11-20' if ds == '2020-03-09' else macros.ds_add(ds, -7*4*2)}}",
-    "--end-date={{ds}}",
+    # start date should be two schedule intervals before end date, 27-28*2=-29
+    "--start-date={{macros.ds_add(ds, -29)}}",
+    # dags run schedule_interval after ds, and end date should be one day
+    # before the dag runs, so 27 days after ds
+    "--end-date={{macros.ds_add(ds, 27)}}",
```
acmiyaguchi, line 28: I see, this is because of how Airflow sets shifts the date a whole interval from the run date.

dags/shredder.py
```diff
@@ -21,13 +21,17 @@
...
 ]
 
 # main_v4 is cheaper to handle in a project without flat-rate query pricing
 on_demand = gke_command(
     task_id="on_demand",
+    name="shredder-on-demand",
```
relud, line 34: and prefix gke pod names with `shredder`, so that they are easier to find in the gcp console

dags/shredder.py
relud: yep.

PR #970, opened 2020-05-04
Add DAG for exporting FxA events to Amplitude

fbertsch:
Depends on https://github.com/mozilla/bigquery-etl/pull/941.

### DIFF
--- a/dags/fxa_amplitude_export.py
+++ b/dags/fxa_amplitude_export.py
@@ -11,7 +11,6 @@ default_args = {
     'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
     'email_on_failure': True,
     'email_on_retry': True,
-    'depends_on_past': True,
     'retries': 1,
     'retry_delay': datetime.timedelta(minutes=20),
 }
@@ -46,9 +45,10 @@ with models.DAG(
             dag_name=task_id,
             parent_dag_name=dag_name,
             default_args=default_args,
+            depends_on_past=True,
             project='moz-fx-data-shared-prod',
             dataset='firefox_accounts',
-            table_or_view='fxa_amplitude_export_v1',
+            table_or_view='fxa_amplitude_export',
             s3_prefix='fxa-active',
         ),
         task_id=task_id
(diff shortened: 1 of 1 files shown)


dags/fxa_amplitude_export.py
```diff
@@ -0,0 +1,46 @@
...
+    'depends_on_past': True,
+    'retries': 1,
+    'retry_delay': datetime.timedelta(minutes=20),
+}
+
+dag_name = 'fxa_amplitude_export'
+
+with models.DAG(
+        dag_name,
+        schedule_interval='0 10 * * *',
```
fbertsch, line 23: This follows the existing fxa job for events, but I'd like some info on where the ten-hour wait period came from.

jklukas: The 10 hour wait period was a very conservative guess since I don't know anything about guarantees on latency for the Stackdriver to BigQuery integration. The existing FxA ETL exists primarily to serve KPI calculations, so all we cared about was having the data ready early enough to have the KPI dashboard updated before 9:00 AM Pacific.

Running this at 1:00 UTC would _probably_ be fine but I have no hard data to back that up.

fbertsch: I checked in with jbuck and got some info about the expected latency. I've recorded it here for future reference. 30 minutes should give us plenty of leeway.

dags/fxa_amplitude_export.py
```diff
@@ -0,0 +1,57 @@
...
+from airflow.operators.subdag_operator import SubDagOperator
+from utils.amplitude import export_to_amplitude
+
+default_args = {
+    'owner': 'frank@mozilla.com',
+    'start_date': datetime.datetime(2020, 5, 1),
+    'email': ['telemetry-alerts@mozilla.com', 'frank@mozilla.com'],
+    'email_on_failure': True,
+    'email_on_retry': True,
+    'depends_on_past': True,
```
jklukas, line 14: I don't think we should set this at the dag level, but instead just mark `fxa_amplitude_export_v1` as depends_on_past

PR #977, opened 2020-05-07
Schedule anomdtct pipeline

scholtzan:
Schedules https://github.com/mozilla/forecasting/pull/27

Depends on https://github.com/mozilla/bigquery-etl/pull/961

### DIFF
--- a/dags/analysis.py
+++ b/dags/analysis.py
@@ -33,13 +33,6 @@ with DAG("public_analysis", default_args=default_args, schedule_interval="0 1 * 
         dag=dag,
     )
 
-    wait_for_clients_daily = ExternalTaskSensor(
-        task_id="wait_for_clients_daily",
-        external_dag_id="main_summary",
-        external_task_id="clients_daily",
-        dag=dag,
-    )
-
     wait_for_clients_first_seen = ExternalTaskSensor(
         task_id="wait_for_clients_first_seen",
         external_dag_id="main_summary",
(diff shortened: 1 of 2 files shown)


dags/analysis.py
```diff
@@ -17,10 +17,47 @@
...
+    # Built from https://github.com/mozilla/forecasting/tree/master/anomdtct
+    anomdtct_image = "gcr.io/moz-fx-data-forecasting/anomdtct:latest"
+
+    anomdtct = GKEPodOperator(
+        task_id="anomdtct",
+        name="anomdtct",
+        image=anomdtct_image,
+        email=["ascholtz@mozilla.com", "jmccrosky@mozilla.com",],
+        arguments=["{{ds}}"]
+        + ["--spreadsheet-id=" + Variable.get('anomdtct_spreadsheet_id')],
```
scholtzan, line 30: I'll add a variable in Airflow for this.

dags/analysis.py
```diff
@@ -17,10 +19,47 @@
...
+        task_id="wait_for_clients_first_seen",
+        external_dag_id="main_summary",
+        external_task_id="clients_first_seen",
+        dag=dag,
+    )
+
+    anomdtct.set_upstream(
+        [
+            wait_for_clients_daily,
+            wait_for_clients_first_seen,
```
jklukas, line 53: We could skip wait_for_clients_daily since clients_first_seen depends on it anyway.

dags/analysis.py
```diff
@@ -17,10 +19,47 @@
...
+        ]
+    )
+
     deviations = bigquery_etl_query(
         task_id="deviations",
         destination_table="deviations_v1",
         dataset_id="telemetry_derived",
+        arguments=("--replace",),
```
jklukas, line 61: I would have expected that we need to explicitly set project to shared-prod for the query. Not sure why this hasn't caused problems.

PR #981, opened 2020-05-11
Make FxA export PDT-based

fbertsch:


dags/fxa_amplitude_export.py
```diff
@@ -27,6 +32,11 @@
 From there the data is streamed to BigQuery. This is one record at-a-time
 and data should be available very quickly.
 Reference: https://cloud.google.com/logging/docs/export/bigquery
+
+However, the above exports to UTC-partitioned tables. We are running
+with PDT-based days, so the following will actually wait until
```
jklukas, line 37: The `fxa_amplitude_export_v1` has a `submission_date` parameter and populates data only for the prior UTC day, so I don't think we actually get any extra data by moving this start time back.

If the intention here is to export entire Pacific-based dates, then we'd need more extensive reworking of the ETL and I'm hesitant to introduce that here. @irrationalagent has requested Pacific-based ETL to ease comparing with Amplitude in the past and I have so far pushed back. I think we'd need to have a small proposal process for how to handle tables based on Pacific time and document the business case for it vs. the additional maintenance burden.

fbertsch: I understand this sentiments. However in the short-term I am merely looking to decrease our Amplitude spend. The current pipeline is quite literally incorrect for the data it would be pushing to Amplitude (that is, user it says are active on certain days may not be), and I'm hesitant to block that work on asking the FxA org to switch their Amplitude project and associated metrics to UTC.

My proposal is that we accept this short-term change for pacific-based times, and in the future work to see what would allow them to switch their analyses to UTC.

PR #982, opened 2020-05-11
Schedule fxa_amplitude_user_ids table

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/973

### DIFF
--- a/dags/fxa_export_to_amplitude.py
+++ b/dags/fxa_export_to_amplitude.py
@@ -67,6 +67,21 @@ with models.DAG(
 
     fxa_export_table_create >> fxa_amplitude_export
 
+    fxa_amplitude_user_ids = bigquery_etl_query(
+        task_id='fxa_amplitude_user_ids',
+        project_id='moz-fx-data-shared-prod',
+        destination_table='fxa_amplitude_user_ids_v1',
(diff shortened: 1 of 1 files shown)


dags/fxa_amplitude_export.py
```diff
@@ -67,6 +67,21 @@
...
 
+    fxa_amplitude_user_ids = bigquery_etl_query(
+        task_id='fxa_amplitude_user_ids',
+        project_id='moz-fx-data-shared-prod',
+        destination_table='fxa_amplitude_user_ids_v1',
+        dataset_id='firefox_accounts_derived',
+        depends_on_past=True,
+        start_date=datetime.datetime(2020, 5, 10),
+        email=['telemetry-alerts@mozilla.com', 'jklukas@mozilla.com'],
+        # This query updates the entire existing table every day rather than appending
```
fbertsch, line 78: This is becoming a pattern that is used in a few places. It might be time to set it as one obvious arg than a few non-obvious ones with explanatory comments. We can address that later if you'd like.

jklukas: Filed https://github.com/mozilla/telemetry-airflow/issues/985 to address later.

PR #986, opened 2020-05-12
Bug 1633638 - wait for event and main tables before exporting DevTools events to Amplitude

akkomar:
Devtools event view uses [event](https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/devtools_events_amplitude_v1/view.sql#L36) and [main](https://github.com/mozilla/bigquery-etl/blob/master/sql/telemetry_derived/devtools_events_amplitude_v1/view.sql#L85) tables. This makes the Amplitude export job wait for deduplication of these tables.

### DIFF
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -1,6 +1,7 @@
 import datetime
 
 from airflow import models
+from airflow.operators.sensors import ExternalTaskSensor
 from airflow.operators.subdag_operator import SubDagOperator
 from utils.amplitude import export_to_amplitude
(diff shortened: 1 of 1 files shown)


dags/bq_events_to_amplitude.py
```diff
@@ -83,3 +98,5 @@
         ),
         task_id=devtools_task_id
     )
+
+    [wait_for_telemetry_event, wait_for_telemetry_main] >> devtools_export
```
fbertsch, line 102: TIL about this syntax!

PR #989, opened 2020-05-12
Use correct connection id for job running in docker

acmiyaguchi:
This is causing job failures, causing aggregated fenix data to not be propagated into the nonprod bucket.


acmiyaguchi:
Yep, for context: https://github.com/mozilla/telemetry-airflow/pull/964

PR #991, opened 2020-05-13
Don't execute probe counts queries in parallel.

emtwo:
Refer to note in https://github.com/mozilla/bigquery-etl/pull/982

### DIFF
--- a/dags/glam.py
+++ b/dags/glam.py
@@ -282,6 +282,8 @@ client_scalar_probe_counts = bigquery_etl_query(
     dag=dag,
 )
 
+# SubdagOperator uses a SequentialExecutor by default
+# so its tasks will run sequentially.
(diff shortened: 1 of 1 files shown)


dags/glam.py
```diff
@@ -294,7 +294,6 @@
         None,
     ),
     task_id="clients_histogram_probe_counts",
-    executor=get_default_executor(),
```
acmiyaguchi, line 297: Should this be explicitly set to use `SequentialExecutor` e.g. `from airflow.executors.sequential_executor import SequentialExecutor`?

emtwo: `SubDagOperator` defaults to `SeqientialExecutor` so it does not need to be set explicitly. But I'll add a comment so it's clear

PR #997, opened 2020-05-20
Add job for hashing Adjust Ids

fbertsch:
Short-term this will re-hash all ids every day. I tested this in my sandbox and data is available at `frank-sandbox:firefox_preview_external.adjust_install_time_v1`

### DIFF
--- a/dags/adjust_import.py
+++ b/dags/adjust_import.py
@@ -1,0 +1,83 @@
+import json
+import os
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+from airflow.operators.subdag_operator import SubDagOperator
+from utils.dataproc import (
+    moz_dataproc_pyspark_runner,
+    copy_artifacts_dev,
+    get_dataproc_parameters,
+)
+
+
+EXPORT_TO_AVRO = True
+
+default_args = {
+    "owner": "frank@mozilla.com",
+    "depends_on_past": True,
+    "start_date": datetime(2020, 3, 15),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "frank@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
+}
+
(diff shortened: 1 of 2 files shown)


dags/adjust_import.py
```diff
@@ -0,0 +1,83 @@
...
+            "gs://dataproc-initialization-actions/python/pip-install.sh"
+        ],
+        additional_properties={
+            "spark:spark.jars": "gs://spark-lib/bigquery/spark-bigquery-latest.jar"
+        },
+        additional_metadata={"PIP_PACKAGES": "click==7.1.2"},
+        python_driver_code="gs://{}/jobs/adjust_import.py".format(params.artifact_bucket),
+        py_args=[
+            "--salt",
+            "org.mozilla.fenix-salt",
```
acmiyaguchi, line 61: The salt should probably be hidden in a variable.

jobs/adjust_import.py
```diff
@@ -0,0 +1,55 @@
...
+        .builder
+        .appName("adjust_gps_hash")
+        .getOrCreate())
+
+    @udf("string")
+    def pbkdf2_sha1hmac(msg, salt, iterations, klen):
+        import hashlib
+        import base64
+        return base64.b64encode(
+            hashlib.pbkdf2_hmac('sha1', str.encode(msg), str.encode(salt), iterations, klen)
```
acmiyaguchi, line 33: Why not `sha256` and 100k iterations, as per https://docs.python.org/3/library/hashlib.html#hashlib.pbkdf2_hmac?

fbertsch: This is actually to [match code in data sent from Fenix.](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsUtils.kt#L116)

dags/adjust_import.py
fbertsch: If you're worried about it being public, [it already is.](https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsUtils.kt#L116) Or do you mean that from a maintainability perspective?

PR #1001, opened 2020-05-26
Update parameters for webrender dashboard

acmiyaguchi:
As per @saptarshiguha, this should fix recent [airflow failures](https://workflow.telemetry.mozilla.org/log?task_id=webrender_ds_283&dag_id=webrender_ds_283&execution_date=2020-05-25T00%3A00%3A00%2B00%3A00).

### DIFF
--- a/dags/webrender.py
+++ b/dags/webrender.py
@@ -37,7 +37,8 @@ with DAG(
             # TODO: remove this after verifying permissions
             RUNFAST="true",
             BUCKET="gs://moz-fx-data-prod-analysis",
-            PROJECT_ID="moz-fx-data-bq-data-science",
+            PROJECT_ID="moz-fx-data-shared-prod",
+            DATASET="telemetry",
         ),
         dag=dag,
     )


dags/webrender.py
```diff
@@ -37,7 +37,8 @@
             # TODO: remove this after verifying permissions
             RUNFAST="true",
             BUCKET="gs://moz-fx-data-prod-analysis",
-            PROJECT_ID="moz-fx-data-bq-data-science",
+            PROJECT_ID="moz-fx-data-shared-prod",
+            DATASET="telemetry",
```
scholtzan, line 41: Shouldn't the table be written to `telemetry_derived` instead? We could then create a view in `telemetry` that references this table.

saptarshiguha:
The data sets is just for authentication. the command `bq_auth` in
https://github.com/mozilla/webrender_intel_win10_nightly/blob/master/query.R#L44
requires some data set.

All tables are written to locations here:
https://github.com/mozilla/webrender_intel_win10_nightly/blob/master/query.R#L13

PR #1002, opened 2020-05-27
Bug 1639373 - Remove options for validating permissions in webrender DAG

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1639373)

cc @saptarshiguha


acmiyaguchi:
Added a comment that the DATASET refers to the source dataset. The container is tested within a sandbox project, and those projects typically do not have permissions to write to production (e.g. moz-fx-data-shared-prod.telemetry) datasets.

PR #1007, opened 2020-06-02
Fix schedule_interval in public_data DAG

scholtzan:
I forgot to update the `schedule_interval` in the DAG for exporting public metadata after moving the anomdtc pipeline to a docker container.

### DIFF
--- a/dags/public_data.py
+++ b/dags/public_data.py
@@ -14,7 +14,7 @@ default_args = {
     "retry_delay": timedelta(minutes=30),
 }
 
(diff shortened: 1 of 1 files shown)


dags/public_data.py
```diff
@@ -35,7 +35,7 @@
         task_id="wait_for_ssl_ratios",
         external_dag_id="ssl_ratios",
         external_task_id="ssl_ratios",
-        execution_delta=timedelta(hours=6),
+        execution_delta=timedelta(hours=1),
```
scholtzan, line 38: The `schedule_interval` for `ssl_ratios` is `@daily`.

PR #1009, opened 2020-06-03
Factor AMO stats queries to their own DAG

jklukas:
This is the first DAG generated bigquery-etl that we're including here.
We're manually copying over the DAG for now, but eventually we will be
able to remove this definition from telemetry-airflow once WTMO automatically
imports bqetl-generated DAGs.


jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/1027

PR #1010, opened 2020-06-03
Add operator for running containers on GCE

acmiyaguchi:


acmiyaguchi:
It's easier to deal with a kubernetes cluster with a long life-span. Using a g1-small instance with an autoscaled node pool, it should be possible to have readily available compute with low start-up costs.

PR #1011, opened 2020-06-03
Use generated error_aggregates DAG

scholtzan:
This depends on https://github.com/mozilla/bigquery-etl/pull/1031

This is DAG is generated in bigquery-etl using the lower-friction scheduled queries machinery.
For now the DAG is copied manually, but eventually we will be
able to remove this definition from telemetry-airflow once WTMO automatically
imports bqetl-generated DAGs. 

The main difference here is that the error_aggregates table now has a version. So before merging, we'd need to create a `error_aggregates_v1` table in `telemetry_derived`, copy existing data over and create a `error_aggregates` view to make sure any queries that depend on `error_aggregates` don't break.

### DIFF
--- a/dags/error_aggregates.py
+++ b/dags/error_aggregates.py
@@ -1,29 +1,39 @@
-import datetime
+# Generated via query_scheduling/generate_airflow_dags
 
-from airflow import models
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+import datetime
 from utils.gcp import bigquery_etl_query
 
 default_args = {
-    'owner': 'bewu@mozilla.com',
-    'start_date': datetime.datetime(2019, 11, 1),
-    'email': ['telemetry-alerts@mozilla.com', 'bewu@mozilla.com', 'wlachance@mozilla.com'],
-    'email_on_failure': True,
-    'email_on_retry': True,
-    'depends_on_past': False,
-    'retries': 1,
-    'retry_delay': datetime.timedelta(minutes=20),
+    "owner": "bewu@mozilla.com",
+    "start_date": datetime.datetime(2019, 11, 1, 0, 0),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "bewu@mozilla.com",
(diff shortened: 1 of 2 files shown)


dags/error_aggregates.py
```diff
@@ -1,29 +1,39 @@
-import datetime
+# Generated via query_scheduling/generate_airflow_dags
```
jklukas, line 1: We should make sure to update the DAG generation so that this line specifies bigquery-etl. We might want to use the full GH URL: `Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py`

dags/error_aggregates.py
```diff
@@ -1,29 +1,39 @@
...
 
-from airflow import models
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
```
Ben-Wu, line 4: `ExternalTaskSensor` isn't used here.  Is this just an artifact of the dag generation?

dags/error_aggregates.py
```diff
@@ -1,29 +1,39 @@
...
 }
 
-dag_name = 'error_aggregates'
-
-with models.DAG(
-        dag_name,
-        schedule_interval=datetime.timedelta(hours=3),
-        default_args=default_args) as dag:
+with DAG(
+    "bqetl_error_aggregates",
```
Ben-Wu, line 24: Does changing the dag name remove the historical runs in the ui?  Doesn't matter in this case but worth noting.

jklukas: I would expect it does, yes. That's a good point to keep in mind as we transition, but in most cases I don't see any significant benefit to keeping it as long as recent runs have been successful.

dags/error_aggregates.py
scholtzan: It is. Currently, the imports are hard-coded into the DAG template. I guess as a future improvement it could make sense to keep track of whether the imports are actually used.

PR #1013, opened 2020-06-04
Use bigquery-etl generated deviations DAG

scholtzan:
Use the `bqetl_deviations` DAG which was generated in bigquery-etl. Depends on https://github.com/mozilla/bigquery-etl/pull/1034


dags/anomdtct.py
```diff
@@ -46,14 +46,3 @@
             wait_for_clients_first_seen,
         ]
     )
-
-    deviations = bigquery_etl_query(
-        task_id="deviations",
-        project_id="moz-fx-data-shared-prod",
-        destination_table="deviations_v1",
-        dataset_id="telemetry_derived",
-        arguments=("--replace",),
```
scholtzan, line 55: Just noticed this. We need to add support for these parameters.

jklukas: Does this line have any effect? The bigqueryrc in the image sets `--replace=true` as the default.

scholtzan: Oh right, then it doesn't make a difference here. There are some tasks that set other arguments though, so generally I think we want to support defining arguments.

PR #1016, opened 2020-06-05
bqetl_nondesktop DAG

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/1015

Generated in https://github.com/mozilla/bigquery-etl/pull/1041

### DIFF
--- a/dags/bqetl_nondesktop.py
+++ b/dags/bqetl_nondesktop.py
@@ -1,0 +1,128 @@
+# Generated via https://github.com/mozilla/bigquery-etl/blob/master/bigquery_etl/query_scheduling/generate_airflow_dags.py
+
+from airflow import DAG
+from airflow.operators.sensors import ExternalTaskSensor
+import datetime
+from utils.gcp import bigquery_etl_query
+
+default_args = {
+    "owner": "jklukas@mozilla.com",
+    "start_date": datetime.datetime(2019, 7, 25, 0, 0),
+    "email": ["telemetry-alerts@mozilla.com", "jklukas@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=300),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+}
+
+with DAG(
+    "bqetl_nondesktop", default_args=default_args, schedule_interval="0 1 * * *"
+) as dag:
+
+    telemetry_derived__firefox_nondesktop_day_2_7_activation__v1 = bigquery_etl_query(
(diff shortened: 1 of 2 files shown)


dags/copy_deduplicate.py
```diff
@@ -324,7 +261,15 @@
...
-    firefox_nondesktop_exact_mau28 >> simpleprophet_forecasts_mobile
+    wait_for_firefox_nondesktop_exact_mau28 = ExternalTaskSensor(
+        task_id="wait_for_firefox_nondesktop_exact_mau28",
+        external_dag_id="bqetl_nondesktop",
+        external_task_id="telemetry__firefox_nondesktop_exact_mau28_raw__v1",
+        check_existence=True,
+        dag=dag,
+    )
+
+    wait_for_firefox_nondesktop_exact_mau28 >> simpleprophet_forecasts_mobile
```
jklukas, line 272: I'm concerned about the fact that we now have the nondesktop DAG waiting on `copy_deduplicate`, and then  this later task in `copy_deduplicate` that depends on the nondesktop DAG.

I think the `simpleprophet_forecasts_mobile` task should rightfully move into the `kpi_dashboard` DAG, as the forecasts exist primarily for the KPI dash. 

But we're working toward generating that one via bq-etl as well, which would mean we can't put this task there.

But also the kpi dash DAG contains a smoot-related query that's only there because it had similar dependencies. Really, we should probably move all the smoot queries to a `bqetl_gud` DAG that's generated.

So.... that's a lot of potential for moving thing around. Out of scope for this PR, certainly. But I think the above steps would all be good ones to take in follow-ups. The KPI dash DAG should probably remain hand-written; all the forecast jobs (desktop, nondesktop, and fxa) would live there, and then it would perform just the one query for materializing the kpi_dashboard table.

scholtzan: Yeah moving things around and having a DAG for GUD related queries seems reasonable. I'll open a follow-up PR.

dags/bqetl_nondesktop.py
```diff
@@ -0,0 +1,128 @@
...
+        dataset_id="telemetry_derived",
+        project_id="moz-fx-data-shared-prod",
+        owner="jklukas@mozilla.com",
+        email=["jklukas@mozilla.com"],
+        date_partition_parameter="submission_date",
+        depends_on_past=False,
+        dag=dag,
+    )
+
+    telemetry__firefox_nondesktop_exact_mau28_raw__v1 = bigquery_etl_query(
```
scholtzan, line 71: This task is currently failing in Airflow. The table does not exist in `moz-fx-data-shared-prod` but actually is in `moz-fx-data-derived-datasets`

PR #1019, opened 2020-06-08
bqetl_gud DAG

scholtzan:
Generated via https://github.com/mozilla/bigquery-etl/pull/1046

### DIFF
--- a/dags/bqetl_gud.py
+++ b/dags/bqetl_gud.py
@@ -16,25 +16,11 @@ default_args = {
     "retries": 1,
 }
 
-with DAG(
-    "bqetl_nondesktop", default_args=default_args, schedule_interval="0 1 * * *"
-) as dag:
+with DAG("bqetl_gud", default_args=default_args, schedule_interval="0 1 * * *") as dag:
 
-    telemetry_derived__firefox_nondesktop_day_2_7_activation__v1 = bigquery_etl_query(
-        task_id="telemetry_derived__firefox_nondesktop_day_2_7_activation__v1",
-        destination_table="firefox_nondesktop_day_2_7_activation_v1",
-        dataset_id="telemetry_derived",
-        project_id="moz-fx-data-shared-prod",
-        owner="gkaberere@mozilla.com",
-        email=["gkaberere@mozilla.com", "jklukas@mozilla.com"],
-        date_partition_parameter="submission_date",
-        depends_on_past=False,
-        dag=dag,
-    )
-
-    firefox_nondesktop_exact_mau28_by_client_count_dimensions = bigquery_etl_query(
-        task_id="firefox_nondesktop_exact_mau28_by_client_count_dimensions",
-        destination_table="firefox_nondesktop_exact_mau28_by_client_count_dimensions_v1",
+    telemetry_derived__smoot_usage_new_profiles__v2 = bigquery_etl_query(
+        task_id="telemetry_derived__smoot_usage_new_profiles__v2",
+        destination_table="smoot_usage_new_profiles_v2",
         dataset_id="telemetry_derived",
         project_id="moz-fx-data-shared-prod",
         owner="jklukas@mozilla.com",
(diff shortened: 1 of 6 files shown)


dags/bqetl_gud.py
```diff
@@ -0,0 +1,145 @@
...
+    "start_date": datetime.datetime(2019, 7, 25, 0, 0),
+    "email": ["telemetry-alerts@mozilla.com", "jklukas@mozilla.com"],
+    "depends_on_past": False,
+    "retry_delay": datetime.timedelta(seconds=300),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+}
+
+with DAG("bqetl_gud", default_args=default_args, schedule_interval="0 1 * * *") as dag:
```
scholtzan, line 19: The schedule interval changes for new profile smoot queries here (compared to https://github.com/mozilla/telemetry-airflow/blob/6c7a407dec5b3f4a7e7044ebb8e06750c535789f/dags/kpi_dashboard.py#L23). I don't think it makes a difference though.

dags/kpi_dashboard.py
```diff
@@ -30,18 +33,23 @@
...
-        task_id='smoot_usage_new_profiles_compressed_v2',
-        project_id='moz-fx-data-shared-prod',
-        destination_table='smoot_usage_new_profiles_compressed_v2',
-        dataset_id='telemetry_derived',
+    wait_for_firefox_nondesktop_exact_mau28 = ExternalTaskSensor(
+        task_id="wait_for_firefox_nondesktop_exact_mau28",
+        external_dag_id="bqetl_nondesktop",
+        external_task_id="telemetry__firefox_nondesktop_exact_mau28_raw__v1",
+        check_existence=True,
+        execution_delta=timedelta(hours=14, minutes=45),
```
scholtzan, line 51: `bqetl_nondesktop` runs at `0 1 * * *`, so the `execution_delta` should be this.

dags/bqetl_gud.py
jklukas: Agreed. It's fine for this to run whenever. The later schedule was implicit dependency management, to make sure we wait for the fxa tables. Not relevant here.

dags/kpi_dashboard.py
jklukas: If we also add an ExternalTaskSensor for firefox_desktop_exact_mau28 and for the fxa forecast, then we no longer need to have the weird schedule for this DAG. It can also be defined to run at 01:00 UTC.

scholtzan: Added the fxa forecast and changed the scheduling interval

dags/kpi_dashboard.py
```diff
@@ -30,18 +33,65 @@
...
+        table_id='simpleprophet_forecasts_fxa_v1',
+        email=["telemetry-alerts@mozilla.com", "jklukas@mozilla.com"],
+    )
+
+    wait_for_firefox_accounts_exact_mau28_raw = ExternalTaskSensor(
+        task_id="wait_for_firefox_accounts_exact_mau28_raw",
+        external_dag_id="fxa_events",
+        external_task_id="firefox_accounts_exact_mau28_raw",
+        check_existence=True,
+        execution_delta=timedelta(hours=9),
```
scholtzan, line 92: The schedule interval for `fxa_events` is `* 10 * * *`

dags/kpi_dashboard.py
```diff
@@ -30,18 +33,65 @@
...
+        owner="jklukas@mozilla.com",
+        email=["telemetry-alerts@mozilla.com", "jklukas@mozilla.com"],
+        dag=dag)
+
+    wait_for_exact_mau_by_dimensions = ExternalTaskSensor(
+        task_id="wait_for_exact_mau_by_dimensions",
+        external_dag_id="main_summary",
+        external_task_id="exact_mau_by_dimensions",
+        check_existence=True,
+        mode="reschedule",
```
scholtzan, line 72: We might want to use this mode for `ExternalTaskSensor` in general. 
_"When set to reschedule the sensor task frees the worker slot when the criteria is not yet met and it’s rescheduled at a later time."_ (see https://airflow.apache.org/docs/stable/_api/airflow/sensors/base_sensor_operator/index.html)

PR #1032, opened 2020-06-12
Add DAG for mozfun

scholtzan:
This is for scheduling when public UDFs should be deployed. Only Airflow has permissions to deploy UDFs to `mozfun`.

### DIFF
--- a/dags/mozfun.py
+++ b/dags/mozfun.py
@@ -1,0 +1,24 @@
+from airflow import DAG
+from datetime import timedelta, datetime
+from utils.gcp import gke_command
+
+default_args = {
+    "owner": "ascholtz@mozilla.com",
+    "email": ["ascholtz@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 11),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("mozfun", default_args=default_args, schedule_interval="0 1 * * *") as dag:
+    docker_image = "mozilla/bigquery-etl:latest"
+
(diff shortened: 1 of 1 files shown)


dags/mozfun.py
```diff
@@ -0,0 +1,24 @@
...
+    "email": ["ascholtz@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 11),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("mozfun", default_args=default_args, schedule_interval="0 1 * * *") as dag:
```
scholtzan, line 16: The `schedule_interval` could also be more frequently, but maybe once a day is also alright for now.

jklukas: Once a day sounds good, and we can clear the most recent task in the UI manually to trigger a mid-day update.

dags/mozfun.py
```diff
@@ -0,0 +1,24 @@
...
+    docker_image = "mozilla/bigquery-etl:latest"
+
+    public_data_gcs_metadata = gke_command(
```
jklukas, line 19:
```suggestion
    publish_public_udfs = gke_command(
```

dags/mozfun.py
```diff
@@ -0,0 +1,24 @@
...
+        task_id="publish_public_udfs",
+        command=["script/publish_public_udfs"],
+        docker_image=docker_image,
+        dag=dag
```
jklukas, line 23:
```suggestion
        docker_image=docker_image
```

The `dag` keyword is redundant within a DAG context manager.

dags/mozfun.py
scholtzan: oops, copy-paste ;)

PR #1034, opened 2020-06-12
Bugfix: duplicate task IDs in fxa_events

jklukas:


dags/fxa_events.py
```diff
@@ -48,14 +48,14 @@
     )
 
     fxa_log_auth_events = bigquery_etl_query(
-        task_id='fxa_auth_events',
+        task_id='fxa_log_auth_events',
```
jklukas, line 51: I accidentally gave these the same names as the non-log variants.

PR #1038, opened 2020-06-15
Bug 1633918 Schedule Amplitude email click export

jklukas:
Depends on https://github.com/mozilla/bigquery-etl/pull/1070

### DIFF
--- a/dags/utils/amplitude.py
+++ b/dags/utils/amplitude.py
@@ -1,3 +1,5 @@
+from six.moves.urllib.request import urlopen
+
 from airflow import models
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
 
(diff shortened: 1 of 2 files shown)


jklukas:
> I mentioned in the other PR that we need a `submission_timestamp` field. The alternative is update the operator to use an arbitrary field for date, and set it to `time` for this job.

The view now includes a `submission_timestamp` field.

dags/utils/amplitude.py
```diff
@@ -66,6 +70,24 @@ def export_to_amplitude(
                 use_legacy_sql=False
         )
 
+        # If the underlying view requires special permissions only available to the
+        # Airflow workers, it may not be possible to maintain the view through the
+        # normal Jenkins-controlled view publishing process, so we provide a hook
+        # here to create the view with the same privileges that will be used when
+        # executing the query.
```
jklukas, line 77: @fbertsch - This is a bit of a weird compromise to handle permissions issues. I add here a `recreate_view` option that's turned on just for this email_click task. It downloads the query contents from a github url so that it can execute the `CREATE OR REPLACE VIEW` statement on the Airflow worker rather than in kube. I don't want to materialize this data to a long-lived table (which would let us sidestep this issue), as I want to keep the scope in which email clicks are available for analysis as limited as possible.

I'm definitely not in love with this solution, but I'd like to move forward with this so that I can begin testing the export in Amplitude stage. I'm open to suggestions on how better to approach this going forward.

fbertsch: I'm really not sure of any good alternatives offhand. We could try to add an option to bq_etl that runs the cli with a specific service account, and provide that to the container at runtime; however that would take some time to configure and test so let's stick with this.

PR #1039, opened 2020-06-15
Create DAG for burnham tests

hackebrot:
Hi folks! 👋 

This pull-request adds a new DAG for running Glean end-to-end tests using the Glean Python SDK. There are a few TODOs before we can merge this, but I'd like to get your feedback on what I have so far. I copied the GCP info from other DAGs. Remaining TODOs:

- [x] Push burnham Docker image to registry (https://github.com/mozilla/burnham/issues/26)
- [x] Update burnham Docker image project in DAG
- [x] Push Docker image to registry for verifying data (https://github.com/mozilla/burnham/issues/27)
- [x] Update GKEPodOperator for verifying data in DAG
- [x] Update burnham test information in DAG
- [x] Update SQL for BigQuerySQLSensorOperator in DAG

Please let me know your thoughts! Thanks! 🧑‍🚀 

Resolve https://github.com/mozilla/burnham/issues/28


dags/burnham.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+    # TODO: Update SQL for Sensor
+    # Wait for 10 pings with this test run ID in live tables
+    wait_for_data = BigQuerySQLSensorOperator(
+        task_id="wait_for_data",
+        sql="TODO",
+        bigquery_conn_id=DEFAULT_GCP_CONN_ID,
+        use_legacy_sql=False,
+    )
```
jklukas, line 137: We'll probably want to configure a `timeout` parameter here, so that the DAG fails in the case data never arrives rather than sitting in the waiting state forever.

dags/burnham.py
```diff
@@ -0,0 +1,147 @@
...
+DEFAULT_GKE_CLUSTER_NAME = "bq-load-gke-1"
+DEFAULT_GKE_NAMESPACE = "default"
+
+DAG_OWNER = "rpierzina@mozilla.com"
+DAG_EMAIL = ["telemetry-alerts@mozilla.com", "rpierzina@mozilla.com"]
+
+BURNHAM_PLATFORM_URL = "https://incoming.telemetry.mozilla.org"
+
+
+def new_burnham_operator(
```
jklukas, line 23: This should probably be factored out to a `dags/util/burnham.py` module to match style elsewhere. We generally keep any significant code outside of the DAG itself.

dags/burnham.py
```diff
@@ -0,0 +1,147 @@
...
+    # TODO: Update test information
+    burnham_test_run = "TEST RUN"
+    burnham_test_name = "TEST NAME"
+
+    client1 = new_burnham_operator(
+        task_id="client1",
+        burnham_test_run=burnham_test_run,
+        burnham_test_name=burnham_test_name,
+        burnham_missions=["MISSION G: FIVE WARPS, FOUR JUMPS", "MISSION C: ONE JUMP"],
+        burnham_spore_drive="tardigrade",
```
jklukas, line 104: Poor poor tardigrade...

dags/burnham.py
hackebrot: Done! ✅

dags/burnham.py
hackebrot: Added a timeout of 4 hours. Please let me now if that makes sense.

dags/burnham.py
```diff
@@ -0,0 +1,83 @@
...
+        email=DAG_EMAIL,
+    )
+
+    # TODO: Update SQL for Sensor
+    # Wait for 10 pings with this test run ID in live tables
+    wait_for_data = new_burnham_sensor(
+        task_id="wait_for_data", sql="TODO", timeout=60 * 60 * 4
+    )
+
+    # TODO: Run another GKEPodOperator to verify the data
```
fbertsch, line 75: This query could live in bq_etl, and use the bigquery_etl util to run.

dags/burnham.py
fbertsch: This will be querying the live data? We could probably push this down to an hour. Rarely do we have delay longer than that.

dags/burnham.py
fbertsch: Do you need help creating this query? If so, let us know.

dags/utils/burnham.py
```diff
@@ -0,0 +1,90 @@
...
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
+        location=gke_location,
+        cluster_name=gke_cluster_name,
+        namespace=gke_namespace,
+        image="gcr.io/<PROJECT>/burnham:latest",
+        image_pull_policy="Always",
+        env_vars=env_vars,
+        arguments=burnham_missions,
+        **kwargs
+    )
```
jklukas, line 70:
```suggestion
    return GKEPodOperator(
        task_id=task_id,
        gcp_conn_id=gcp_conn_id,
        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
        location=gke_location,
        cluster_name=gke_cluster_name,
        namespace=gke_namespace,
        image="gcr.io/{{ var.value.gcr_project_id }}/burnham:latest",
        image_pull_policy="Always",
        env_vars=env_vars,
        arguments=burnham_missions,
        **kwargs
    )
```

Based on new suggestions in https://bugzilla.mozilla.org/show_bug.cgi?id=1608958

dags/burnham.py
hackebrot: Yes, from what I understand we can query live data. I'll change this to one hour!

dags/utils/burnham.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def new_burnham_sensor(task_id, sql, gcp_conn_id=DEFAULT_GCP_CONN_ID, **kwargs):
```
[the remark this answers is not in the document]
hackebrot: I moved this here so it's consistent with the `new_burnham_operator` function and we don't have to copy or import the `DEFAULT_GCP_CONN_ID` to the DAG. Do you want me to move this back to the DAG?

dags/utils/burnham.py
```diff
@@ -0,0 +1,90 @@
...
+
+DEFAULT_GCP_CONN_ID = "google_cloud_derived_datasets"
+DEFAULT_GKE_LOCATION = "us-central1-a"
+DEFAULT_GKE_CLUSTER_NAME = "bq-load-gke-1"
+DEFAULT_GKE_NAMESPACE = "default"
+
+BURNHAM_PLATFORM_URL = "https://incoming.telemetry.mozilla.org"
+
+
+def new_burnham_operator(
```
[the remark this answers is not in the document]
hackebrot: Yes, that's a good idea. 👍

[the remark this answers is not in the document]
hackebrot: This function creates a new `GKEPodOperator` that runs the `burnham` Docker image. Do you have suggestions for a name that would better explain what it's doing?

dags/burnham.py
hackebrot: Yes, I will need help creating this query as well as the query for the sensor!

dags/burnham.py
```diff
@@ -0,0 +1,143 @@
...
+    "depends_on_past": False,
+    "retries": 0,
+}
+
+with models.DAG(
+    "burnham", schedule_interval="@daily", default_args=default_args,
+) as dag:
+
+    # Generate a UUID for this test run
+    burnham_test_run = str(uuid.uuid4())
```
jklukas, line 65: Generating a UUID here is unlikely to do what you might expect. See https://stackoverflow.com/questions/55748050/generating-uuid-and-use-it-across-airflow-dag

The top answer there has an example of using XCOM that happens to be exactly what we're here: calling `uuid.uuid4()`. I think we should follow exactly what they do there:

```suggestion
    generate_burnham_test_run_uuid = PythonOperator(
        task_id='generate_burnham_test_run_uuid',
        python_callable=lambda: str(uuid.uuid4())
    )
    burnham_test_run = '{{ task_instance.xcom_pull("generate_uuid") }}'
```

Note that the contents of `burnham_test_run` will literally be `'{{ task_instance.xcom_pull("generate_uuid") }}'`. The contents only get replaced when the individual operators run. Airflow operators can designate different parameters as "templated" meaning that the strings passed in will have variables replaced before running the body of the operator. It looks like all the relevant places this ID gets used _are_ templated params, so I think we're good.

jklukas: See for example https://airflow.apache.org/docs/stable/_api/airflow/contrib/operators/kubernetes_pod_operator/index.html#airflow.contrib.operators.kubernetes_pod_operator.KubernetesPodOperator for docs that show which params are templated.

dags/utils/burnham.py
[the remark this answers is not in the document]
jklukas: Perhaps this can be `burnham_run` and the query can be `burnham_sensor`.

dags/utils/burnham.py
```diff
@@ -0,0 +1,132 @@
...
+        env_vars["BURNHAM_SPORE_DRIVE"] = burnham_spore_drive
+
+    return GKEPodOperator(
+        task_id=task_id,
+        gcp_conn_id=gcp_conn_id,
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
+        location=gke_location,
+        cluster_name=gke_cluster_name,
+        namespace=gke_namespace,
+        image="gcr.io/{{ var.value.gcr_project_id }}/burnham:latest",
```
jklukas, line 63: It turns out `image` is _not_ templated, so we need to include the GCR project id verbosely here.
```suggestion
        image="gcr.io/moz-fx-data-airflow-prod-88e0//burnham:latest",
```

dags/utils/burnham.py
```diff
@@ -0,0 +1,132 @@
...
+    kwargs["name"] = kwargs.get("name", task_id.replace("_", "-"))
+
+    return GKEPodOperator(
+        task_id=task_id,
+        gcp_conn_id=gcp_conn_id,
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id,
+        location=gke_location,
+        cluster_name=gke_cluster_name,
+        namespace=gke_namespace,
+        image="gcr.io/{{ var.value.gcr_project_id }}/burnham-bigquery:latest",
```
jklukas, line 124:
```suggestion
        image="gcr.io/moz-fx-data-airflow-prod-88e0/burnham-bigquery:latest",
```

dags/utils/burnham.py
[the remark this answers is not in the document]
hackebrot: Done! ✅

dags/burnham.py
hackebrot: @jklukas helped me with creating the queries.

dags/burnham.py
hackebrot: Thank you! I updated the UUID generation as you suggested. Note that I used the task ID in the template.

hackebrot:
Thank you for your feedback and help with this pull-request @fbertsch @jklukas! 🚀

PR #1040, opened 2020-06-16
Change forecast tasks to use prod GCR images

jklukas:
Depends on https://github.com/mozilla/forecasting/pull/31

### DIFF
--- a/dags/utils/forecasting.py
+++ b/dags/utils/forecasting.py
@@ -12,7 +12,7 @@ def simpleprophet_forecast(
     gke_location="us-central1-a",
     gke_cluster_name="bq-load-gke-1",
     gke_namespace="default",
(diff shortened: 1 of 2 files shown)


dags/utils/forecasting.py
```diff
@@ -12,7 +12,7 @@ def simpleprophet_forecast(
     gke_location="us-central1-a",
     gke_cluster_name="bq-load-gke-1",
     gke_namespace="default",
-    docker_image="gcr.io/moz-fx-data-forecasting/simpleprophet:latest",
+    docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/simpleprophet:latest",
```
jklukas, line 15: I think we can't use `{{ var.value.gcr_project_id }}` here because it's in a module rather than directly in a jinja template.

PR #1041, opened 2020-06-16
Some fixes for Amplitude email clicks export

jklukas:


dags/utils/amplitude.py
```diff
@@ -80,13 +80,16 @@ def export_to_amplitude(
                 'https://raw.githubusercontent.com/mozilla/bigquery-etl/master/sql',
                 dataset, table_or_view, 'view.sql']))
             sql = response.read().decode('utf-8')
+            # The sql param is interpreted as a Jinja template, so we need to
+            # mark the content as raw to avoid interpretation.
+            sql = '{% raw %}\n' + sql + '\n{% endraw %}'
```
jklukas, line 85: I'm not 100% sure this will work, as the templating of the DAG itself might interpret these as tags in this line.

fbertsch: It may be because of this line: https://github.com/mozilla/telemetry-airflow/blob/master/dags/operators/bq_sensor.py#L44
If it is, then changing the name of the variable should work.

jklukas: It appears to be working in local docker-compose configuration.

jklukas: >  If it is, then changing the name of the variable should work.

What do you mean by "changing the name of the variable"? The parameter that the operator expects is `sql`. That's the name I have to pass in. The thing that's getting interpreted within the query text is I have a call to `FORMAT('{%t}', ...` And the `{%t}` is getting interpreted as a Jinja tag.

jklukas: Also, note that what I'm using here is `from operators.backport.bigquery_operator_1_10_2 import BigQueryOperator` rather than our custom BigQuerySQLSensorOperator.

BigQueryOperator defines `sql` as a templated parameter.

jklukas: I'm going to sidestep this by updating the query to avoid `{%`.

jklukas: https://github.com/mozilla/bigquery-etl/pull/1078

PR #1042, opened 2020-06-17
Use explicit project name for anomdtct image

jklukas:
Variable replacement only happens for operator parameters that are specified as templated. GKEPodOperator's image parameter is not templated.


jklukas:
Looking for retroactive review here on a simple fixup.

jklukas:
I was able to find in the logs in the derived-datasets project:

> Failed to apply default image tag "gcr.io/{{ var.value.gcr_project_id }}/anomdtct:latest": couldn't parse image reference "gcr.io/{{ var.value.gcr_project_id }}/anomdtct:latest": invalid reference format"

PR #1043, opened 2020-06-18
Revenue data writing

fbertsch:
This is going to actually be writing to a different project. We don't support that with the current bq_etl deploy procedures, so I'm writing it directly using the bigquery operator.

### DIFF
--- a/dags/ltv.py
+++ b/dags/ltv.py
@@ -1,18 +1,18 @@
 import json
 import os
-from datetime import datetime, timedelta
 
 from airflow import DAG
 from airflow.operators.sensors import ExternalTaskSensor
 from airflow.operators.subdag_operator import SubDagOperator
+from datetime import datetime, timedelta
+from operators.backport.bigquery_operator_1_10_2 import BigQueryOperator
+from six.moves.urllib.request import urlopen
 from utils.dataproc import (
     moz_dataproc_pyspark_runner,
     copy_artifacts_dev,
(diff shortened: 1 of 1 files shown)


dags/ltv.py
```diff
@@ -101,3 +101,17 @@
         dag=dag,
     )
     wait_for_search_clients_last_seen >> ltv_daily
+
+response = urlopen('/'.join([
+    'https://raw.githubusercontent.com/mozilla/bigquery-etl/master/sql',
+    'revenue_derived', 'client_ltv_v1', 'query.sql']))
```
acmiyaguchi, line 107: I think this will break the dag if it doesn't url doesn't exist (like it does now).

dags/ltv.py
```diff
@@ -101,3 +101,17 @@
...
+
+ltv_revenue_join=BigQueryOperator(
```
acmiyaguchi, line 109: Missing an import (though CI should pick up on this).

dags/ltv.py
```diff
@@ -101,3 +101,17 @@
         dag=dag,
     )
     wait_for_search_clients_last_seen >> ltv_daily
+
+response = urlopen('/'.join([
```
acmiyaguchi, line 105: Import is missing.

dags/ltv.py
acmiyaguchi: The import is also missing.

PR #1045, opened 2020-06-23
Add TAAR user profile BigTable ETL 

crankycoder:
This adds tasks to generate a TAAR user profile table in BigTable.

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -132,6 +132,7 @@ init_variables() {
     airflow variables -s "taar_bigtable_instance_id" "taar_bigtable_instance_id"
     airflow variables -s "taar_etl_storage_bucket" "taar_etl_storage_bucket"
     airflow variables -s "taar_gcp_project_id" "taar_gcp_project_id"
+    airflow variables -s "taar_dataflow_subnetwork" "taar_dataflow_subnetwork"
 }
 
 [ $# -lt 1 ] && usage
--- a/bin/run
+++ b/bin/run
@@ -128,6 +128,10 @@ init_variables() {
     airflow variables -s "surveygizmo_api_secret" "tapsekret"
     airflow variables -s "anomdtct_spreadsheet_id" "secret-id"
     airflow variables -s "anomdtct_spreadsheet_api_key" "secret-key"
+
+    airflow variables -s "taar_bigtable_instance_id" "taar_bigtable_instance_id"
+    airflow variables -s "taar_etl_storage_bucket" "taar_etl_storage_bucket"
+    airflow variables -s "taar_gcp_project_id" "taar_gcp_project_id"
 }
 
 [ $# -lt 1 ] && usage
--- a/dags/taar_weekly.py
+++ b/dags/taar_weekly.py
@@ -22,6 +22,7 @@ TAAR_ETL_STORAGE_BUCKET = Variable.get("taar_etl_storage_bucket")
 TAAR_PROFILE_PROJECT_ID = Variable.get("taar_gcp_project_id")
 TAAR_DATAFLOW_SUBNETWORK = Variable.get("taar_dataflow_subnetwork")
 
+# This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
 TAAR_ETL_CONTAINER_IMAGE = (
     "gcr.io/moz-fx-data-airflow-prod-88e0/taar_gcp_etl:0.4.7"
 )
--- a/dags/taar_weekly.py
+++ b/dags/taar_weekly.py
@@ -61,6 +61,19 @@ def wipe_gcs_files():
     ]
 
 
+def taar_profile_common_args():
+    return [
+        "-m",
+        "taar_etl.taar_profile_bigtable",
+        "--iso-date={{ ds_nodash }}",
+        "--gcp-project=%s" % TAAR_PROFILE_PROJECT_ID,
+        "--avro-gcs-bucket=%s" % TAAR_ETL_STORAGE_BUCKET,
+        "--bigtable-instance-id=%s" % TAAR_BIGTABLE_INSTANCE_ID,
+        "--sample-rate=1.0",
+        "--subnetwork=%s" % TAAR_DATAFLOW_SUBNETWORK,
+    ]
+
+
 wipe_gcs_bucket = GKEPodOperator(
     owner="vng@mozilla.com",
     email=["vng@mozilla.com", "mlopatka@mozilla.com",],
@@ -79,16 +92,7 @@ dump_bq_to_tmp_table = GKEPodOperator(
     task_id="dump_bq_to_tmp_table",
     name="dump_bq_to_tmp_table",
     image=TAAR_ETL_CONTAINER_IMAGE,
-    arguments=[
-        "-m",
-        "taar_etl.taar_profile_bigtable",
-        "--iso-date={{ ds_nodash }}",
-        "--gcp-project=%s" % TAAR_PROFILE_PROJECT_ID,
-        "--avro-gcs-bucket=%s" % TAAR_ETL_STORAGE_BUCKET,
-        "--bigtable-instance-id=%s" % TAAR_BIGTABLE_INSTANCE_ID,
-        "--sample-rate=1.0",
-        "--fill-bq",
-    ],
+    arguments=taar_profile_common_args() + ["--fill-bq",],
     dag=taar_weekly,
 )
 
@@ -98,15 +102,7 @@ extract_bq_tmp_to_gcs_avro = GKEPodOperator(
     task_id="extract_bq_tmp_to_gcs_avro",
     name="extract_bq_tmp_to_gcs_avro",
     image=TAAR_ETL_CONTAINER_IMAGE,
-    arguments=[
-        "-m",
-        "taar_etl.taar_profile_bigtable",
-        "--iso-date={{ ds_nodash }}",
-        "--gcp-project=%s" % TAAR_PROFILE_PROJECT_ID,
-        "--avro-gcs-bucket=%s" % TAAR_ETL_STORAGE_BUCKET,
-        "--bigtable-instance-id=%s" % TAAR_BIGTABLE_INSTANCE_ID,
-        "--bq-to-gcs",
-    ],
+    arguments=taar_profile_common_args() + ["--bq-to-gcs",],
     dag=taar_weekly,
 )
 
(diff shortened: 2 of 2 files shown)


dags/taar_weekly.py
```diff
@@ -30,50 +42,184 @@
...
+
+wipe_gcs_bucket = GKEPodOperator(
+    task_id="wipe_taar_gcs_bucket",
+    name="wipe_taar_gcs_bucket",
+    image="google/cloud-sdk:242.0.0-alpine",
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com"],
+    arguments=wipe_gcs_files(),
+    gcp_conn_id=taar_gcpdataproc_conn_id,
+    project_id=TAAR_PROFILE_PROJECT_ID,
```
haroldwoo, line 64: this wont work because the gke cluster doesnt exist in this project

crankycoder:
I've run this job to completion successfully on a local instance of Airflow using a 0.0001 sample rate off of `clients_last_seen`

Airflow_-_DAGs

I've added a `get_logs=False` settings in the `dataflow_import_avro_to_bigtable` GKEPodOperator to prevent airflow timing out while monitoring the pod when we are using a full 1.0 sample rate of users.

I tested this with the current values in w.t.m.o for the following Airflow variables as of 2020-June-24:

* taar_bigtable_instance_id
* taar_etl_storage_bucket
* taar_gcp_project_id

dags/taar_weekly.py
```diff
@@ -30,6 +46,128 @@
...
+    # See: https://github.com/mozilla/telemetry-airflow/issues/844
+    get_logs=False,
+    arguments=[
+        "-m",
+        "taar_etl.taar_profile_bigtable",
+        "--iso-date={{ ds_nodash }}",
+        "--gcp-project=%s" % TAAR_PROFILE_PROJECT_ID,
+        "--avro-gcs-bucket=%s" % TAAR_ETL_STORAGE_BUCKET,
+        "--bigtable-instance-id=%s" % TAAR_BIGTABLE_INSTANCE_ID,
+        "--subnetwork=regions/us-west1/subnetworks/gke-taar-nonprod-v1",
```
jasonthomas, line 132: Use variable `taar_dataflow_subnetwork` instead.

jasonthomas: `taar_dataflow_subnetwork` is an airflow variable.

bin/run
```diff
@@ -128,6 +128,10 @@ init_variables() {
     airflow variables -s "surveygizmo_api_secret" "tapsekret"
     airflow variables -s "anomdtct_spreadsheet_id" "secret-id"
     airflow variables -s "anomdtct_spreadsheet_api_key" "secret-key"
+
+    airflow variables -s "taar_bigtable_instance_id" "taar_bigtable_instance_id"
+    airflow variables -s "taar_etl_storage_bucket" "taar_etl_storage_bucket"
+    airflow variables -s "taar_gcp_project_id" "taar_gcp_project_id"
 }
```
haroldwoo, line 135: add taar_dataflow_subnetwork now that you've added it elsewhere

dags/taar_weekly.py
```diff
@@ -14,14 +17,28 @@
 taar_ensemble_cluster_name = "dataproc-taar-ensemble"
 taar_gcpdataproc_conn_id = "google_cloud_airflow_dataproc"
 
+TAAR_BIGTABLE_INSTANCE_ID = Variable.get("taar_bigtable_instance_id")
+TAAR_ETL_STORAGE_BUCKET = Variable.get("taar_etl_storage_bucket")
+TAAR_PROFILE_PROJECT_ID = Variable.get("taar_gcp_project_id")
+TAAR_DATAFLOW_SUBNETWORK = Variable.get("taar_dataflow_subnetwork")
+
+TAAR_ETL_CONTAINER_IMAGE = (
+    "gcr.io/moz-fx-data-airflow-prod-88e0/taar_gcp_etl:0.4.7"
```
haroldwoo, line 26: add ref comment on where the repo that builds this image is

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+
+dump_bq_to_tmp_table = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="dump_bq_to_tmp_table",
+    name="dump_bq_to_tmp_table",
+    # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
+    image=TAAR_ETL_CONTAINER_IMAGE,
+    arguments=[
+        "-m",
```
haroldwoo, line 83: just a nit that in the previous step you call a function to generate args and here you're inlining them

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+    cluster_name="bq-load-gke-1",
+    dag=taar_weekly,
+)
+
+dump_bq_to_tmp_table = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="dump_bq_to_tmp_table",
+    name="dump_bq_to_tmp_table",
+    # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
```
haroldwoo, line 80: move this comment to line 25?

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+    ],
+    dag=taar_weekly,
+)
+
+extract_bq_tmp_to_gcs_avro = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="extract_bq_tmp_to_gcs_avro",
+    name="extract_bq_tmp_to_gcs_avro",
+    # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
```
haroldwoo, line 100: same

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+    ],
+    dag=taar_weekly,
+)
+
+dataflow_import_avro_to_bigtable = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="dataflow_import_avro_to_bigtable",
+    name="dataflow_import_avro_to_bigtable",
+    # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
```
haroldwoo, line 119: same

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+    image=TAAR_ETL_CONTAINER_IMAGE,
+    # Due to the nature of the container run, we set get_logs to False,
+    # To avoid urllib3.exceptions.ProtocolError: 'Connection broken: IncompleteRead(0 bytes read)' errors
+    # Where the pod continues to run, but airflow loses its connection and sets the status to Failed
+    # See: https://github.com/mozilla/telemetry-airflow/issues/844
+    get_logs=False,
```
haroldwoo, line 125: nice im glad you found this obscure bug i had to workaround for probe scraper. Another option if possible is to have your container fork the dataflow job, write some sort of log every 15 seconds or so and poll until job is done. Though Im not familiar with the dataflow sdk.

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+)
+
+wipe_gcs_bucket_cleanup = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="wipe_gcs_bucket_cleanup",
+    name="wipe_taar_gcs_bucket",
+    location="us-central1-a",
+    cluster_name="bq-load-gke-1",
+    namespace="default",
```
haroldwoo, line 146: namespace line not needed

dags/taar_weekly.py
```diff
@@ -30,6 +47,128 @@
...
+    arguments=wipe_gcs_files(),
+    dag=taar_weekly,
+)
+
+wipe_bigquery_tmp_table = GKEPodOperator(
+    owner="vng@mozilla.com",
+    email=["vng@mozilla.com", "mlopatka@mozilla.com",],
+    task_id="wipe_bigquery_tmp_table",
+    name="wipe_bigquery_tmp_table",
+    # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
```
haroldwoo, line 157: same as before, rm in favor of commented where the image is defined

bin/run
haroldwoo: this will probably fix your cirlceci test environment check as well

dags/taar_weekly.py
haroldwoo: nvm i see that you reuse these args elsewhere

haroldwoo:
great work btw

dags/taar_weekly.py
```diff
@@ -14,14 +17,28 @@
...
+        "mlopatka@mozilla.com",
+        "telemetry-alerts@mozilla.com",
+    ],
     "depends_on_past": False,
     "start_date": datetime(2020, 4, 4),
-    "email": ["telemetry-alerts@mozilla.com"],
     "email_on_failure": True,
     "email_on_retry": True,
-    "retries": 2,
+    "retries": 0,
```
jasonthomas, line 41: We don't want to retry on failure?

dags/taar_weekly.py
crankycoder: Fixed in 97111e9b5

bin/run
crankycoder: Fixed in ec9155412944

dags/taar_weekly.py
crankycoder: Added in 3600e88e7a54

dags/taar_weekly.py
crankycoder: Fix in 6291b4a5ffd

dags/taar_weekly.py
crankycoder: I've extracted the common arguments into a function in 3f35557

dags/taar_weekly.py
crankycoder: I don't think there's a point to retry.  

I've come to mistrust the airflow emails as the email notifications come on every retry attempt.    I would rather have no retries so that every failure email is treated as something that should be important enough to look at.

crankycoder: @jasonthomas Maybe we should revisit the retry count after a couple weeks of having this job in production?

PR #1046, opened 2020-06-23
Run probe-scraper only on weekdays

fbertsch:
We've talked about this change before. @whd does this match your expectations, where it runs on Monday-Friday? Alternatively we could run Sunday-Thursday, but I'm afraid people expect schemas to be deployed on Fridays. Sunday deploys wouldn't be too bad since we could check in on them Monday morning.


fbertsch:
@Ben-Wu this won't cause any issues with the probe alerts, will it?

fbertsch:
Probe-scraper finishes at ~6:45 Pacific, which is indeed past everyone's EOD. With this in mind I think we should do Sunday - Thursday.

PR #1047, opened 2020-06-23
Move client queries out of main_summary

scholtzan:
Depends on https://github.com/mozilla/telemetry-airflow/pull/1036

Generated via https://github.com/mozilla/bigquery-etl/pull/1086 and https://github.com/mozilla/telemetry-airflow/pull/1047


dags/main_summary.py
```diff
@@ -259,18 +195,17 @@
     email=["telemetry-alerts@mozilla.com", "ssuh@mozilla.com"],
     dag=dag)
 
+wait_for_clients_daily = ExternalTaskSensor(
+    task_id="wait_for_clients_daily",
+    external_dag_id="bqetl_clients_daily",
+    external_task_id="telemetry_derived__clients_daily__v6",
+    dag=dag)
+
```
jklukas, line 203: Having an external task sensor here for clients_daily now feels very strange. I think within the scope of this PR, we should break out these two tasks that depend on clients_daily.

`experiments_daily_active_clients` should probably be in its own DAG generated by bigquery-etl. 

And then the clients_daily export should probably also be its own DAG. Or perhaps we should call the DAG main_summary_export, and include the main_summary_export task there, too.

scholtzan: https://github.com/mozilla/bigquery-etl/pull/1094 moves queries out of `main_summary`. I named the export DAG `parquet_export` for now, but I'm open to alternative names too.

dags/copy_deduplicate.py
```diff
@@ -40,6 +40,20 @@
         # copy_deduplicate job in another DAG.
         except_tables=["telemetry_live.main_v4"])
 
+    # We copy yesterday's main pings from telemetry_live to telemetry_stable
+    # at the root of this DAG because telemetry_stable.main_v4 will become
+    # the source for main_summary, etc. once we are comfortable retiring parquet
+    # data imports.
```
jklukas, line 46:
```suggestion
```

This comment is very outdated now.

dags/parquet_export.py
```diff
@@ -34,44 +35,7 @@
...
-    destination_table="main_summary_v4",
-    project_id="moz-fx-data-shared-prod",
-    dataset_id="telemetry_derived",
-    sql_file_path="sql/telemetry_derived/main_summary_v4/",
-    multipart=True,
-    owner="relud@mozilla.com",
-    email=["telemetry-alerts@mozilla.com", "relud@mozilla.com"],
-    start_date=datetime(2019, 10, 25),
-    dag=dag)
+dag = DAG('parquet_export', default_args=default_args, schedule_interval='0 1 * * *', max_active_runs=10)
```
jklukas, line 38:
```suggestion
dag = DAG('parquet_export', default_args=default_args, schedule_interval='0 1 * * *')
```

I assume max_active_runs was an optimization for main_summary. We should definitely get @haroldwoo's opinion on all of this before merging. He has implemented some main_summary-specific configuration due to the criticality of that workload; I believe we have a separate task queue or slot allotment for main_summary.

Now that we're splitting this ETL to multiple DAGs, we may want to remove that main_summary-specific configuration. Or perhaps extend it to the bqetl_main_summary and bqetl_clients_daily DAGs.

Or, there may be an argument for keeping the whole main_summary through clients_daily and exact_mau tasks in one DAG due to the ability to reserve task capacity for that pipeline.

scholtzan:
The generated DAGs use 'mode' = 'reschedule' by default for ExternalTaskSensors. So there are a few already running. One issue that sometimes alerts get sent when a task gets re-scheduled, like:
```
Try 0 out of 3
Exception:
Executor reports task instance finished (success) although the task says its queued. Was the task killed externally?
Log: Link
Host: ...
Log file: /app/logs/.../.../00:00.log
Mark success: Link
```

I found some open issues related to the problem: https://issues.apache.org/jira/browse/AIRFLOW-5071 and possibly https://issues.apache.org/jira/browse/AIRFLOW-5074
Other than increasing the number of slots or core parallelism those don't provide a good solution though.

haroldwoo:
ah thats unfortunate. the next time it happens can you open a bug with the details? Any idea what use cases trigger this since you mentioned it only happens sometimes?

scholtzan:
I opened a bug linking to a task that caused the mentioned alert a couple of days ago: https://bugzilla.mozilla.org/show_bug.cgi?id=1648594

dags/copy_deduplicate.py
jklukas: Can we remove this comment?

PR #1048, opened 2020-06-23
Schedule forwarding deletion requests to amplitude

relud:
requires https://github.com/mozilla/bigquery-etl/pull/1082


dags/bq_events_to_amplitude.py
```diff
@@ -98,5 +114,20 @@
...
+            "--secret-key={{ var.value.amplitude_secret_key }}",
+            "--table-id=moz-fx-data-shared-prod.telemetry.deletion_request_v4",
+            "--user-id-field=client_id",
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        dag=dag,
+    )
 
     [wait_for_telemetry_event, wait_for_telemetry_main] >> devtools_export
+    wait_for_telemetry_event >> [fenix_shredder, devtools_shredder]
```
fbertsch, line 133: It's not clear to me why we need this dependency.

relud: because we need to wait for copy_deduplicate before reading from stable tables

dags/bq_events_to_amplitude.py
```diff
@@ -99,4 +115,20 @@
...
+            "--secret-key={{ var.value.amplitude_secret_key }}",
+            "--table-id=moz-fx-data-shared-prod.telemetry.deletion_request_v4",
+            "--user-id-field=client_id",
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        dag=dag,
+    )
+
+    [wait_for_copy_deduplicate_all, wait_for_copy_deduplicate_main_ping] >> devtools_export
+    wait_for_copy_deduplicate_all >> [fenix_shredder, devtools_shredder]
```
relud, line 134: @fbertsch does this make it more clear?

fbertsch: It does! Thank you for clarifying.

PR #1049, opened 2020-06-25
Add task for play store export

Ben-Wu:
This will probably need updates to the cluster permissions

### DIFF
--- a/dags/play_store_export.py
+++ b/dags/play_store_export.py
@@ -1,0 +1,37 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from utils.gcp import gke_command
+
+
+default_args = {
+    "owner": "bewu@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 23),
+    "email_on_failure": True,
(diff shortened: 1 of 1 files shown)


dags/play_store_export.py
```diff
@@ -0,0 +1,37 @@
...
+        task_id="play_store_export",
+        command=[
+            "python", "play_store_export/export.py",
+            "--date={{ ds }}",
+            "--backfill-day-count=35",
+            "--project", project_id,
+            "--transfer-config={{ var.value.play_store_transfer_config_id }}",
+        ],
+        docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/play-store-export:latest",
+        gcp_conn_id="google_cloud_derived_datasets",
```
fbertsch, line 32: Will these permissions work with the marketing project?

Ben-Wu: From what I understand, the service account here just needs permissions to create pods so it should work.  Permissions to the marketing project then need to be given to the service account used by the cluster.  I want to see this fail though so I can verify this.

PR #1054, opened 2020-06-26
Move util/burnham directly into burnham DAG

jklukas:
This is a workaround to deal with skew between the libraries that the Airflow UI node sees and what's on master.


jklukas:
This goes back to the structure you originally had, @hackebrot, where all the utility functions are defined directly in the DAG. It's less clean, but it avoids the skew in code between the DAG and the separate util module.

PR #1057, opened 2020-06-29
Use submission_date parameter as partition for LTV

fbertsch:


dags/ltv.py
```diff
@@ -106,13 +106,16 @@
     'https://raw.githubusercontent.com/mozilla/bigquery-etl/master/sql',
     'revenue_derived', 'client_ltv_v1', 'query.sql']))
 
+BigQueryOperator.template_fields += ('query_params',)
```
fbertsch, line 109: I tested this locally on a sample query, but not this one directly.

PR #1059, opened 2020-06-29
Remove exports into bespoke project

acmiyaguchi:
The permissions for this have not been resolved. Any services that want to fetch data from the `glam-fenix-dev` project can be granted access from the sandbox project instead.


robhudson:
What's the longer term plan for all Glean products?

acmiyaguchi:
Once a production project is set up, the dataset/datasets will live in that project, and the resources won't have to move across projects. It's been almost a month since I've filed the bugs for modifying the permissions so I can copy into the bucket, but there doesn't seem to be any urgency, and the job failures are just noise at this point.

https://bugzilla.mozilla.org/show_bug.cgi?id=1637794

PR #1060, opened 2020-06-30
Temporarily comment out broken shredder tasks

fbertsch:


fbertsch:
I have a hunch that this is causing some sort of issue with the deploy of the new onboarding task.

PR #1068, opened 2020-07-06
Add tasks for per process daily histogram aggregates

Ben-Wu:
https://github.com/mozilla/bigquery-etl/pull/1124

I made a new `generate_and_run_query` that wraps the `gke_command` logic


dags/glam.py
```diff
@@ -183,54 +132,42 @@
...
-    gcp_conn_id="google_cloud_derived_datasets",
+clients_daily_histogram_aggregates_content = generate_and_run_query(
+    task_id="clients_daily_histogram_aggregates_content",
+    project_id=project_id,
+    source_dataset_id=dataset_id,
+    sample_size=PERCENT_RELEASE_WINDOWS_SAMPLING,
+    overwrite=False,
+    probe_type="histogram",
+    process="content",
+    get_logs=False,
```
emtwo, line 158: Do we still need this now that the query lengths are smaller?

dags/glam.py
```diff
@@ -366,8 +303,10 @@
 clients_scalar_aggregates >> scalar_percentiles
 clients_scalar_aggregates >> client_scalar_probe_counts
 
-latest_versions >> clients_daily_histogram_aggregates
-clients_daily_histogram_aggregates >> clients_daily_keyed_histogram_aggregates
+latest_versions >> clients_daily_histogram_aggregates_parent
+clients_daily_histogram_aggregates_parent >> clients_daily_histogram_aggregates_content
+clients_daily_histogram_aggregates_parent >> clients_daily_keyed_histogram_aggregates
+clients_daily_histogram_aggregates_content >> clients_histogram_aggregates
```
emtwo, line 309: I think we need `clients_daily_keyed_histogram_aggregates >> clients_histogram_aggregates` too here.

Ben-Wu: It's here right below the diff

dags/glam.py
Ben-Wu: It's possible to remove it but it's cutting really close.  If the query was about a minute longer it would fail so I would keep it

Ben-Wu: Actually I just tested and both the content and keyed queries just failed after 7 minutes but the parent query succeeded after 8 minutes so definitely should keep get_logs=False

dags/glam.py
emtwo: lol oops missed that

PR #1072, opened 2020-07-08
Forward deletion requests to amplitude

relud:
requires https://github.com/mozilla/bigquery-etl/pull/1142

### DIFF
--- a/dags/bq_events_to_amplitude.py
+++ b/dags/bq_events_to_amplitude.py
@@ -37,20 +37,20 @@ with models.DAG(
         task_id=fenix_task_id
     )
 
-#    fenix_shredder = gke_command(
-#        task_id="fenix_amplitude_shredder",
-#        name="shredder-amplitude-fenix",
-#        command=[
-#            "script/shredder_amplitude",
-#            "--date={{ ds }}",
-#            "--api-key={{ var.value.amplitude_api_key }}",
-#            "--secret-key={{ var.value.amplitude_secret_key }}",
-#            "--table-id=moz-fx-data-shared-prod.org_mozilla_fenix.deletion_request_v1",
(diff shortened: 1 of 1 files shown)


dags/fxa_events.py
```diff
@@ -60,6 +60,20 @@
...
     )
 
+    shredder_fxa = gke_command(
+        task_id="shredder_amplitude_fxa",
+        name="shredder-amplitude-fxa",
+        command=[
+            "script/shredder_amplitude",
+            "--api-key={{ var.value.fxa_amplitude_api_key }}",
+            "--secret-key={{ var.value.fxa_amplitude_secret_key }}",
+            "--table-id=moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_{{ ds_nodash }}",
```
relud, line 70: @jklukas can you confirm this table id looks correct?

jklukas: Yes, this looks correct except that the `jsonPayload.fields.uid` field in this table contains the raw FxA UID. Amplitude receives events with an HMAC-hashed version of the FxA UID, so we need to hash the field before sending (see [an example where we do this in a query](https://github.com/mozilla/bigquery-etl/blob/473be4aef1fab68dcc0ea9fbb6a3154562fa1602/sql/firefox_accounts_derived/fxa_amplitude_user_ids_v1/init.sql#L4-L19)). We may need to add a new scheduled query to produce a table with the hashed IDs for deletion requests, and use that as the source here rather than the prod logs directly.

relud: > We may need to add a new scheduled query

Yeah, this is going to be needed in a couple places, too. I will break fxa out into a separate PR

PR #1075, opened 2020-07-08
Reschedule DAGs

scholtzan:
Generated via https://github.com/mozilla/bigquery-etl/pull/1143


dags/bqetl_activity_stream.py
```diff
@@ -36,6 +36,7 @@
         task_id="wait_for_copy_deduplicate_copy_deduplicate_all",
         external_dag_id="copy_deduplicate",
         external_task_id="copy_deduplicate_all",
+        execution_delta=datetime.timedelta(days=-1, seconds=82800),
```
scholtzan, line 39: Ah, missed those... fix incoming

dags/incline_dash.py
```diff
@@ -32,6 +32,7 @@
         task_id="wait_for_core_clients_last_seen",
         external_dag_id="bqetl_core",
         external_task_id="telemetry_derived__core_clients_last_seen__v1",
+        execution_delta=timedelta(hours=-1),
```
jklukas, line 35: I'd recommend that all these DAGs for which we're setting `hours=-1` dependencies, we should instead change when the DAG runs. We should make these also start at 02:00 UTC, or perhaps even later at 03:00 UTC and set `hours=1`.

PR #1077, opened 2020-07-09
Bug 1639196 - schedule user_activity and annotations exports for Public Data Report

akkomar:


dags/firefox_public_data_report.py
```diff
@@ -82,4 +83,60 @@
     )
 )
 
-wait_for_main_ping >> hardware_report
\ No newline at end of file
+wait_for_clients_last_seen = ExternalTaskSensor(
+    task_id="wait_for_clients_last_seen",
+    external_dag_id="main_summary",
+    external_task_id="clients_last_seen",
+    execution_delta=timedelta(days=-6),
```
fbertsch, line 90: Offhand, this is surprising

akkomar: I'm always confused about this, hence the comment on line 40.

fbertsch: Got it, that makes sense now

PR #1079, opened 2020-07-09
Reschedule DAGs and set priority and pools

scholtzan:
Generated via https://github.com/mozilla/bigquery-etl/pull/1146

### DIFF
--- a/dags/bqetl_amo_stats.py
+++ b/dags/bqetl_amo_stats.py
@@ -101,7 +101,6 @@ with DAG(
         external_dag_id="bqetl_clients_daily",
         external_task_id="telemetry_derived__clients_daily__v6",
         execution_delta=datetime.timedelta(seconds=7200),
-        pool="sensor_pool",
         check_existence=True,
         mode="reschedule",
     )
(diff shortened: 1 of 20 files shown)


dags/bqetl_amo_stats.py
```diff
@@ -100,7 +100,8 @@
         task_id="wait_for_telemetry_derived__clients_daily__v6",
         external_dag_id="bqetl_clients_daily",
         external_task_id="telemetry_derived__clients_daily__v6",
-        execution_delta=datetime.timedelta(seconds=3600),
+        execution_delta=datetime.timedelta(seconds=7200),
+        pool="sensor_pool",
```
scholtzan, line 104: Should we consult data ops before creating this pool?

jklukas: Do we even have permissions to create a pool? Yes, I think we should consult with data ops before creating a pool. And probably shouldn't do so at all while hwoo is out. I think I was assuming that if we used a pool name that didn't exist, it would default to the existing pool. So perhaps best to take out the sensor_pool change for now.

scholtzan: Apparently assigning tasks to non-existing pools might crash the scheduler: https://issues.apache.org/jira/browse/AIRFLOW-1157
Better remove it for now.

scholtzan:
🤞

PR #1081, opened 2020-07-09
Use correct user id field in shredder_amplitude_fxa

relud:
I changed it in bigquery-etl before I merged, but I forgot to fix it here.


relud:
> Do you mind pointing out where this can be found in bigquery-etl?

https://github.com/mozilla/bigquery-etl/pull/1140/files#diff-af0a25761d73c3b334fe00e70ddcb3eeR10

PR #1087, opened 2020-07-11
Fix main_summary copy_dedupe dependency

scholtzan:


jklukas:
Looks like we didn't deadlock this morning! ` bqetl_public_data_json` looks to be the only DAG with issues. `ltv_daily` also had a failure,  but is likely due to a change unrelated to this rescheduling.

PR #1092, opened 2020-07-13
Replace fxa_events with generated DAG

jklukas:
Generated via https://github.com/mozilla/bigquery-etl/pull/1163

The one non-query task in the DAG was sending delete events to Amplitude,
so that task is migrated to the bq_events_to_amplitude DAG where deletions
for other Amplitude projects also happen.


dags/bq_events_to_amplitude.py
```diff
@@ -150,3 +150,29 @@
         ),
         task_id=onboarding_task_id
     )
+
+    shredder_amplitude_fxa = gke_command(
```
jklukas, line 154: cc @relud that I'm moving this fxa amplitude shredder task into the bq_events_to_amplitude DAG.

relud: lgtm

PR #1094, opened 2020-07-14
Allow field addition for specified dates in bigquery_etl_query

relud:
blocks https://github.com/mozilla/bigquery-etl/pull/1170

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -457,6 +457,7 @@ def bigquery_etl_query(
     docker_image="mozilla/bigquery-etl:latest",
     date_partition_parameter="submission_date",
     multipart=False,
+    allow_field_addition_on_date=None,
     **kwargs
 ):
     """ Generate.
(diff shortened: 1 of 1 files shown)


scholtzan:
It's already set for some queries: https://github.com/mozilla/bigquery-etl/blob/f267e38f14769d78ffc04e483904da378b483a1c/sql/firefox_accounts_derived/fxa_auth_bounce_events_v1/metadata.yaml#L11

jklukas:
> It's already set for some queries via arguments

And the fxa_events tables are a bit of a special case where we're doing a `SELECT *` from a table in another project that we have no control over, so we always want to be able to automatically pull in new fields.

I can see the extra safety of specifying certain dates for allowing field addition, but I do worry it will be a bit of a pain when making a PR. If review takes a few days, it will be necessary to update the configured dates before merging. It's definitely a better situation than requiring manual updates to the schema via the BQ console, which is error prone.

dags/utils/gcp.py
```diff
@@ -500,6 +503,10 @@ def bigquery_etl_query(
         + ["--dataset_id=" + dataset_id]
         + (["--project_id=" + project_id] if project_id else [])
         + ["--parameter=" + parameter for parameter in parameters]
+        + [
+            "--schema_update_option={{ 'ALLOW_FIELD_ADDITION' if ds in %r else '' }}"
+            % (allow_field_addition_dates,)
+        ]
```
jklukas, line 509: Can we wrap this in `([ ...] if allow_field_addition_dates else [])`?

relud: yeah

PR #1099, opened 2020-07-21
Add fxa_basket_export DAG

jklukas:
See https://github.com/mozmeao/basket/issues/426

Relies on permissions in
https://github.com/mozilla-services/cloudops-infra/pull/2326

### DIFF
--- a/dags/fxa_basket_export.py
+++ b/dags/fxa_basket_export.py
@@ -1,0 +1,88 @@
+from datetime import datetime, timedelta
+
+from airflow import models
+
+# Backported 1.10.2 bqoperator
+from operators.backport.bigquery_operator_1_10_2 import BigQueryOperator
+
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
+from airflow.contrib.operators.bigquery_to_gcs import BigQueryToCloudStorageOperator
+from airflow.contrib.operators.gcs_delete_operator import GoogleCloudStorageDeleteOperator
+from airflow.contrib.operators.gcs_to_s3 import GoogleCloudStorageToS3Operator
+from operators.bq_sensor import BigQuerySQLSensorOperator
+
+
+default_args = {
+    "owner": "jklukas@mozilla.com",
+    "start_date": datetime(2020, 7, 21),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "jklukas@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 3,
+    "retry_delay": timedelta(minutes=30),
+}
+
+GCS_BUCKET = "fxa-prod-basket"
+GCS_PREFIX = "last-active-timestamp"
+
+with models.DAG("adjust_import", default_args=default_args, schedule_interval="15 1 * * *"):
+    # Create the table with yesterday's data
+    gcp_conn_id = 'google_cloud_derived_datasets'
+    project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
+    dataset = 'temp'
+    temp_table_name = 'fxa_basket_export_{{ ds_nodash }}'
+    fully_qualified_table_name = '{}.{}.{}'.format(project_id, dataset, temp_table_name)
+
+    sql = """
+    SELECT
+      jsonPayload.fields.uid,
+      UNIX_SECONDS(MIN(timestamp)) AS timestamp
+    FROM
+      `moz-fx-fxa-prod-0712.fxa_prod_logs.docker_fxa_auth_{{ ds_nodash }}`
+    WHERE
+      jsonPayload.type = 'activityEvent'
+      AND jsonPayload.fields.event = 'account.signed'
+    GROUP BY
+      1
+    ORDER BY
+      2 ASC;
+    """
+
+    create_table = BigQueryOperator(
+        task_id='create_temporary_table',
+        sql=sql,
+        destination_dataset_table=fully_qualified_table_name,
+        bigquery_conn_id=gcp_conn_id,
+        use_legacy_sql=False,
+        write_disposition='WRITE_TRUNCATE',
+    )
+
+    directory = '/'.join((GCS_PREFIX, '{{ ds_nodash }}'))
+    extension = '.csv'
+
+    # Export from bq to gcs
+    # Docs: https://github.com/apache/airflow/blob/master/airflow/contrib/operators/bigquery_to_gcs.py#L28 # noqa: E501
+    gcs_uri = 'gs://{}/{}/*{}'.format(GCS_BUCKET, directory, extension)
+    table_extract = BigQueryToCloudStorageOperator(
+        task_id='bq_to_gcs',
+        source_project_dataset_table=fully_qualified_table_name,
+        destination_cloud_storage_uris=[gcs_uri],
+        bigquery_conn_id=gcp_conn_id,
+        compression='GZIP',
+        export_format='CSV',
+        field_delimiter=',',
+        print_header=False,
+    )
+
+    # Drop the temporary table
+    table_drop = BigQueryOperator(
+        task_id='drop_temp_table',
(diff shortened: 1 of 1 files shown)


dags/fxa_basket_export.py
```diff
@@ -0,0 +1,85 @@
...
+    gcp_conn_id = 'google_cloud_derived_datasets'
+    project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
+    dataset = 'temp'
+    temp_table_name = 'fxa_basket_export_{{ ds_nodash }}'
+    fully_qualified_table_name = '{}.{}.{}'.format(project_id, dataset, temp_table_name)
+
+    sql = """
+    SELECT
+      jsonPayload.fields.uid,
+      UNIX_SECONDS(MIN(timestamp)) AS timestamp
```
jklukas, line 39: Wrapped in UNIX_SECONDS, so this outputs rows like:

```
abcdd43402d746d0a7136acd04ddabcd,1590451200
```

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,85 @@
...
+        use_legacy_sql=False,
+        write_disposition='WRITE_TRUNCATE',
+    )
+
+    directory = '/'.join((GCS_PREFIX, '{{ ds_nodash }}'))
+    extension = '.csv'
+
+    # Export from bq to gcs
+    # Docs: https://github.com/apache/airflow/blob/master/airflow/contrib/operators/bigquery_to_gcs.py#L28 # noqa: E501
+    gcs_uri = 'gs://{}/{}/*{}'.format(GCS_BUCKET, directory, extension)
```
jklukas, line 65: Will create files like:

```
gs://fxa-prod-basket/last-active-timestamp/20200720/000000000000001.csv
gs://fxa-prod-basket/last-active-timestamp/20200720/000000000000002.csv
```

Definitely open to suggestions on specifics on what these paths should look like.

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,85 @@
...
+    table_extract = BigQueryToCloudStorageOperator(
+        task_id='bq_to_gcs',
+        source_project_dataset_table=fully_qualified_table_name,
+        destination_cloud_storage_uris=[gcs_uri],
+        bigquery_conn_id=gcp_conn_id,
+        compression='GZIP',
```
jklukas, line 71: Do we want to GZIP these?

jklukas:
cc @jbuck

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,89 @@
...
+        use_legacy_sql=False,
+        write_disposition="WRITE_TRUNCATE",
+    )
+
+    directory = "/".join((GCS_PREFIX, "{{ ds_nodash }}"))
+    extension = ".csv"
+
+    # Export from bq to gcs
+    # Docs: https://github.com/apache/airflow/blob/master/airflow/contrib/operators/bigquery_to_gcs.py#L28 # noqa: E501
+    gcs_uri = "gs://{}/{}/*{}".format(GCS_BUCKET, directory, extension)
```
jklukas, line 69: Will create files like:

gs://fxa-prod-basket/last-active-timestamp/20200720/000000000000001.csv
gs://fxa-prod-basket/last-active-timestamp/20200720/000000000000002.csv

Definitely open to suggestions on specifics on what these paths should look like.

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,89 @@
...
+    table_extract = BigQueryToCloudStorageOperator(
+        task_id="bq_to_gcs",
+        source_project_dataset_table=fully_qualified_table_name,
+        destination_cloud_storage_uris=[gcs_uri],
+        bigquery_conn_id=gcp_conn_id,
+        compression="GZIP",
```
jklukas, line 75: Do we want to GZIP these files?

dags/fxa_basket_export.py
jbuck: Yeah, this looks good to me

dags/fxa_basket_export.py
jbuck: Yeah, lets have them gzipped

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,89 @@
...
+GCS_BUCKET = "fxa-prod-basket"
+GCS_PREFIX = "last-active-timestamp"
+
+with models.DAG(
+    "adjust_import", default_args=default_args, schedule_interval="15 1 * * *"
+):
+    # Create the table with yesterday's data
+    gcp_conn_id = "google_cloud_derived_datasets"
+    project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
+    dataset = "temp"
```
fbertsch, line 36:
```suggestion
    dataset = "tmp"
```

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,89 @@
...
+}
+
+GCS_BUCKET = "fxa-prod-basket"
+GCS_PREFIX = "last-active-timestamp"
+
+with models.DAG(
+    "adjust_import", default_args=default_args, schedule_interval="15 1 * * *"
+):
+    # Create the table with yesterday's data
+    gcp_conn_id = "google_cloud_derived_datasets"
```
fbertsch, line 34: Any reason not to use shared-prod?

dags/fxa_basket_export.py
```diff
@@ -0,0 +1,89 @@
...
+        write_disposition="WRITE_TRUNCATE",
+    )
+
+    directory = "/".join((GCS_PREFIX, "{{ ds_nodash }}"))
+    extension = ".csv"
+
+    # Export from bq to gcs
+    # Docs: https://github.com/apache/airflow/blob/master/airflow/contrib/operators/bigquery_to_gcs.py#L28 # noqa: E501
+    gcs_uri = "gs://{}/{}/*{}".format(GCS_BUCKET, directory, extension)
+    table_extract = BigQueryToCloudStorageOperator(
```
fbertsch, line 70: I recommend adding a delete data step before this. The Amplitude pipeline deletes after we export to S3, but this one should delete it before we try exporting the day of data.

dags/fxa_basket_export.py
jklukas: Will change. I think I only used this because the GKE cluster is in derived-datasets, but that's not relevant to this DAG.

dags/fxa_basket_export.py
jklukas: This DAG is responsible for exporting the results of a query to GCS. It runs a query to create a table, extracts the contents of that table to gzipped CSVs in GCS, then drops the table.

I don't see how we could delete the temporary table before letting the extract job complete.

dags/fxa_basket_export.py
jklukas: Reverted this change back to `google_cloud_derived_datasets`. That is the name of the relevant connection in Airflow. There is no connection named "shared_prod".

dags/fxa_basket_export.py
fbertsch: Sorry, meant the files. If this overwrites the destination that works too; I can't remember offhand if that is the case.

dags/fxa_basket_export.py
fbertsch: Right, of course. Can we write to shared prod with that connection ID?

dags/fxa_basket_export.py
jklukas: Yes, the default (and non-configurable) behavior is to overwrite.

I can see there would be some safety in making sure the destination directory is empty before writing, though. I will add that.

dags/fxa_basket_export.py
jklukas: > Right, of course. Can we write to shared prod with that connection ID?

Yes.

dags/fxa_basket_export.py
jklukas: Added GCS delete

jklukas:
Ready for @fbertsch review again.

PR #1100, opened 2020-07-22
Bug 1654038 Tolerate compact hist in update_orphaning_dashboard job

jklukas:


jobs/update_orphaning_dashboard_etl.py
```diff
@@ -104,15 +104,26 @@ def parse_args():
 AGGREGATION_TABLE_NAME = "out_of_date_longitudinal_shim"
 
 def longitudinal_shim_aggregate(date_from, date_to, destination_project, destination_dataset, destination_table):
-    """Aggregate selected metrics to destination table.
-
-       Until full GCP backfill is completed, we have to union `telemetry.main` table with
-       `moz-fx-data-shared-prod.static.main_1pct_backfill` in the query. The latter contains a sample of main pings
-       having sample_id=42 from '2019-01-01' to '2019-09-01'."""
+    """Aggregate selected metrics to destination table."""
```
jklukas, line 107: It looks like this no longer uses `main_1pct_backfill`. I believe this table still exists, though, and we can likely delete it now.

PR #1104, opened 2020-07-23
Undo ExternalTaskSensor pools

scholtzan:
Reverts https://github.com/mozilla/telemetry-airflow/pull/1103, if something goes wrong


jklukas:
Looks like #1003 was fine.

PR #1105, opened 2020-07-23
Add DAG for monitoring tasks

scholtzan:
For now this DAG only schedules one task for determining partition sizes but I guess there will be more monitoring related tasks that can be added here in the future.

### DIFF
--- a/dags/monitoring.py
+++ b/dags/monitoring.py
@@ -48,5 +48,5 @@ with DAG("monitoring", default_args=default_args, schedule_interval="0 2 * * *")
         owner="ascholtz@mozilla.com",
         email=["telemetry-alerts@mozilla.com", "ascholtz@mozilla.com"])
 
-    stable_table_sizes.set_upstream(wait_for_copy_deduplicate_all)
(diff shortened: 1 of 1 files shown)


dags/monitoring.py
```diff
@@ -0,0 +1,52 @@
...
+from airflow.operators.sensors import ExternalTaskSensor
+from datetime import timedelta, datetime
+from operators.gcp_container_operator import GKEPodOperator
+from utils.gcp import gke_command
+
+default_args = {
+    "owner": "ascholtz@mozilla.com",
+    "email": ["ascholtz@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2020, 7, 1),
```
scholtzan, line 11: I'm not sure if we also want to backfill some of this data, then we can set it to something earlier. Or later

jklukas: I do think it could be useful to backfill through the full history we have for main_v4, back to 2018-10-30.

scholtzan: 👍  Updated the timestamp

scholtzan:
I guess this is ready to go?

PR #1106, opened 2020-07-23
Ensure repeated_subdag partition size is an integer

Ben-Wu:
`python3`‼️ 🙆‍♂️

### DIFF
--- a/dags/glam_subdags/general.py
+++ b/dags/glam_subdags/general.py
@@ -5,8 +5,8 @@ from utils.gcp import bigquery_etl_query
 
 def merge_params(min_param, max_param, additional_params):
     parameters = (
-        "min_sample_id:INT64:{}".format(min_param),
-        "max_sample_id:INT64:{}".format(max_param),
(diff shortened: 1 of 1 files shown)


dags/glam_subdags/general.py
```diff
@@ -34,7 +34,12 @@ def repeated_subdag(
     # This task runs first and replaces the relevant partition, followed
     # by the next tasks that append to the same partition of the same table.
     NUM_SAMPLE_IDS = 100
-    PARTITION_SIZE = NUM_SAMPLE_IDS / num_partitions
+    PARTITION_SIZE = NUM_SAMPLE_IDS // num_partitions
+
+    if NUM_SAMPLE_IDS % num_partitions != 0:
+        raise ValueError(f"Number of partitions must be a divisor "
+                         f"of the number of sample ids ({NUM_SAMPLE_IDS})")
```
acmiyaguchi, line 41: A wild format string!

Ben-Wu:
That file also has both `.format()` and `%`.  I changed those to f strings for consistency

PR #1108, opened 2020-07-24
Add dag for app store analytics export

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1576788

Corresponding scripts: https://github.com/mozilla/app-store-analytics-export

### DIFF
--- a/dags/app_store_analytics.py
+++ b/dags/app_store_analytics.py
@@ -1,0 +1,67 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from utils.gcp import gke_command
+
+
+default_args = {
+    "owner": "bewu@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 23),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
+    "email": [
+        "bewu@mozilla.com",
+    ],
+}
+
+PROJECT_ID = "moz-fx-data-marketing-prod"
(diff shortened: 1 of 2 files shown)


dags/app_store_analytics.py
```diff
@@ -0,0 +1,67 @@
...
+            "--password={{ var.value.app_store_connect_password }}",
+            f"--app-id={app_id}",
+            f"--app-name={app_name}",
+            "--start-date={{ yesterday_ds }}",  # previous day data is incomplete
+            f"--project={PROJECT_ID}",
+            f"--dataset={DATASET_ID}",
+        ]
+
+        if i == 0:
+            commands.append("--overwrite")
```
wlach, line 54: What's the purpose of this conditional? Can you add a comment?

Ben-Wu: This clears the current day's partition in the destination tables to ensure the only data for the day is the data written by this dag run.  This means if the table has any apps that aren't recognized here they'll be deleted.

dags/app_store_analytics.py
```diff
@@ -0,0 +1,69 @@
...
+
+        app_store_analytics = gke_command(
+            task_id=f"app_store_analytics_{app_name}",
+            command=commands,
+            docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/app-store-analytics-export:latest",
+            gcp_conn_id="google_cloud_derived_datasets",
+            dag=dag,
+        )
+
+        if i != 0:
```
wlach, line 66:
```suggestion
        if i > 0:
```

PR #1110, opened 2020-07-28
fixup! Bug 1654038 Tolerate compact histograms in update_orphaning_da…

akkomar:
…shboard job (#1100)


jobs/update_orphaning_dashboard_etl.py
```diff
@@ -108,7 +108,7 @@ def longitudinal_shim_aggregate(date_from, date_to, destination_project, destina
 
     bq = bigquery.Client()
 
-    longitudinal_shim_sql = f"""
```
akkomar, line 111: F-string was failing here on line 121:
```
        FORMAT('{"values":{%s}}', ARRAY_TO_STRING(ARRAY(
```

PR #1111, opened 2020-07-28
Set app store export to two day lag 

Ben-Wu:
Available data latency seems to be inconsistent and previous day data is available but incomplete for many metrics.  Setting to two day lag should work but I'll check over the next week to see if three days is needed.

Also I'll make a change to the export to warn about incomplete data.


Ben-Wu:
looks like there's a circleci outage.  I'll rerun builds when I can

PR #1115, opened 2020-07-29
Delete shredder pods regardless of exit status

relud:


haroldwoo:
heh sorry, meant True

PR #1116, opened 2020-07-30
Add tasks for app store derived tables

Ben-Wu:
To create derived tables combining metrics per dimension

Related bq-etl pr: https://github.com/mozilla/bigquery-etl/pull/1207


Ben-Wu:
It would need to be a separate dag that depends on this one because of the `gke_command`.  It's also a bit of a special case with the date parameter being `execution_date - 2 days`.  I'd like to do this way instead.

PR #1121, opened 2020-08-05
Add org_mozilla_fenix_nightly and comment on sources

acmiyaguchi:
A small change to get historical data flowing through the (deprecated?) org_mozilla_fenix_nightly app id, and a link to the spreadsheet with all of the app names.

### DIFF
--- a/dags/glam_fenix.py
+++ b/dags/glam_fenix.py
@@ -9,23 +9,28 @@ default_args = {
     "owner": "amiyaguchi@mozilla.com",
     "depends_on_past": False,
     "start_date": datetime(2020, 2, 19),
-    "email": ["telemetry-alerts@mozilla.com", "amiyaguchi@mozilla.com", "bewu@mozilla.com"],
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "amiyaguchi@mozilla.com",
+        "bewu@mozilla.com",
+    ],
     "email_on_failure": True,
     "email_on_retry": True,
     "retries": 1,
     "retry_delay": timedelta(minutes=30),
 }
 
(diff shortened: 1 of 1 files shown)


dags/glam_fenix.py
```diff
@@ -9,23 +9,28 @@
...
 PRODUCTS = [
-    "org_mozilla_fenix",
-    "org_mozilla_firefox",
-    "org_mozilla_firefox_beta",
-    "org_mozilla_fennec_aurora",
+    "org_mozilla_fenix",  # 2019-07-03 - present (nightly)
+    "org_mozilla_fenix_nightly",  # 2019-06-30 - 2020-07-03
+    "org_mozilla_firefox",  # 2020-07-28 - present
+    "org_mozilla_firefox_beta",  # 2020-03-26 - present
+    "org_mozilla_fennec_aurora", # 2020-01-21 - present
```
Ben-Wu, line 30: What do the dates mean?  When they started sending glean pings?  Shouldn't fennec_aurora be in the same situation as fenix_nightly since they're both no longer listed on the app store and messaging was sent out to switch to the other apps.  In any case, it might be useful to note that fennec_aurora is the deprecated nightly.

Ben-Wu: Oh wait nvm I see the dates

acmiyaguchi: The dates are the first appearance in the app store, which I assume is the earliest date where we'd be collecting data from an actual user population. As for fennec aurora, I misread the column for name end date being the same as app store end date, so it's correct that fenix nightly and fennec aurora are in the same boat.

PR #1127, opened 2020-08-13
Add leanplum messages tasks

Ben-Wu:
https://github.com/mozilla/leanplum-data-export/pull/19


Ben-Wu:
DAG runs every 4 hours so this replaces the partition in the message tables multiple times a day but imo this is fine since data size is < 1 MB

dags/leanplum_export.py
```diff
@@ -17,47 +17,62 @@
...
 
-with DAG("leanplum_export",
-         default_args=default_args,
-         schedule_interval=timedelta(hours=4)) as dag:
+with DAG(
+        "leanplum_export",
+        default_args=default_args,
+        # last run of a day will be at 03:59 of the next day
+        schedule_interval="59 3/4 * * *",
+        max_active_runs=1,
```
fbertsch, line 53: Why limit active runs?

dags/leanplum_export.py
```diff
@@ -17,47 +17,62 @@
...
-    firefox_android_nightly_export = leanplum.export(
-        task_id="firefox_android_nightly_export",
-        bq_project="moz-fx-data-shared-prod",
-        s3_prefix="firefox_android_nightly",
-        bq_dataset_id="firefox_android_nightly_external",
-        table_prefix="leanplum",
-        version="2",
-        dag=dag
-    )
+        message_export.set_upstream(data_export)
```
fbertsch, line 78: Can you explain why these are dependent?

dags/leanplum_export.py
Ben-Wu: Oh right, this has to do with the exports and not the messages.  It's because the next run may start before the previous one ends if there are a bunch of sequential airflow `Pod took to long to start` errors or for some other reason.  If that happens the second run will try to process files that the first job has in queue because a file isn't marked as done until it's processed.  This makes it so each file is only processed once

dags/leanplum_export.py
Ben-Wu: Great question!  They are not

PR #1128, opened 2020-08-18
Schedule experiment_search_aggregates queries

scholtzan:
Schedules https://github.com/mozilla/bigquery-etl/pull/1233

Very similar to scheduling `experiment_enrollment_aggregates`

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -136,11 +136,11 @@ with models.DAG(
     experiment_search_query_task_id = "experiment_search_aggregates_live_generate_query"
 
     # setting xcom_push to True outputs this query to an xcom
-    experiment_search_aggregates_live_generate_query = gke_command(
+    experiment_search_aggregates_live_generate_view = gke_command(
         task_id=experiment_search_query_task_id,
         command=[
             "python",
-            "sql/telemetry_derived/experiment_search_aggregates_live/view.sql.py",
+            "sql/telemetry_derived/experiment_search_aggregates_live_v1/view.sql.py",
             "--submission-date",
             "{{ ds }}",
             "--json-output",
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -123,6 +123,48 @@
...
+        email=["telemetry-alerts@mozilla.com", "ascholtz@mozilla.com"])
+
+    experiment_search_query_task_id = "experiment_search_aggregates_live_generate_query"
+
+    # setting xcom_push to True outputs this query to an xcom
+    experiment_search_aggregates_live_generate_query = gke_command(
+        task_id=experiment_search_query_task_id,
+        command=[
+            "python",
+            "sql/telemetry_derived/experiment_search_aggregates_live/view.sql.py",
```
akkomar, line 143:
```suggestion
            "sql/telemetry_derived/experiment_search_aggregates_live_v1/view.sql.py",
```
typo?

dags/copy_deduplicate.py
```diff
@@ -123,6 +123,48 @@
...
+            "--json-output",
+            "--wait-seconds",
+            "15",
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        xcom_push=True,
+        owner="ascholtz@mozilla.com",
+        email=["telemetry-alerts@mozilla.com", "ascholtz@mozilla.com"])
+
+    experiment_enrollment_aggregates_live_run_query = bigquery_xcom_query(
```
akkomar, line 155:
```suggestion
    experiment_search_aggregates_live_run_query = bigquery_xcom_query(
```
Seems like this should be called `_search_`, instead of `_enrollment_`?

Also nit: it's not clear to me that this task creates a view, just by looking at its name.

akkomar (changes requested):

dags/copy_deduplicate.py
scholtzan: I updated the name. `..._deploy_view` is probably more accurate for this step

akkomar (approved):

PR #1130, opened 2020-08-25
Add new parameters to burnham-bigquery operator

hackebrot:
Depends on https://github.com/mozilla/burnham/pull/62


jklukas:
I reran the DAG for today, and looks like it works! Results are in BQ. It looks like the emojis fail to display properly, which probably points to a string encoding issue at some point in the code.

hackebrot:
Woohoo! Thank you @jklukas! 🥳

PR #1133, opened 2020-08-26
Deduplicate burnham documents

hackebrot:
Hi folks! 👋 

I updated the SQL queries in the burnham DAG to deduplicate documents submitted through automation. As discussed on Slack there is no guarantee live tables don't have duplicates, so we need this extra step to ensure reproducibility.

We don't need to deduplicate rows for the distinct client IDs test, I think. I ran an example for each of the updated queries manually and they seem to produce the expected values.

Resolve https://github.com/mozilla/burnham/issues/61

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -17,20 +17,22 @@ from operators.gcp_container_operator import GKEPodOperator
 DAG_OWNER = "rpierzina@mozilla.com"
 DAG_EMAIL = ["rpierzina@mozilla.com"]
 
+PROJECT_ID = "moz-fx-data-shared-prod"
+
 # We use a template for the test run UUID in the DAG. Because we base64 encode
 # this query SQL before the template is rendered, we need to use a parameter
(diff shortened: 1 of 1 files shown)


dags/burnham.py
```diff
@@ -25,24 +25,39 @@
 # metric values reported by the Glean SDK across several documents from three
 # different clients are correct.
 TEST_LABELED_COUNTER_METRICS = """
```
jklukas, line 27: Given that the dedupe logic is identical for these various queries, it may be worth extracting it to a constant, so that this becomes:

```
TEST_LABELED_COUNTER_METRICS = WITH_DISCOVERY_V1_DEDUPED + """
  SELECT
    technology_space_travel.key,
    ...
```

hackebrot: That's a great idea!

hackebrot:
Yay, Python 3.7. Thank you, @acmiyaguchi! 🚀

PR #1135, opened 2020-08-28
Fix failing prio-processor DAG and keep GKE cluster online

acmiyaguchi:
This departs from the original idea that the GKE clusters for the prio-processor jobs are ephemeral. This creates a new cluster (if it already doesn't exist) with a single baseline node-pool to keep the cluster alive (a single `g1-small` with 10gb of hdd) and a burst-able node-pool (as per https://medium.com/google-cloud/scale-your-kubernetes-cluster-to-almost-zero-with-gke-autoscaler-9c78051cbf40). 

In addition, I've been adding companion changes in a separate PR and testing the deploy using GCR. It helps to use a docker repository in cloud storage to debug deployments in GKE.

### DIFF
--- a/dags/prio/kubernetes.py
+++ b/dags/prio/kubernetes.py
@@ -23,10 +23,11 @@ def container_subdag(
     env_vars={},
     arguments=[],
     machine_type="n1-standard-1",
-    image="mozilla/prio-processor:v1.6.1",
+    image="gcr.io/amiyaguchi-dev/prio-processor:latest",
     location="us-west1-b",
     owner_label="amiyaguchi",
     team_label="dataeng",
+    **kwargs,
 ):
     """Run a command on an ephemeral container running the
     `mozilla/prio-processor:latest` image.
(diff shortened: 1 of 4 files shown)


dags/prio/kubernetes.py
```diff
@@ -23,10 +23,11 @@ def container_subdag(
     env_vars={},
     arguments=[],
     machine_type="n1-standard-1",
-    image="mozilla/prio-processor:v1.6.1",
-    location="us-west1-b",
+    image="gcr.io/amiyaguchi-dev/prio-processor:latest",
```
acmiyaguchi, line 26: This needs to be changed to the latest pinned tag.

wlach (approved):

acmiyaguchi:
Reran this in my local environment with the the `v2.3.0` image. It looks like it works fine. It takes ~35 minutes to run, which is primarily cluster boot time.

image

PR #1140, opened 2020-09-10
Update burnham glean error test

hackebrot:
[Glean v31.5.0][glean] changed the error reporting behavior. String values that are too long now record `invalid_overflow` rather than `invalid_value`.

Resolve https://github.com/mozilla/burnham/issues/70

[glean]: https://github.com/mozilla/glean/releases/tag/v31.5.0

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -143,9 +143,9 @@ WANT_TEST_EXPERIMENTS = [
     {"experiment": "spore_drive:tardigrade-dna", "document_count": 6},
 ]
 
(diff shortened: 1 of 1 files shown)


Dexterp37 (approved):
Makes sense to me!

jklukas (approved):

jklukas:
I just cleared the DAG so that it will run again with these updates.

jklukas:
Looks like the new run was successful 👍

PR #1141, opened 2020-09-10
Revert "[Bug 1656910] Remove saved-session ping processing"

scholtzan:
Reverts mozilla/telemetry-airflow#1136


fbertsch (commented):
We don't have to revert this one, just the other two

PR #1142, opened 2020-09-10
Allow field addition for fxa export

fbertsch:


dags/fxa_export_to_amplitude.py
```diff
@@ -49,7 +49,8 @@
         project_id='moz-fx-data-shared-prod',
         destination_table='fxa_amplitude_export_v1',
         dataset_id='firefox_accounts_derived',
-        depends_on_past=True
+        depends_on_past=True,
+        allow_field_addition_on_date="20200909"
```
scholtzan, line 53: Isn't it `allow_field_addition_on_date="2020-09-09"`?

scholtzan (approved):

dags/fxa_export_to_amplitude.py
fbertsch: Indeed, it is `{{ds}}` not `{{ds-nodash}}`

PR #1145, opened 2020-09-14
Add job for materialized normalized LTV

fbertsch:


dags/ltv.py
```diff
@@ -118,4 +118,15 @@
     time_partitioning={"type": "DAY", "field": "submission_date"},
 )
 
-ltv_daily >> ltv_revenue_join
+ltv_normalized_view=BigQueryOperator(
+    task_id='ltv_normalized_view',
+    sql=response.read().decode('utf-8'),
+    query_params=[{"name": "submission_date", "parameterType": {"type": "DATE"}, "parameterValue": {"value": "{{ ds }}"}}],
+    destination_dataset_table='moz-fx-data-shared-prod.revenue_derived.client_ltv_v1${{ ds_nodash }}',
```
fbertsch, line 125: Two things here:
1. This dataset does not currently exist
2. This connection id may not have write access to it

cc @jasonthomas, can you help me out with those?

dags/ltv.py
```diff
@@ -118,4 +118,15 @@
     time_partitioning={"type": "DAY", "field": "submission_date"},
 )
 
-ltv_daily >> ltv_revenue_join
+ltv_normalized_view=BigQueryOperator(
```
acmiyaguchi, line 121: The only difference between this and `ltv_revenue_join` is the destination table? As in, these contain the same data? Why not copy the partition instead of rerunning the job?

dags/ltv.py
```diff
@@ -118,4 +118,15 @@
...
+    sql=response.read().decode('utf-8'),
+    query_params=[{"name": "submission_date", "parameterType": {"type": "DATE"}, "parameterValue": {"value": "{{ ds }}"}}],
+    destination_dataset_table='moz-fx-data-shared-prod.revenue_derived.client_ltv_v1${{ ds_nodash }}',
+    bigquery_conn_id='google_cloud_it_revenue',
+    use_legacy_sql=False,
+    default_args=default_args,
+    time_partitioning={"type": "DAY", "field": "submission_date"},
+)
+
+ltv_daily >> ltv_revenue_join >> ltv_normalized_view
```
acmiyaguchi, line 132: This is an unnecessary dependency. 

```suggestion
ltv_daily >> ltv_revenue_join
ltv_daily >> ltv_normalized_view
```

dags/ltv.py
fbertsch: Hah, I forgot to update the sql itself. Thanks for pointing this out.

fbertsch:
@acmiyaguchi this PR will now probably make a bit more sense. The normalized view contains a subset of fields from the revenue join, and writes it to a location that anyone can use.

acmiyaguchi (approved):
LGTM, aside from potential issues with the `google_cloud_it_revenue` connection.

PR #1148, opened 2020-09-15
Use new service account from bug 1665111

fbertsch:


fbertsch:
@jasonthomas Chris has given read access to `moz-it-eip-revenue-users:ltv_derived` for this service account.

jasonthomas (approved):
LGTM

PR #1149, opened 2020-09-16
Add clustering spec to revenue join

fbertsch:


acmiyaguchi (approved):
I believe this will require the table to be dropped, since it hasn't been declared with clustering. An alternative is to add the clustering spec to the query itself.

fbertsch:
Good points - the table is actually clustered right now. We don't have an init query for it, so that will need to be added with this spec defined.

PR #1155, opened 2020-09-21
Schedule stripe events import

relud:


dags/stripe.py
```diff
@@ -0,0 +1,29 @@
...
+    "depends_on_past": False,
+    "start_date": datetime(2020, 6, 11),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=10),
+}
+
+with DAG("stripe", default_args=default_args, schedule_interval="@daily") as dag:
+    publish_public_udfs = gke_command(
```
acmiyaguchi, line 17: This variable name seems wrong.

dags/stripe.py
```diff
@@ -0,0 +1,29 @@
...
+        command=[
+            "bqetl",
+            "stripe",
+            "import",
+            "--date={{ ds }}",
+            "--api-key={{ var.value.stripe_api_key }}",
+            "--resource=Event",
+            "--table-id=moz-fx-data-shared-prod.stripe_external.events_v1",
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
```
acmiyaguchi, line 28: The dag is set implicitly?

acmiyaguchi (approved):

dags/stripe.py
relud: oops, fixed

dags/stripe.py
relud: yeah, same as https://github.com/mozilla/telemetry-airflow/blob/7b4de27b17d3100dba9617c898eee84bcc9587fa/dags/mozfun.py#L16-L23

PR #1159, opened 2020-09-28
Enable DEBUG logging for burnham operators

hackebrot:
We currently don't store DEBUG logs from the burnham application and don't log when the Glean SDK is sending pings. This pull-request sets the respective environment variables for the burnham operators which should  help us diagnose any issues.

Resolve https://github.com/mozilla/burnham/issues/73

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -285,6 +285,8 @@ def burnham_run(
         "BURNHAM_PLATFORM_URL": BURNHAM_PLATFORM_URL,
         "BURNHAM_TEST_RUN": burnham_test_run,
         "BURNHAM_TEST_NAME": burnham_test_name,
+        "BURNHAM_VERBOSE": "true",
+        "GLEAN_LOG_PINGS": "true",
     }
 
     if burnham_spore_drive is not None:


jklukas (approved):

hackebrot:
Thank you for the review @jklukas! 🚀

jklukas:
I reran today's DAG after this was merged, so the most recent logs should reflect these new settings.

hackebrot:
Yes, I just checked the logs and it works. Thank you! 😄

hackebrot:
The logs of the most recent scheduled run appear to miss the DEBUG log messages. 🤔

jklukas:
I see DEBUG-level messages in https://workflow.telemetry.mozilla.org/log?task_id=client2&dag_id=burnham&execution_date=2020-09-28T00%3A00%3A00%2B00%3A00

What appears to be missing?

hackebrot:
The DEBUG messages from the burnham application like here https://workflow.telemetry.mozilla.org/log?task_id=client2&dag_id=burnham&execution_date=2020-09-27T00%3A00%3A00%2B00%3A00

jklukas:
I can see `'BURNHAM_VERBOSE': 'true'` in the rendered template: https://workflow.telemetry.mozilla.org/rendered?dag_id=burnham&task_id=client2&execution_date=2020-09-28T00%3A00%3A00%2B00%3A00

So this is indeed strange. I'll look a bit more into it.

jklukas:
I was able to find [associated logs in the Google Cloud Logging interface](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22moz-fx-data-derived-datasets%22%0Aresource.labels.location%3D%22us-central1-a%22%0Aresource.labels.cluster_name%3D%22bq-load-gke-1%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name%3D%22client2-9bfc0779%22;timeRange=P7D?project=moz-fx-data-derived-datasets) and they _do_ contain the burnham debug messages. It looks to me like the Airflow view is missing earlier messages, and then consistently shows all messages past a certain point in time, and that point in time happens to be after burnham has done all its work, so only the Glean messages remain.

This may be an intermittent problem. I know there are some gotchas with the connection of cloud logs to Airflow and that Airflow sometimes loses that connection. Let's monitor if this is an ongoing problem, and we can bring it up to data ops after we have some more observations.

PR #1161, opened 2020-09-28
Add burnham-bigquery start-timestamp parameter

hackebrot:
This pull-request updates SQL queries of the burnham DAG to use the DAG run `start_date` instead of `CURRENT_TIMESTAMP()` to filter out rows. The new burnham-bigquery parameter is added in https://github.com/mozilla/burnham/pull/74.

Once both pull-requests have been reviewed and approved, we can land the burnham pull-request to deploy the updated Docker image and then land this to update Airflow.


dags/burnham.py
```diff
@@ -33,7 +33,8 @@
   FROM
     `{project_id}.burnham_live.{table}`
   WHERE
-    submission_timestamp > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 3 HOUR)
+    submission_timestamp BETWEEN TIMESTAMP_SUB(@burnham_start_timestamp, INTERVAL 1 HOUR)
```
jklukas, line 36: This should work as written if we pass the param as a TIMESTAMP as discussed in https://github.com/mozilla/burnham/pull/74/files#r496722221 otherwise I think we'll need to wrap all uses in `TIMESTAMP()`

dags/burnham.py
```diff
@@ -202,35 +204,17 @@
 WANT_TEST_SPACE_SHIP_READY_PING = [{"count_documents": 3}]
 
 # Sensor templates for the different tables
-DISCOVERY_SENSOR_TEMPLATE = """
```
jklukas, line 205: If I recall correctly, we use query params in the queries we send to the burnham image and we pass in the query string base64-encoded.  But for the sensor query, Airflow is executing it directly and we don't have the base64 encoding issue, so we use string substitution.

It would be worth adding a comment here to further document that difference.

dags/burnham.py
```diff
@@ -202,35 +204,17 @@
...
+SENSOR_TEMPLATE = """
 SELECT
-  COUNT(*) >= 10
+  COUNT(*) >= {min_count_rows}
 FROM
-  `{project_id}.burnham_live.discovery_v1`
+  `{project_id}.burnham_live.{table}`
```
jklukas, line 211: I like this refactor of making table configurable 👍

jklukas (approved):
r+wc

dags/burnham.py
hackebrot: I changed the type in burnham-bigquery! 👍

dags/burnham.py
hackebrot: Thank you! 😄

dags/burnham.py
hackebrot: Done! ✅

jklukas (approved):

PR #1162, opened 2020-09-29
bigquery_etl specify projects for query files

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/1353


dags/app_store_analytics.py
```diff
@@ -89,7 +89,7 @@
             task_id=f"{derived_table}_query",
             project_id=PROJECT_ID,
             dataset_id=DERIVED_DATASET_ID,
-            sql_file_path=f"sql/{DERIVED_DATASET_ID}/{derived_table}/query.sql",
+            sql_file_path=f"moz-fx-data-shared-prod/sql/{DERIVED_DATASET_ID}/{derived_table}/query.sql",
```
jklukas, line 92:
```suggestion
            sql_file_path=f"sql/moz-fx-data-shared-prod/{DERIVED_DATASET_ID}/{derived_table}/query.sql",
```

And similar for the other changes here.

dags/utils/gcp.py
```diff
@@ -487,7 +487,9 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
-    sql_file_path = sql_file_path or "sql/{}/{}/query.sql".format(dataset_id, destination_table)
+    if not project_id:
+        project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
```
jklukas, line 491: Does this return shared-prod or does it return derived-datasets? I have trouble keeping track, since the Airflow cluster for Airflow is in derived-datasets.

dags/utils/gcp.py
```diff
@@ -487,7 +487,9 @@ def bigquery_etl_query(
...
+    sql_file_path = sql_file_path or "{}/sql/{}/{}/query.sql".format(project_id, dataset_id, destination_table)
```
jklukas, line 492:
```suggestion
    sql_file_path = sql_file_path or "sql/{}/{}/{}/query.sql".format(project_id, dataset_id, destination_table)
```

scholtzan:
Ah, I forgot to push the newest changes

jklukas (approved):

dags/utils/gcp.py
scholtzan: Right, it returns derived-datasets. We should default to shared-prod in this case though.

dags/utils/gcp.py
```diff
@@ -487,7 +487,9 @@ def bigquery_etl_query(
     """
     kwargs["task_id"] = kwargs.get("task_id", destination_table)
     kwargs["name"] = kwargs.get("name", kwargs["task_id"].replace("_", "-"))
-    sql_file_path = sql_file_path or "sql/{}/{}/query.sql".format(dataset_id, destination_table)
+    if not project_id:
```
jklukas, line 490: I often find `project` or `project_id` params to be difficult to reason about. This param here might or might not control each of:

- The sql file path referenced within bigquery-etl
- The project in which the QueryJob runs
- The project in which to write the results

In our case here, I think we need to make sure that the `project_id` passed in determines both the sql_file_path and the project in which we write results. The directory structure is meant to guarantee that coupling.

But, do we want to run the QueryJob to always run in shared-prod? Or do we want the query to execute in the project associated with the destination table? Maybe we need to support both?

In particular, if we add a new use case where we want to allow Airflow to write into a new project, what permissions do we expect to have set up? Do we require that the Airflow user has jobs.create permissions in the destination project?

My gut says it likely makes sense that we want to have QueryJobs run in the destination project, which allows billing and BQ reserved slot usage to be associated with that project. But this may end up being something that needs to be configurable.

Sorry for not thinking through this issue earlier.

scholtzan: Right, currently `project_id` determines where the query is run and the path. In the current setup it would be possible to set `sql_file_path` manually and change `project_id` to some other project that does not necessarily need to match the project in the SQL file path. I think the most common use case is to run the query in the project it is associated to, so maybe this workaround would be sufficient for queries where that is not the case?
Otherwise, we could also add an optional `billing_project` (or some other name) parameter to `bigquery_etl_query` that determines the project to run the query in.

Yes, Airflow will need to have the right permissions (jobs.create) for these projects.

PR #1168, opened 2020-10-15
Fix python 3 string handling issue with backfill UI

jklukas:
I've not had success over the past few days getting the backfill UI to work.
I submit jobs, but they never produce output or result in BQ output.

I checked in cloud logging and found ERROR messages with tracebacks ([example](https://console.cloud.google.com/logs/query;pinnedLogId=2020-10-15T13:57:27.221935242Z%2Fr7d2ll39deckufib8;query=backfill%0Aseverity%3DERROR?project=moz-fx-data-airflow-prod-88e0)):

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base_async.py", line 56, in handle
    self.handle_request(listener_name, req, client, addr)
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/ggevent.py", line 160, in handle_request
    addr)
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base_async.py", line 114, in handle_request
    for item in respiter:
  File "/usr/local/lib/python3.7/site-packages/werkzeug/wsgi.py", line 506, in __next__
    return self._next()
  File "/usr/local/lib/python3.7/site-packages/werkzeug/wrappers/base_response.py", line 45, in _iter_encoded
    for item in iterable:
  File "/app/pvmount/telemetry-airflow/plugins/backfill/main.py", line 133, in read_process
    result = re.match(pattern, line)
  File "/usr/local/lib/python3.7/re.py", line 175, in match
    return _compile(pattern, flags).match(string)
TypeError: cannot use a string pattern on a bytes-like object
```

So looks like a Python 3 migration bug, and submitted tasks are generating
exceptions before they're able to run.

### DIFF
--- a/plugins/backfill/main.py
+++ b/plugins/backfill/main.py
@@ -130,7 +130,7 @@ class Backfill(get_baseview()):
             while g.is_pending():
                 lines = g.readlines()
                 for proc, line in lines:
-                    result = re.match(pattern, line)
+                    result = re.match(pattern, line.decode("utf-8"))
 
                     if result:
                         # Adhere to text/event-stream format
--- a/plugins/backfill/main.py
+++ b/plugins/backfill/main.py
@@ -130,7 +130,8 @@ class Backfill(get_baseview()):
             while g.is_pending():
                 lines = g.readlines()
                 for proc, line in lines:
-                    result = re.match(pattern, line.decode("utf-8"))
+                    line = line.decode("utf-8")
+                    result = re.match(pattern, line)
 
                     if result:
                         # Adhere to text/event-stream format


haroldwoo (approved):

haroldwoo:
have you tested this? I'm not sure if this is why the error is occurring. Can you give me the example backfill command you're trying to run on prod?

jklukas:
> have you tested this? I'm not sure if this is why the error is occurring. 

I have not tested this. It's been a while since I've tried to run this repo locally, but I'll take a stab at it now.

> Can you give me the example backfill command you're trying to run on prod?

`airflow backfill --dry_run -t delete_events -s 2020-10-01T00:00:00 -e 2020-10-12T00:00:00 bqetl_fxa_events`

jklukas:
I have now tested this locally, which uncovered a further bytes/str detail, so pushed an additional commit to fix that.

With the second fix in place, I now see output getting populated in the "Realtime Console Output" tab. It eventually fails with `airflow.exceptions.AirflowException: dag_id could not be found: bqetl_fxa_events. Either the dag did not exist or it failed to parse.` but that makes sense because this is a dag from bigquery-etl which I haven't imported.

I then tested on the `burnham` dag that exists here and got "👍 Backfill has been completed successfully".

haroldwoo:
sweet thanks

jklukas:
Is there any cycling of pods that needs to happen to make this change take effect?

haroldwoo:
I dont think so but will roll the webapp just to be sure. If this doesn't fix your issue, we'll have to look at the actual g.readlines() causing the issue.

PR #1169, opened 2020-10-15
Make client_scalar_probe_counts depend on scalar_percentiles

Ben-Wu:
This is a temporary workaround for a resources exceeded exception when running clients_scalar_probe_counts tracked here: https://github.com/mozilla/bigquery-etl/issues/1432

The task almost always fails on the first run but succeeds afterwards.  My guess is other queries are taking up too many slots on the first run.  This graph https://workflow.telemetry.mozilla.org/gantt?dag_id=glam shows the successful re-run.  The first run starts at the same time as scalar_percentiles.  My hypothesis is that making these two run sequentially will avoid the exception.  

This is a temporary workaround to avoid daily failures.  There's no guarantee that other queries in the project won't also cause the issue.


acmiyaguchi (approved):
For posterity:

image

PR #1174, opened 2020-11-04
[Bug 1675293] Update schema_update_option handling in bigquery_etl_query

scholtzan:
This might be a cleaner fix compared to https://github.com/mozilla/bigquery-etl/pull/1518

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -507,11 +507,9 @@ def bigquery_etl_query(
         + ["--parameter=" + parameter for parameter in parameters]
         + (
             [
-                "--schema_update_option="
-                + "{{ 'ALLOW_FIELD_ADDITION' if ds == %r else '' }}"
-                % allow_field_addition_on_date
+                "--schema_update_option=ALLOW_FIELD_ADDITION"
             ]
-            if allow_field_addition_on_date
+            if allow_field_addition_on_date == "${{ds}}"
             else []
         )
         + list(arguments)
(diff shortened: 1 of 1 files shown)


dags/utils/gcp.py
```diff
@@ -507,11 +507,9 @@ def bigquery_etl_query(
         + ["--parameter=" + parameter for parameter in parameters]
         + (
             [
-                "--schema_update_option="
-                + "{{ 'ALLOW_FIELD_ADDITION' if ds == %r else '' }}"
-                % allow_field_addition_on_date
+                "--schema_update_option=ALLOW_FIELD_ADDITION"
             ]
-            if allow_field_addition_on_date
+            if allow_field_addition_on_date == "${{ds}}"
```
jklukas, line 512: It gets very hard to reason about this, but I don't think we can make this comparison here. Jinja template substitution doesn't occur at the whole DAG level. Instead it happens within the operator. `arguments` is one of the options of `GKEPodOperator` that allows templated arguments.

I believe that this comparison here would compare the value of `allow_field_addition_on_date` to the literal string value `${{ds}}`.

This may be a fundamental limitation that we can't overcome, and we have to deal with the possibility of an empty string value here.

jklukas: Another option here would be to return `""` rather than `"--schema_update_option="`, in which case we would probably need to have the run_query scripts drop empty string arguments from the list.

I think probably leaving behavior as-is, but including a comment to better explain the templating limitation and the significance of empty string is the best bet here.

scholtzan: Ah I see, it's also explained in the Airflow docs: https://airflow.apache.org/docs/stable/concepts.html#jinja-templating

I added a comment though, to avoid future confusion here

jklukas (approved):

PR #1176, opened 2020-11-05
Fix jetstream arguments

scholtzan:


tdsmith (approved):

scholtzan:
It appears that there is some issue with running tests in Airflow. I'll hold off for today with merging this since the exact issue here is unknown (and might break something in Airflow potentially?). I asked in the data ops Slack channel and hope they have an idea what the issue might be here.

PR #1177, opened 2020-11-06
Bug 1675348 - Decouple prio server B from admin and server A

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1675348)

This moves server B into it's own DAG, which means it can be turned off and replaced by a job that's configured elsewhere, like https://github.com/mozilla-services/cloudops-infra/pull/2685


wlach (approved):

acmiyaguchi:
This may need some minor adjustments based on runs later today, I'll make them as necessary.

PR #1178, opened 2020-11-06
Update fission experiment monitoring task dependencies

akkomar:
Crash aggregates table depends on both main and crash ping tables.
This makes crash aggregation task wait for main ping deduplication in addition to the crash ping deduplication.


scholtzan (approved):
Looks like there is still an issues with CircleCI and running the test step :|

PR #1182, opened 2020-11-11
Add burnham tests for disabling Glean upload

hackebrot:
This pull-request makes the following updates to the burnham DAG:

- Add a new burnham client `client4` which submits with `test_name` set to `test_disable_upload`
- Add or extend `WHERE` clauses to existing queries to filter out pings produced by `client4`
- Add a new sensor for the `discovery_v1` table that waits for pings produced by `client4`
- Add a new sensor for the `deletion_request_v1` table that waits for pings produced by `client4`
- Add two new burnham-bigquery operators that verify the data in the aforementioned tables for `client4`

The required burnham capabilities were added in https://github.com/mozilla/burnham/pull/110.

@mdboom and I worked on this together: We would like to extend the `test_deletion_request_ping` scenario with a query that joins data from `discovery_v1` and `deletion_request_v1` to verify that the client ID values are identical for the same client before it disabled the upload. I suggest we review and merge these changes first and then create a follow-up pull-request based on your recommendations to add that in.

Please let us know if you have any questions or suggestions! Thank you. 🧑‍🚀

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -18,6 +18,11 @@ DAG_EMAIL = ["glean-team@mozilla.com", "rpierzina@mozilla.com"]
 
 PROJECT_ID = "moz-fx-data-shared-prod"
 
+# We cover multiple test scenarios with pings submitted from client1, client2
+# and client3. They don't submit pings using a specific test name, but all
+# share the following default test name.
+DEFAULT_TEST_NAME = "test_burnham"
+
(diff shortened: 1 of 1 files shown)


jklukas (approved):
Thanks for the high-level overview. The new code here is clear 👍

hackebrot:
Thanks for the review! 🙇

PR #1183, opened 2020-11-11
Colorblind friendly UI

haroldwoo:
Success/Failed look the same to R/G colorblind

https://bugzilla.mozilla.org/show_bug.cgi?id=1675380


whd (approved):
Does this change any colors other than R/G? Also not sure if there's a standard colorblind friendly scheme we should follow, but s/green/blue/ seems acceptable to me.

Is the plan here to run this in dev first, make sure the new color scheme is acceptable, and then add it to the prod container?

haroldwoo:
I used all the same colors as prod except Green -> Blue. This works in dev (locally), I will need to publish a prod container and deploy it later at some point.

PR #1184, opened 2020-11-12
Fix 1.10.12 regression with gkepodoperator re-runs

haroldwoo:
Somewhat related:

If we ever override reattach_on_restart=True, and have > 1 Pods that match the same labels, the PodOperator will exit with 'More than one pod running with labels...'

https://github.com/apache/airflow/blob/1.10.12/airflow/contrib/operators/kubernetes_pod_operator.py#L273-L276


jklukas:
@scholtzan has thought more deeply about this than I have, so I'll leave review to her.

scholtzan (approved):

PR #1188, opened 2020-11-19
Use highmem node pool for copy_deduplicate_all

scholtzan:
Run `copy_deduplicate_all` on the highmem node pool

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -33,13 +33,20 @@ with models.DAG(
     # This single task is responsible for sequentially running copy queries
     # over all the tables in _live datasets into _stable datasets except those
     # that are specifically used in another DAG.
+    resources = {'request_memory':'10240Mi', 'request_cpu': None,
+                'limit_memory':'20480Mi', 'limit_cpu': None, 'limit_gpu': None}
+
     copy_deduplicate_all = bigquery_etl_copy_deduplicate(
         task_id="copy_deduplicate_all",
         target_project_id="moz-fx-data-shared-prod",
         priority_weight=100,
         # Any table listed here under except_tables _must_ have a corresponding
         # copy_deduplicate job in another DAG.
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -33,13 +33,20 @@
...
         target_project_id="moz-fx-data-shared-prod",
         priority_weight=100,
         # Any table listed here under except_tables _must_ have a corresponding
         # copy_deduplicate job in another DAG.
-        except_tables=["telemetry_live.main_v4"])
+        except_tables=["telemetry_live.main_v4"],
+        kwargs={
+            "node_selectors": {"nodepool" : "highmem"},
+            "resources": resources
+        })
```
jklukas, line 49:
```suggestion
        node_selectors={"nodepool" : "highmem"},
        resources=resources)
```

The last argument of `bigquery_etl_copy_deduplicate` is `**kwargs` which captures all additional keyword arguments. I believe that passing literal `kwargs` here would end end up as `kwargs["kwargs"]` in the destination function.

jklukas (changes requested):

dags/copy_deduplicate.py
scholtzan: Ah indeed

jklukas (approved):
LGTM. I'm concerned about why this job would be consuming substantial memory at all, so not sure whether this will make a difference. But I think it's worth trying, and we should definitely kick off a rerun during the day today to be sure this works.

haroldwoo:
I did add another node to the default pool today. You may want to hold off on merging this to see if that helps tonight.

Looking at the other 2 nodes in the default pool, there are memory usage spikes, although they could be attributed to another GKE job.

scholtzan:
Alright, let's wait and see what happens tonight. If we see the same issue then I'll merge this.

scholtzan:
copy_deduplicate_all did fail again. I'll merge this and then rerun the task

PR #1191, opened 2020-11-26
Create task for graphics dataproc job

BenWu:
Goes with https://github.com/mozilla/python_mozetl/pull/351

I tested that this is working in a sandbox project


dags/graphics_telemetry.py
```diff
@@ -0,0 +1,115 @@
...
+from utils.dataproc import moz_dataproc_pyspark_runner, get_dataproc_parameters
+
+default_args = {
+    "owner": "bewu@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime.datetime(2020, 11, 26),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "bewu@mozilla.com",
+    ],
```
akkomar, line 17: Can we add here someone from platform team who's actually using this data?

akkomar (approved):
LGTM, I'd only add real owner's email for notifications and documentation purposes.

dags/graphics_telemetry.py
BenWu: Yes forgot to do that

PR #1195, opened 2020-12-03
Add burnham scenario for client IDs in deletion-request ping

hackebrot:
I refactored the SQL templates so that we can use multiple templates in a `WITH` block. This allows us to re-use the deduplication logic for multiple tables in a new test scenario for verifying the client ID in deletion-request pings. 👩‍🚀 

Please let me know your thoughts!

Resolve https://github.com/mozilla/burnham/issues/77


jklukas (approved):
Looks great!

PR #1197, opened 2020-12-04
Burnham prune deps

jklukas:
The DAG currently looks a bit cluttered. By pruning these dependencies,
it will be easier to tell what's going on at a glance in the tree view:

https://workflow.telemetry.mozilla.org/tree?dag_id=burnham

Also, Airflow now can render docs on the tree view page.


hackebrot (approved):
LGTM! 📝

Issue #1198, opened 2020-12-07
Task `crash_report_parquet` job is failing in socorro_import failing


willkg:
The `create_dataproc_cluster` task is failing at some point after `dataproc_init.sh` is called. You can see the requirements getting installed in the logs and then nothing after that:

https://workflow.telemetry.mozilla.org/log?task_id=crash_report_parquet&dag_id=socorro_import&execution_date=2020-12-06T00%3A00%3A00%2B00%3A00

```
DAG ID                               Task ID                  Execution date               Try number
-----------------------------------  -----------------------  -------------------------  ------------
socorro_import.crash_report_parquet  create_dataproc_cluster  2020-12-06 00:00:00+00:00             4
socorro_import.crash_report_parquet  run_dataproc_pyspark     2020-12-06 00:00:00+00:00             1
...
[2020-12-07 02:20:58,017] {taskinstance.py:1194} INFO - Marking task as FAILED. dag_id=socorro_import, task_id=crash_report_parquet, execution_date=20201206T000000, start_date=20201207T022052, end_date=20201207T022058
```

```
$ gsutil cat gs://moz-fx-data-prod-dataproc-scratch/google-cloud-dataproc-metainfo/21a4a832-667d-4b80-85c2-72d4076d6736/socorro-import-dataproc-cluster-m/dataproc-initialization-script-0_output
...
Requirement already satisfied: python-dateutil in /opt/conda/anaconda/lib/python3.6/site-packages (from arrow==0.10.0->-r /tmp/requirements.txt (line 1)) (2.8.1)
Requirement already satisfied: pytz>=2011k in /opt/conda/anaconda/lib/python3.6/site-packages (from pandas==0.23.4->-r /tmp/requirements.txt (line 8)) (2020.1)
Requirement already satisfied: setuptools in /opt/conda/anaconda/lib/python3.6/site-packages (from protobuf==3.6.1->-r /tmp/requirements.txt (line 15)) (50.3.0.post20201103)
Requirement already satisfied: py4j==0.10.7 in /opt/conda/anaconda/lib/python3.6/site-packages (from -r /tmp/requirements.txt (line 16)) (0.10.7)
Requirement already satisfied: python-dateutil in /opt/conda/anaconda/lib/python3.6/site-packages (from arrow==0.10.0->-r /tmp/requirements.txt (line 1)) (2.8.1)
[...]

Issue #1200, opened 2020-12-08
fix requirements installed by dataproc_init.sh


willkg:
`dataproc_init.sh` runs pip 20.3.1 with the new resolver and the requirements that it's trying to install don't gel well with one another causing the resolver to take infinite time to figure out a set of packages that meet the requirements.

https://github.com/mozilla/telemetry-airflow/issues/1198 covers changing dataproc_init.sh to use `pip<20.3.0` which puts us back on the old resolver.

This issue covers fixing our requirements so that they don't cause the infinite-time-to-resolve problem.

willkg:
The pip issue suggests switching to a requirements.in file and pip-tools to compile it pushing the dependency resolution step out of installation time. I did that with socorro and tecken a while back. We could do that here.

I don't know where the requirements file is coming from. It's this line in `dataproc_init.sh`:

https://github.com/mozilla/telemetry-airflow/blob/60f7dd1236193da4cfb73ab71210fb129b9638df/dataproc_bootstrap/dataproc_init.sh#L26-L27

Maybe that requirements file is already compiled in which case maybe all we need to do is update the `requirements.in` and recompile it?

jklukas:
> I don't know where the requirements file is coming from. It's this line in dataproc_init.sh

It looks like it's defined within [telemetry-airflow/dataproc_bootstrap](https://github.com/mozilla/telemetry-airflow/blob/60f7dd1236193da4cfb73ab71210fb129b9638df/dataproc_bootstrap/python-requirements.txt) and the contents of that folder are rsync'd to GCS.

Looks like it is not currently compiled, but we could do so within telemetry-airflow and commit the results.

acmiyaguchi:
> The pip issue suggests switching to a requirements.in file and pip-tools to compile it pushing the dependency resolution step out of installation time. I did that with socorro and tecken a while back. We could do that here.

I strongly agree with using pip-tools for compiling requirements. We use it for the airflow container to pin dependencies (https://github.com/mozilla/telemetry-airflow/pull/1008)

PR #1201, opened 2020-12-08
Pin pip to <20.3.0 for now until we fix resolution issue (#1198)

willkg:
Currently, `dataproc_init.sh` updates pip to 20.3.1 which tries to
install the dependencies, but takes > 10 minutes to figure out a
solution set that satisfies the version dependencies.

This has us go back to the previous pip which has the old resolver which
works while we figure out how to straighten out the dependency
requirements.


willkg:
This is a stop-gap fix until we figure out #1200. After we figure that out, we can undo this.

jklukas (approved):

jklukas:
We'll need to check with @haroldwoo about how the `rsync` happens. I don't know if that's automated, so there may be manual steps needed to apply the change.

jklukas:
@haroldwoo pointed out that the rsync is codified in the `.circleci` config, so merging this should be sufficient to kick off a series of events that deploys the new content and will allow new runs to pick it up.

haroldwoo:
Thanks for this

PR #1203, opened 2020-12-08
Change glam bucket for desktop exports to glam-prod

acmiyaguchi:
This changes the bucket for the glam export to the production location at gs://moz-fx-data-glam-prod-fca7-etl-data


Ben-Wu (approved):
As mentioned on slack we should check if this actually gets updated in the dag run

PR #1208, opened 2020-12-16
Fix code url and arguments in symbolication job

BenWu:


dags/crash_symbolication.py
```diff
@@ -64,7 +64,7 @@
             default_args=default_args,
             cluster_name="modules-with-missing-symbols-{{ ds }}",
             job_name="modules-with-missing-symbols",
-            python_driver_code="https://raw.githubusercontent.com/mozilla/python_mozetl/mozetl/symbolication/modules_with_missing_symbols.py",
+            python_driver_code="https://raw.githubusercontent.com/mozilla/python_mozetl/master/mozetl/symbolication/modules_with_missing_symbols.py",
```
jklukas, line 67: Are we aware of any other dependencies right now on `python_mozetl` paths? If not, this may be the golden opportunity to switch that project's default branch to `main`.

jklukas (approved):
r+wc

Just bringing up the `master` branch problem in case that's an easy one to tackle at this point. Fine to move along without that change for now if it's going to be complicated.

dags/crash_symbolication.py
BenWu: There are some jobs like the taar jobs that use files in gcs that match the name of files in python_mozetl.  I don't know where the gcs files come from but I don't think it's python_mozetl.

So I don't know if it's safe to change the branch name but I'll find out.  I'm going to go ahead and merge this because I want to get the job running asap but I'm still doing some work with python_mozetl so I can look into changing the branch name.

PR #1211, opened 2020-12-18
Attempt to fix failing taar colaborative job

eu9ene:
Pass checkpoint directory to the job. I hope it's ok to reuse ETL bucket for this.

Didn't test it locally, except dag parsing. To launch actual job, I need either permissions for prod cluster or to set up sandbox cluster, which is a lot of work. So, testing live. Let me know if you ok with this or I should set everything up and run it from local airflow first.

PR for a corresponding job https://github.com/mozilla/telemetry-batch-view/pull/562

### DIFF
--- a/dags/taar_daily.py
+++ b/dags/taar_daily.py
@@ -13,6 +13,7 @@ from utils.dataproc import (
     moz_dataproc_jar_runner,
 )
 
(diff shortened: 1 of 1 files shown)


haroldwoo (approved):

haroldwoo:
im okay with this testing live. if it fails or we do larger changes we should go over how to test locally and I can send you the needed credentials

PR #1215, opened 2021-01-05
Run copy deduplicate DAG in shared-prod

relud:
fixes https://github.com/mozilla/bigquery-etl/issues/1614

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -39,6 +39,7 @@ with models.DAG(
     copy_deduplicate_all = bigquery_etl_copy_deduplicate(
         task_id="copy_deduplicate_all",
         target_project_id="moz-fx-data-shared-prod",
+        billing_projects=("moz-fx-data-shared-prod",),
         priority_weight=100,
         # Any table listed here under except_tables _must_ have a corresponding
         # copy_deduplicate job in another DAG.
@@ -49,6 +50,7 @@ with models.DAG(
     copy_deduplicate_main_ping = bigquery_etl_copy_deduplicate(
         task_id="copy_deduplicate_main_ping",
         target_project_id="moz-fx-data-shared-prod",
+        billing_projects=("moz-fx-data-shared-prod",),
         only_tables=["telemetry_live.main_v4"],
         parallelism=24,
         slices=100,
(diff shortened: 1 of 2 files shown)


dags/utils/gcp.py
```diff
@@ -533,7 +533,7 @@ def bigquery_etl_copy_deduplicate(
     priority="INTERACTIVE",
     hourly=False,
     slices=None,
-    gcp_conn_id="google_cloud_derived_datasets",
+    gcp_conn_id="google_cloud_shared_prod",
```
relud, line 536: this impacts the copy_deduplicate jobs

dags/copy_deduplicate.py
```diff
@@ -172,7 +172,7 @@
 
     # Daily and last seen views on top of every Glean application.
 
-    gcp_conn_id = "google_cloud_derived_datasets"
+    gcp_conn_id = "google_cloud_shared_prod"
```
relud, line 175: this impacts the downstream `baseline_clients_(daily|last_seen)` jobs

bin/run
```diff
@@ -79,6 +79,7 @@ init_connections() {
     gcp_conn=(
       "google_cloud_airflow_dataproc"
       "google_cloud_derived_datasets"
+      "google_cloud_shared_prod"
```
jklukas, line 82: Did you newly set up this connection?

I assume changing the conn_id changes what kube cluster this connects to. Do we have a `bq-load-gke-1` cluster in shared-prod? I had assumed we'd need to use the existing conn_id, but specify a different project_id parameter for the queries.

relud: > I had assumed we'd need to use the existing conn_id, but specify a different project_id parameter for the queries.

you are totally right. and it's `--billing-project` that we are missing.

dags/utils/gcp.py
```diff
@@ -579,6 +582,7 @@ def bigquery_etl_copy_deduplicate(
         image=docker_image,
         arguments=["script/copy_deduplicate"]
         + ["--project-id=" + target_project_id]
+        + (["--billing-projects"] + list(bp) if billing_projects else [])
```
jklukas, line 585:
```suggestion
        + (["--billing-projects"] + list(billing_projects) if billing_projects else [])
```

jklukas (approved):

PR #1224, opened 2021-01-20
Increase number of slices in main ping deduplication task

akkomar:
copy_deduplicate_main_ping task fails consistently on 2020-01-20 due to increased volume of data. By increasing number of slices we deduplicate in we should be able to avoid out of memory BigQuery errors.


akkomar:
@jklukas @scholtzan FYI - merged this using admin privileges.

jklukas (commented):
Retroactive r+

PR #1225, opened 2021-01-20
experiments_live DAG

scholtzan:
The former `experiment_monitoring_data_export` DAG was consuming a lot of slots, especially due to the large number of wait tasks. To remove those, this PR merges the `bqetl_experiments_live` tasks and `experiment_monitoring_data_export` tasks into one DAG.

cc @haroldwoo


jklukas (approved):

dags/experiments_live.py
```diff
@@ -0,0 +1,84 @@
...
+    'owner': 'ascholtz@mozilla.com',
+    'depends_on_past': False,
+    'start_date': datetime(2021, 1, 8),
+    'email_on_failure': True,
+    'email_on_retry': True,
+}
+
+with DAG('experiments_live',
+         default_args=default_args,
+         concurrency=5,
```
haroldwoo, line 20: this would be 5 for each...meaning on line 35, 59

haroldwoo: 59 being called in a loop also.

PR #1227, opened 2021-01-20
Fix upstream dependency of experiment_live tasks

scholtzan:


dags/experiments_live.py
```diff
@@ -17,7 +17,7 @@
 
 with DAG('experiments_live',
          default_args=default_args,
-         concurrency=1,
+         concurrency=4,
          max_active_runs=1,
```
scholtzan, line 21: Based on the [docs](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dag/index.html#airflow.models.dag.DAG) this will limit the number of tasks concurrently executed in the DAG to 4 and also prevent more than 1 DAG from running at the same time.

dags/experiments_live.py
```diff
@@ -45,6 +45,17 @@
...
+        task_id="export_monitoring_data",
+        command=[
+            "python",
+            "script/experiments/export_experiment_monitoring_data.py",
+            "--date", "{{ ds }}",
+            "--datasets"
+        ] + datasets,
+        docker_image=docker_image
+    )
+
```
scholtzan, line 58: Taking of the pressure of Airflow a little by running the export script in one slot for all the queries, instead of one slot for each query.

dags/experiments_live.py
```diff
@@ -60,25 +70,14 @@
...
-
+        query_etl.set_upstream(experiment_enrollment_aggregates_recents)
         export_monitoring_data.set_upstream(query_etl)
 
     export_daily_active_population_monitoring_data = gke_command(
         task_id=f"export_experiment_enrollment_daily_active_population",
         command=[
             "python",
             "script/experiments/export_experiment_monitoring_data.py",
-            "--date", "{{ ds }}",
```
scholtzan, line 81: Depends on https://github.com/mozilla/bigquery-etl/pull/1693

jklukas (approved):

PR #1228, opened 2021-01-20
Rm unused refs of load_to_bigquery

haroldwoo:


dags/utils/gcp.py
```diff
@@ -15,7 +15,7 @@
 import json
 import re
 
-
+'''
```
haroldwoo, line 18: should we just rm this code

jasonthomas: +1 to remove

jasonthomas (approved):

PR #1229, opened 2021-01-20
Add metric and channel to probe count queries for GLAM

acmiyaguchi:
This fixes the other 2/4 queries for fixing https://github.com/mozilla/bigquery-etl/issues/1660 in addition to https://github.com/mozilla/bigquery-etl/pull/1695. Note that the tables will need to be deleted before these jobs are actually scheduled.

### DIFF
--- a/dags/glam.py
+++ b/dags/glam.py
@@ -175,7 +175,7 @@ histogram_percentiles = bigquery_etl_query(
     project_id=project_id,
     owner="msamuel@mozilla.com",
     date_partition_parameter=None,
-    arguments=("--replace",),
+    arguments=("--replace", "--clustering_fields=metric,channel"),
     dag=dag,
 )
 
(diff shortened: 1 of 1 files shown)


BenWu (approved):
Have you tried creating the clustered tables to see how much the bytes scanned is decreased?  Also are you planning on recreating the tables after deleting or just letting the job create them?

acmiyaguchi:
I haven't, but I'm going to make an educated guess that scans will be under 1GB instead of 750+. I'm also just going to let the job create the table instead of rerunning it.

BenWu:
Ok I was just curious because I've seen clustering sometimes not decrease bytes scanned fully proportionally

acmiyaguchi:
It probably depends on the size of the table and the number of default partitions that they use to store the data. Spark defaulted to 200 partitions (which would give us here something like 4GB partitions), and presumably there's some sort of range filtering built in so it doesn't have to scan over every partition. I'll let you know how much this improves things.

PR #1236, opened 2021-01-25
Switch BigQuerySQLSensorOperator tasks to `reschedule` mode on a dedicated sensor pool

akkomar:
This is a follow up to #1233 where other sensors were switched to 'reschedule' mode. This will help us avoid deadlock situation that occured on 2021-01-20, when we had large number of sensor tasks not letting copy_deduplicate jobs to queue.


fbertsch (approved):
LGTM

PR #1238, opened 2021-01-27
Bug 1689167 - Up probe-scraper mem by 50%

fbertsch:


fbertsch:
This is not the long-term fix. Clearly this job does not need to use all of this memory.

haroldwoo (approved):
If this doesnt schedule we need to create nodes with more memory

haroldwoo:
Ok i checked and this will likely fail. The highmem pool is comprised of n1-highmem-4 instances, which have only 26GB of memory total. https://cloud.google.com/compute/docs/machine-types

haroldwoo:
> This is not the long-term fix. Clearly this job does not need to use all of this memory.

I'm not sure what you mean by this. If the job doesn't require this memory then why are we doing this?

fbertsch:
> > This is not the long-term fix. Clearly this job does not need to use all of this memory.
> 
> I'm not sure what you mean by this. If the job doesn't require this memory then why are we doing this?

What I mean is this is a temporary fix. We are doing it because this blocks schema deploys for the entire pipeline. We need to do a memory profile analysis and see what is causing this leakage.

PR #1239, opened 2021-01-29
Move copy_deduplicate_main_ping back to 100 slices

jklukas:
Reverts #1224 which should no longer be needed after changes as discussed in
[GCP case 26581883](https://console.cloud.google.com/support/cases/detail/26581883?caseId=26581883&accountId=gcp-sa-1256631965&folder=&organizationId=&project=moz-fx-data-shared-prod)


akkomar (approved):
👍

PR #1240, opened 2021-01-29
Prune unused Airflow code related to AWS EMR and Databricks

acmiyaguchi:
This removes now unused code related to EMR and Databricks. This also removes some dangling environment variables that don't have a use.

One follow up from this is to replace the dockerfile.dev with the the docker-compose.yml completely.


akkomar (approved):
LGTM, thanks!

PR #1242, opened 2021-01-29
Bug 1689750 - trim mozaggregator database on a rolling basis

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1689750)

This will trim the mozaggregator database to only retain the last 2 years of data. This is currently running as dry-run, so we can double check the output. The buildids have been accumulated since the last trim.

```bash
python3 -m mozaggregator.trim_db --retention-period 730
To retain 62078 tables...
----------------------------------------
submission_date_beta_45_20190220
submission_date_nightly_46_20190220
submission_date_nightly_44_20190220
submission_date_aurora_51_20190220
submission_date_beta_44_20190220
...
submission_date_nightly_54_20210128
submission_date_beta_43_20210128
submission_date_nightly_45_20210128
submission_date_nightly_71_20210128
submission_date_beta_73_20210128
========================================
To trim 0 tables...
----------------------------------------

========================================
To retain 1981 tables...
----------------------------------------
build_id_nightly_67_20190130
build_id_nightly_67_20190131
build_id_beta_66_20190131
build_id_beta_66_20190201
build_id_nightly_67_20190201
...
build_id_nightly_87_20210126
build_id_beta_86_20210127
build_id_nightly_87_20210127
build_id_beta_86_20210128
build_id_nightly_87_20210128
========================================
To trim 2129 tables...
----------------------------------------
build_id_nightly_39_20150226
build_id_nightly_39_20150227
build_id_nightly_39_20150228
build_id_nightly_39_20150301
build_id_nightly_39_20150302
...
build_id_nightly_66_20190127
build_id_beta_66_20190128
build_id_nightly_66_20190128
build_id_nightly_67_20190128
build_id_nightly_67_20190129
========================================
Dry run enabled, not dropping tables...
```


fbertsch (approved):
Daily backups are still enabled, so we should be fine to go ahead and drop the data. However we should also add an alert to check that we have the data we expect to have. We could use a re:dash alert for that - if it fires, and something did go wrong, we could at revert to the backup.

acmiyaguchi:
I don't believe we have access to the mozaggregator database from redash at the moment. Once I get `mozaggregator2bq` imports running on a regular basis, this is something we can do.

fbertsch:
@acmiyaguchi we do! I have an alert for prerelease aggregates: https://sql.telemetry.mozilla.org/alerts/52

acmiyaguchi:
Ooh, it looks like that query will already trigger on the failure case that you mentioned, too.

fbertsch:
@acmiyaguchi I'm not sure it will - it only checks for data for the most recent day. We want to ensure we have a full two years of history.

PR #1244, opened 2021-02-01
Remove Dockerfile.dev

acmiyaguchi:
It's kind of silly to maintain two copies of a Dockerfile when the only changes between them are the environment. Now that we're using docker-compose, this is no longer an issue.


fbertsch (approved):
Agreed, it does seem a bit silly. Looks like `docker-compose` isn't used by the deploy so prod will be unaffected.

PR #1245, opened 2021-02-02
Set is_delete_operator_pod=False for experiments_live pods

scholtzan:
Setting `is_delete_operator_pod=True` for tasks in `experiments_live` and merging export tasks to reduce number of tasks spawned

### DIFF
--- a/dags/experiments_live.py
+++ b/dags/experiments_live.py
@@ -34,6 +34,7 @@ with DAG('experiments_live',
         depends_on_past=True,
         parameters=["submission_timestamp:TIMESTAMP:{{ts}}"],
         dag=dag,
+        is_delete_operator_pod=True,
     )
 
     experiment_search_aggregates_recents = bigquery_etl_query(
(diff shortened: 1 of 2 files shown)


dags/experiments_live.py
```diff
@@ -103,33 +103,3 @@
...
-            project_id="moz-fx-data-shared-prod",
-            owner="ascholtz@mozilla.com",
-            email=["ascholtz@mozilla.com", "telemetry-alerts@mozilla.com"],
-            date_partition_parameter=None,
-            depends_on_past=True,
-            dag=dag,
-        )
-
-        query_etl.set_upstream(experiment_search_aggregates_recents)
-        export_search_monitoring_data.set_upstream(query_etl)
```
scholtzan, line 135: Reduce number of tasks by merging export tasks.

dags/experiments_live.py
```diff
@@ -89,10 +87,12 @@
             date_partition_parameter=None,
             depends_on_past=True,
             dag=dag,
+            is_delete_operator_pod=True,
```
scholtzan, line 90: This should get passed into the `GKEPodOperator` as parameter: https://github.com/mozilla/telemetry-airflow/blob/17a87679dce6bd4ab84ca840378daf64c0debc0e/dags/utils/gcp.py#L238

dags/experiments_live.py
```diff
@@ -47,36 +48,33 @@
         depends_on_past=True,
         parameters=["submission_timestamp:TIMESTAMP:{{ts}}"],
         dag=dag,
+        is_delete_operator_pod=False,
```
haroldwoo, line 51: you need to set this to True everywhere in this dag, and then add the parameter to bigquery_etl_query and gke_command and set the default as False. Then pass them to the GkePodOperators that are returned.

scholtzan: done

haroldwoo (approved):

PR #1247, opened 2021-02-03
Fix gke cluster scale down for completed pods without rc

haroldwoo:
Tested locally and works to add annotation to pods. 

Annotation should tell cluster auto scheduler that these pods are safe to delete even though they aren't under an replication controller or deployment (something that is preventing our current node pools from scaling down).

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

### DIFF
--- a/dags/operators/gcp_container_operator.py
+++ b/dags/operators/gcp_container_operator.py
@@ -58,11 +58,16 @@ class GKEPodOperator(UpstreamGKEPodOperator):
         if do_xcom_push:
             reattach_on_restart = False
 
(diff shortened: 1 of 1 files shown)


jasonthomas (approved):
Adding https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node for future self.

PR #1249, opened 2021-02-04
Add mozaggregator exports to gcs in parquet format

acmiyaguchi:
This depends on https://github.com/mozilla/docker-etl/pull/15. This will also require a terraform change. 

This schedules exports of the mozaggregator database to in parquet format to gcs in the mozaggregator2bq project.


dags/mozaggregator_prerelease.py
```diff
@@ -146,7 +146,25 @@
     dag=dag,
 )
 
+mozaggregator2bq_extract = gke_command(
+    task_id="mozaggregator2bq_extract",
+    name="mozaggregator2bq_extract",
+    command=["bin/backfill"],
```
fbertsch, line 152: If I'm understanding this correctly, this does not make the data available in BQ? Or is this being exposed as an external table?

acmiyaguchi: No, there's going to have to be one more step in the airflow job to run the load job. This will start populating the parquet dumps however.

fbertsch (commented):
LGTM

acmiyaguchi:
Did the extra commit I push clear an r+, or did you just comment without approving?

fbertsch (approved):
Odd, it must have cleared it.

PR #1252, opened 2021-02-05
TAAR jobs GCS migration

### LINKED ISSUE #1231: Migrate TAAR jobs from S3 to GCS
Unfinished PR from Victor https://github.com/mozilla/telemetry-airflow/pull/1117

eu9ene:
I tested all jobs using local airflow and prod dataproc. Everything looks working.
The produced data is tested in TAAR web service locally.

Ensemble job uses the latest version of TAAR package now. Before it used some very outdated one with S3 dependencies.

Closes #1231


dags/taar_weekly.py
```diff
@@ -2,38 +2,35 @@
...
 
 
 default_args_weekly = {
-    "owner": "anatal@mozilla.com",
+    "owner": "epavlov@mozilla.com",
     "email": [
         "anatal@mozilla.com",
         "mlopatka@mozilla.com",
-        "telemetry-alerts@mozilla.com",
+        "hwoo@mozilla.com",
```
haroldwoo, line 33: We might want to keep telemetry alerts

haroldwoo (approved):
Nice work!

dags/taar_daily.py
```diff
@@ -167,15 +153,16 @@
         job_name="Train_the_Collaborative_Addon_Recommender",
         main_class="com.mozilla.telemetry.ml.AddonRecommender",
         jar_urls=[
+            # GCS bucket for testing is located in `cfr-personalization-experiment` project
+            # 'gs://taar_models/tmp/telemetry-batch-view-1.2.jar'
             "https://s3-us-west-2.amazonaws.com/net-mozaws-data-us-west-2-ops-ci-artifacts"
```
haroldwoo, line 158: should this change to a gcp bucket as well then? (instead of s3)

eu9ene: good question. It requires reconfiguration of deployment pipeline of Scala jobs and I believe it is out of scope of the current task. It doesn't really affect TAAR operations, so we might think of this later. We can create an issue for this in `telemetry-batch-view` repo

jobs/taar_lite_guidguid.py
```diff
@@ -94,18 +77,47 @@ def get_addons_per_client(broadcast_amo_whitelist, users_df):
     return df
 
 
+def get_table(view):
+    """Helper for determining what table underlies a user-facing view, since the Storage API can't read views."""
+    bq = bigquery.Client()
```
haroldwoo, line 82: nit: empty line here

jobs/taar_lite_guidguid.py
```diff
@@ -94,18 +77,47 @@ def get_addons_per_client(broadcast_amo_whitelist, users_df):
...
+    # partition filter is required, so try a couple options
+    for partition_column in ["DATE(submission_timestamp)", "submission_date"]:
+        try:
+            job = bq.query(
+                f"SELECT * FROM `{view}` WHERE {partition_column} = CURRENT_DATE",
+                bigquery.QueryJobConfig(dry_run=True),
+            )
+            break
+        except Exception:
+            continue
```
haroldwoo, line 93: log some error

eu9ene:
this PR depends on this one btw https://github.com/mozilla/telemetry-batch-view/pull/566 because we need to update jars first

jobs/taar_lite_guidguid.py
```diff
@@ -94,18 +77,47 @@ def get_addons_per_client(broadcast_amo_whitelist, users_df):
...
+    else:
```
haroldwoo, line 94: else block doesn't match try/except indentation

jobs/taar_lite_guidguid.py
```diff
@@ -94,18 +77,47 @@ def get_addons_per_client(broadcast_amo_whitelist, users_df):
...
+    Firefox Clients Daily telemetry table
+    https://docs.telemetry.mozilla.org/datasets/batch_view/clients_daily/reference.html
+
+    BUG 1485152: PR include active_addons to clients_daily table:
+    https://github.com/mozilla/telemetry-batch-view/pull/490
+    """
+
     # Could scale this up to grab more than what is in
     # longitudinal and see how long it takes to run.
     s3_url = "gs://moz-fx-data-derived-datasets-parquet/clients_daily/v6/submission_date_s3={}".format(
```
haroldwoo, line 119: nit: change var name from s3_url to gcs_url

jobs/taar_lite_guidguid.py
```diff
@@ -226,78 +274,68 @@ def load_s3(result_df, date, prefix, bucket):
...
-    transfer.upload_file(source_file_name, bucket, key_path)
-
-
-def load_amo_external_whitelist():
+    byte_data = json.dumps(json_obj).encode("utf8")
+
+    byte_data = bz2.compress(byte_data)
+    logger.info(f"Compressed data is {len(byte_data)} bytes")
+
+    client = storage.Client()
```
haroldwoo, line 302: i generally like to wrap any google client code in a try/catch for exception/error logging, but this is upto you

jobs/taar_lite_guidguid.py
eu9ene: it was copied from official telemetry docs https://docs.telemetry.mozilla.org/tools/spark.html#storage-api. I can add logging, but probably they don't consider an exception here as an error.

jobs/taar_similarity.py
```diff
@@ -30,103 +33,32 @@
...
-
-
-@contextlib.contextmanager
-def selfdestructing_path(dirname):
-    import shutil
-
-    yield dirname
-    shutil.rmtree(dirname)
-
+def read_from_gcs(fname, prefix, bucket):
```
haroldwoo, line 39: at quick glance, it seems the reading/writing to gcs code is duplicated and could be written once as a common lib, but upto you again

eu9ene: yes, I noticed this. this PR is based on the original Victor's work, and I didn't change much, just fixed bugs. It would be a good idea of course but requires some development time which we don't have. I can add it as todo for future

jobs/taar_lite_guidguid.py
eu9ene: good catch, I'll fix this. we probably should report to telemetry docs owners

eu9ene: actually, this is `for .. else` construction which I see the first time myself:) https://book.pythontips.com/en/latest/for_-_else.html#else-clause That's how they handle errors here.

haroldwoo: ah, cool

jobs/taar_lite_guidguid.py
eu9ene: good idea, added

jobs/taar_lite_guidguid.py
eu9ene: renamed

jobs/taar_lite_guidguid.py
eu9ene: added warning

jobs/taar_lite_guidguid.py
eu9ene: fixed formatting everywhere, a bit more readable I guess

dags/taar_daily.py
eu9ene: added note and issue

PR #1255, opened 2021-02-10
Experiments live DAG improvements

scholtzan:
It happened a few times in the past weeks that a task failed (e.g. pod took too long to start) and due to `depends_on_past` all future tasks came to a halt and I had to go into Airflow to clear the failed task. The DAG has `max_active_runs=1` set to ensure only one instance of the DAG is running at a time. This ensures that only a new instance is started after the previous one completed.


jklukas:
This is a somewhat novel concept, so I'd like to see code comments to make this clear. Something like:

```
# We rely on max_active_runs=1 at the DAG level to manage the dependency on past runs.
depends_on_past=False,
```

scholtzan:
The queries do not depend on previous runs. We had `depends_on_past=True` set to limit the number of tasks that Airflow executes, but `max_active_runs=1` has the same effect

jklukas (approved):

PR #1256, opened 2021-02-17
Set tmp-project for glam clustered queries

BenWu:
for mozdata

goes with https://github.com/mozilla/bigquery-etl/pull/1821


dags/glam.py
```diff
@@ -113,7 +114,8 @@
         "python3", "script/glam/run_scalar_agg_clustered_query.py",
         "--submission-date", "{{ds}}",
         "--dst-table", "scalar_percentiles_v1",
-        "--project", project_id,
+        "--dst-project", project_id,
```
acmiyaguchi, line 117: This should still be `--project`, unless there was a typo in the previous PR. I'd recommend keeping the `--project` flag though since this is the api compatible way forward.

acmiyaguchi (changes requested):

dags/glam.py
BenWu: Oops yes I changed it to dst project in the other pr first and then changed it back. Thanks

acmiyaguchi (approved):

PR #1269, opened 2021-03-18
Add make target for sandbox gke clusters to simplify local dev/testing

haroldwoo:
Open for suggestions/comments/concerns

- Adds `make gke` and `make clean-gke` targets
~~Changes local environment to only load dags in the new devdags/ folder. (This may be controversial, but saves about 3-4 minutes of startup time involving importing dummy variables and connections)~~
- More info: go/wtmodev

infra pr https://github.com/mozilla-services/cloudops-infra/pull/2967

### DIFF
--- a/bin/start_gke
+++ b/bin/start_gke
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# This is to be used by the Makefile for a start gke target.
+
+set -eo pipefail
+
+# Set Env var MY_LOCAL_IP or use icanhazip.com to fetch it
+# https://major.io/icanhazip-com-faq/#what-about-my-privacy
+MY_IP=${MY_LOCAL_IP:-$(curl icanhazip.com)}
+echo "local ip is $MY_IP"
+
+USERNAME=$(gcloud config get-value account | awk -F"@" '{print $1}')
+CLUSTERNAME=$USERNAME-gke-sandbox
+
+# Create GKE Cluster - No TTL available, will need a external monitor with cleanup
+if gcloud container clusters describe $CLUSTERNAME --project moz-fx-data-gke-sandbox --region us-west1 >/dev/null 2>&1; then
+    echo "cluster $CLUSTERNAME exists"
+else
+    echo "cluster $CLUSTERNAME doesn't exist. creating..."
+    gcloud container clusters create $CLUSTERNAME \
+           --enable-stackdriver-kubernetes \
+           -m n1-standard-4 \
+           --release-channel="stable" \
+           --enable-master-authorized-networks \
+           --master-authorized-networks="$MY_IP/32" \
+           --region us-west1 \
+           --num-nodes=1 \
+           --scopes="cloud-platform" \
+           --service-account="data-gke-sandbox-runner@moz-fx-data-gke-sandbox.iam.gserviceaccount.com" \
+           --project moz-fx-data-gke-sandbox
+
+fi
+
+echo "fetching secret..."
+JSON_CREDS=$(gcloud secrets versions access latest --secret="gke-sandbox-creds" --project moz-fx-data-gke-sandbox)
+
+# Upload secret to local wtmo
+GCP_CONN_ID="google_cloud_gke_sandbox"
+
+if ! docker ps | grep _web; then
+    echo "ERROR: Airflow container is likely not running (or docker). Run 'make up' to start airflow containers"
+fi
+
+CONTAINER_ID=$(docker ps | grep _web | cut -d' ' -f1)
+
+echo "Web container id is $CONTAINER_ID. Adding gcp connection..."
+docker exec $CONTAINER_ID airflow connections -d --conn_id $GCP_CONN_ID
+
+docker exec $CONTAINER_ID airflow connections -a --conn_id $GCP_CONN_ID \
+       --conn_type google_cloud_platform \
+       --conn_extra "$JSON_CREDS"
+
+echo "visit https://go.corp.mozilla.com/wtmodev for more info"
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,9 @@ help:
 	@echo "  run        Run a airflow command"
 	@echo "  secret     Create a secret to be used for a config variable"
 	@echo "  shell      Opens a Bash shell"
-	@echo "  up         Runs the whole stack, served under http://localhost:8000/\n"
+	@echo "  up         Runs the whole stack, served under http://localhost:8000/"
+	@echo "  gke        Create a sandbox gke cluster for testing"
+	@echo "  clean-gke  Delete the sandbox gke cluster"
 	@echo "  stop       Stops the docker containers"
 
 build:
@@ -39,6 +41,12 @@ stop:
 
 up:
 	docker-compose up
+
+gke:
+	bin/start_gke
+
+clean-gke:
+	bin/stop_gke
 
 test:
 	tox
--- a/README.md
+++ b/README.md
@@ -101,13 +101,14 @@ sed -i "s/10001/$(id -u)/g" Dockerfile.dev
 
 ### Testing GKE Jobs (including BigQuery-etl changes)
 
-For now, follow the steps outlined here to create a service account: https://bugzilla.mozilla.org/show_bug.cgi?id=1553559#c1.
-
-Enable that service account in Airflow with the following:
+See https://go.corp.mozilla.com/wtmodev for more details.
 
 ```
 make build && make up
-./bin/add_gcp_creds $GOOGLE_APPLICATION_CREDENTIALS
+make gke
+
+When done:
+make clean-gke
 ```
 
 From there, [connect to Airflow](localhost:8000) and enable your job.
--- a/bin/stop_gke
+++ b/bin/stop_gke
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# This is to be used by the Makefile for a stop gke target.
+
+set -eo pipefail
+
+USERNAME=$(gcloud config get-value account | awk -F"@" '{print $1}')
+CLUSTERNAME=$USERNAME-gke-sandbox
+
+if gcloud container clusters describe $CLUSTERNAME --region us-west1 --project moz-fx-data-gke-sandbox >/dev/null 2>&1; then
+    gcloud container clusters delete $CLUSTERNAME --region us-west1 --quiet --project moz-fx-data-gke-sandbox
+else
+    echo "cluster $CLUSTERNAME does not exist"
+fi


bin/start_gke
```diff
@@ -0,0 +1,52 @@
...
+set -eo pipefail
+
+MY_IP=`curl icanhazip.com`
+echo "local ip is $MY_IP"
+
+USERNAME=`gcloud config get-value account | awk -F"@" '{print $1}'`
+CLUSTERNAME=$USERNAME-gke-sandbox
+
+# Create GKE Cluster - No TTL available, will need a external monitor with cleanup
+gcloud config set project moz-fx-data-gke-sandbox
```
acmiyaguchi, line 14: You may want to set the project back at the end of the script, or use an explicit `--project` flag with the gcloud tool.

bin/run
```diff
@@ -153,7 +153,7 @@ case $1 in
     airflow upgradedb
 
     # Only init connections in dev
-    [ ! -z ${DEVELOPMENT+check} ] && init_connections && init_variables
+    #[ ! -z ${DEVELOPMENT+check} ] && init_connections && init_variables
```
acmiyaguchi, line 156: I think we should make this configurable via an environment variable in the docker-compose file, so we can still get the old behavior by setting a flag.

docker-compose.yml
```diff
@@ -22,6 +22,9 @@ services:
       - $PWD:/app
       # mount a dummy file to google application credentials, if not provided
       - ${GOOGLE_APPLICATION_CREDENTIALS:-./.add_credentials.md}:/tmp/.credentials
+      - $PWD/devdags:/app/devdags
```
acmiyaguchi, line 25: I'm not sure how ergonomic this will actually be. If I want to modify an existing workflow, I would have to eventually include the appropriate files and init variables. This is probably useful as standalone of the bigquery-etl workflow, though.

acmiyaguchi (commented):
I think there should be a way to target the old behavior, even if it takes a few minutes to come online.

bin/run
fbertsch: Agreed

fbertsch (commented):
Echoing Anthony's comments on workflow. Most of the work we do in GKE Pod Operators involves some sort of connection/variable setup, simply because they don't fit in the bq_etl model.

docker-compose.yml
haroldwoo: I'm not sure what you mean. If modifying an existing dag, you would need to move it to the devdags folder, what other files and variables are you referring to? IIRC the current workflow only injects dummy variables, so you would need to import the real variables when testing anyways.  In terms of files, I mount the operators/utils folders into the devdags folder as well.

haroldwoo: It's also nice to only have to deal with the dag you're working on, and not load the others + vars/conns

acmiyaguchi: Example: if I want to test `dags/mozaggregator_prerelease.py`, I would have to copy it over to the devdags directory to test, then copy it back over. You're right about the variables because I do already have import scripts handy.

docker-compose.yml
```diff
@@ -43,6 +46,7 @@ services:
       - AIRFLOW_RESULT_URL=redis://redis:6379/0
       - AIRFLOW_FLOWER_PORT=5555
       - AIRFLOW_DATABASE_URL=mysql://root:secret@db:3306/airflow
+      - AIRFLOW__CORE__DAGS_FOLDER=/app/devdags
```
acmiyaguchi, line 49: This breaks the `test-parse` CI, since the devdags folder is empty. We do want to test against all of the existing dags to make sure we're not making bad commits.

docker-compose.yml
haroldwoo: Yeah this discussion is why I added you and fbertsch for review. 

IMO longer startup times, less logs, and little UI quirks (e.g. when toggling a dag on, and quickly refreshing, sometimes the wrong dag seems to be toggled on) would go away when only dealing with a single dag. 

The trade off is having to move dags before merging. But at least every new dag wouldn't require devs to add dummy vars/conns.

I did some testing and with prebuilt docker images, the startup time of local airflow without dags/conns/vars is ~10 seconds versus 3 minutes

bin/start_gke
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# This is to be used by the Makefile for a start gke target.
+
+set -eo pipefail
+
+MY_IP=`curl icanhazip.com`
```
whd, line 7: For bash style, we generally prefer `$()` to ``` `` ```. There appears to be only one use of ``` `` ``` in the top-level `airflow.sh` file (side note: looks like this can probably be deleted since it appears to reference AWS).

As to using an arbitrary site to look up your public ip... I'd really prefer if we had a better system than this, but anything else (e.g. IAP and GKE, bastion hosts) are probably significantly more effort. I'd almost prefer making the cluster accessible from anywhere since there's additional auth at the k8s layer.

bin/start_gke
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Web container id is $CONTAINER_ID. Adding gcp connection..."
+docker exec $CONTAINER_ID airflow connections -d --conn_id $GCP_CONN_ID
+
+docker exec $CONTAINER_ID airflow connections -a --conn_id $GCP_CONN_ID \
+       --conn_type google_cloud_platform \
+       --conn_extra "$JSON_CREDS"
+
+echo "visit go.corp.mozilla.com/wtmodev for more info"
+
```
whd, line 52:
```suggestion
```

bin/start_gke
```diff
@@ -0,0 +1,52 @@
...
+CONTAINER_ID=$(docker ps | grep _web | cut -d' ' -f1)
+
+echo "Web container id is $CONTAINER_ID. Adding gcp connection..."
+docker exec $CONTAINER_ID airflow connections -d --conn_id $GCP_CONN_ID
+
+docker exec $CONTAINER_ID airflow connections -a --conn_id $GCP_CONN_ID \
+       --conn_type google_cloud_platform \
+       --conn_extra "$JSON_CREDS"
+
+echo "visit go.corp.mozilla.com/wtmodev for more info"
```
whd, line 51:
```suggestion
echo "visit https://go.corp.mozilla.com/wtmodev for more info"
```

bin/start_gke
```diff
@@ -0,0 +1,52 @@
...
+           --enable-master-authorized-networks \
+           --master-authorized-networks="$MY_IP/32" \
+           --region us-west1 \
+           --num-nodes=1 \
+           --scopes="cloud-platform" \
+           --service-account="data-gke-sandbox-runner@moz-fx-data-gke-sandbox.iam.gserviceaccount.com"
+fi
+
+echo "fetching secret..."
+JSON_CREDS=`gcloud secrets versions access latest --secret="gke-sandbox-creds"`
```
whd, line 33: Looks like I missed this in the ops PR. Is there a reason we don't allow `roles/iam.serviceAccountKeyAdmin` and allow created keys directly instead of using a static secret? In theory we could create / destroy these at the same cadence as the GKE clusters themselves, but we need to consider cleanup steps as well I suppose.

bin/start_gke
whd: I agree we should probably use `--project` explicitly instead of changing the user's general setting. Another option (that needs to be careful about error handling) is to get the current project before setting it to the gke and project and to reset the project after the script runs.

docker-compose.yml
whd: I don't have any real opinion here (as I won't be a consumer of this particular tooling), but 10s seems a lot nicer than 3m. Ergonomically though copying things around doesn't seem ideal.

fbertsch: Testing on this dev cluster is going to be a pretty high-touch piece anyways, so copying over is not really a big deal. I think we should be good to go with this improvement as long as we properly document this workflow in the README: https://github.com/mozilla/telemetry-airflow#testing-gke-jobs-including-bigquery-etl-changes

bin/start_gke
haroldwoo: just for simplicity. Also, isnt granting view on secrets better than granting serviceAccountKeyAdmin?

bin/start_gke
```diff
@@ -0,0 +1,52 @@
...
+
+set -eo pipefail
+
+MY_IP=`curl icanhazip.com`
+echo "local ip is $MY_IP"
+
+USERNAME=`gcloud config get-value account | awk -F"@" '{print $1}'`
+CLUSTERNAME=$USERNAME-gke-sandbox
+
+# Create GKE Cluster - No TTL available, will need a external monitor with cleanup
```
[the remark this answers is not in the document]
haroldwoo: theres another bug that causes the containers to restart locally, so i'll hold off on that

docker-compose.yml
haroldwoo: rm'd.

docker-compose.yml
haroldwoo: i've removed this piece and kept this pr involving gke only. If startup time is a goal in the future, we can revist this

bin/start_gke
haroldwoo: I've been using this site for years without issue. If we want to open this cluster to the public, it may widen the surface area of attack vectors as this is a public repo with the projectid and other things visible.

acmiyaguchi (approved):
I think this looks good as is. 

There are some good ideas with mounting the /devdags folder that I think may be useful elsewhere (particularly bigquery-etl), but we'll want to be careful how it's configured to ensure that we're running CI correctly on PRs.

bin/start_gke
whd: Since this is a development environment tool it's probably fine to rely on such a third-party service, but I'd request using `$(curl https://icanhazip.com)` and  a comment/link to https://major.io/icanhazip-com-faq/#what-about-my-privacy, as well as the option to override this via environment variable (i.e. only run this when the variable isn't set).

Again though, I'm not likely to use this tooling so if others are fine with it as-is I'm not going to object.

bin/start_gke
whd: The `serviceAccountKeyAdmin` idiom is modeled after cloudops-deploylib and has the advantage that secrets are temporal (they're created per-run and deleted afterward). It requires proper cleanup though and we have an [open issue](https://github.com/mozilla-services/cloudops-infra-deploylib/issues/105) for that in deploylib land.

On the whole I'd prefer to not have static secrets that are susceptible to requiring rotation (when e.g. a DE leaves) but I'm also not strenuously objecting to it, especially when we haven't always had success implementing the temporal idiom in more secure environments.

PR #1271, opened 2021-03-23
Bug 1700121 Add sleep task to burnham DAG

jklukas:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1700121#c10

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -6,6 +6,7 @@ import base64
 import datetime
 import json
 import uuid
+import time
 
 from airflow import DAG
 from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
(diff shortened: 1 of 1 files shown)


dags/burnham.py
```diff
@@ -518,6 +519,30 @@ def encode_test_scenarios(test_scenarios):
...
 
 
+def do_sleep(minutes):
+    """Sleep for the given number of minutes.
+
+    Writes out an update every minute to give some indication of aliveness.
+    """
+    print(f"Configured to sleep for {minutes} minutes. Let's begin.")
+    for i in range(minutes, 0, -1):
+        print(f"{i} minute(s) of sleeping left")
```
hackebrot, line 529: Do you generally recommend `print` commands over logging in Airflow operators?

dags/burnham.py
```diff
@@ -518,6 +519,30 @@ def encode_test_scenarios(test_scenarios):
...
+        time.sleep(60)
+
+
+def sleep_task(minutes, task_id):
+    """Return an operator that sleeps for a certain number of minutes.
+
+    :param int    minutes: [Required] Number of minutes to sleep
+    :param string task_id: [Required] ID for the task
```
hackebrot, line 537:
```suggestion
    :param string task_id: [Required] ID for the task
    
    :return: PythonOperator
```

hackebrot (changes requested):
Looks good to me! I added one general question and one suggestion. 📝

dags/burnham.py
jklukas: Logging rather printing is a very good idea 👍

hackebrot (approved):
LGTM! 👩🏻‍🚀

PR #1272, opened 2021-03-23
Bugfix for new sleep operator in burnham DAG

jklukas:


hackebrot (approved):
LGTM! 🚀 

(For reference https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/python.html#passing-in-arguments)

PR #1275, opened 2021-04-02
Schedule telemetry_derived.core_clients_first_seen

acmiyaguchi:
Dependency of [this jira ticket](https://jira.mozilla.com/browse/DS-1424) and the scheduling of https://github.com/mozilla/bigquery-etl/pull/1934.

This also formats the module with black. This is split across two commits.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -118,6 +118,12 @@ with models.DAG(
 
     # Daily and last seen views on top of every Glean application.
 
+    # The core clients first seen dataset is a dependency to glean usage
+    # queries. Ideally, it would belong inside of a generated bigquery-etl DAG
+    # (e.g. bqetl_core), but this would require splitting this DAG into three
+    # separate parts threaded by sensors. Since the first_seen_table will end up
+    # being part of the clients daily table in this DAG, it will be easier to
+    # reason about dependencies in this single DAG while it is being developed.
     telemetry_derived__core_clients_first_seen__v1 = bigquery_etl_query(
         task_id="telemetry_derived__core_clients_first_seen__v1",
         destination_table="core_clients_first_seen_v1",
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -3,10 +3,12 @@ import datetime
 from airflow import models
 from airflow.operators.sensors import ExternalTaskSensor
 from airflow.operators.subdag_operator import SubDagOperator
-from utils.gcp import (bigquery_etl_copy_deduplicate,
-                       bigquery_etl_query,
-                       gke_command,
-                       bigquery_xcom_query)
+from utils.gcp import (
+    bigquery_etl_copy_deduplicate,
+    bigquery_etl_query,
+    gke_command,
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -101,13 +111,27 @@
...
         dag=dag,
-        arguments=('--schema_update_option=ALLOW_FIELD_ADDITION',),
+        arguments=("--schema_update_option=ALLOW_FIELD_ADDITION",),
     )
 
     copy_deduplicate_main_ping >> bq_main_events
 
     # Daily and last seen views on top of every Glean application.
 
+    telemetry_derived__core_clients_first_seen__v1 = bigquery_etl_query(
```
jklukas, line 121: It's awkward that this query is scheduled explicitly here rather than being generated in `bqetl_core`. I can understand that it's desirable to have it here in order to express the future dependency of baseline_clients_first_seen on it.

The "right" thing to do here would be to refactor all of this out of copy_deduplicate. In that scenario, `core_clients_first_seen` would be in `bqetl_core` and we would create a separate DAG in this repo for `glean_usage` that would run at 02:30 and contain the baseline first_seen, daily, and last_seen tasks.

I think that's probably a distraction at this point, but we should add a code comment here to explain the potentially surprising layout.

jklukas (changes requested):

dags/copy_deduplicate.py
acmiyaguchi: If I'm understanding correctly, there would be three DAGs involved with the ETL:

* copy_deduplicate (scheduled at 01:00 UTC), containing live -> stable table queries
* bqetl_core (scheduled at 02:00ish UTC), containing ETL for the core ping
* glean_usage (scheduled 02:30 UTC), containing ETL for glean usage stuff

Sensors or offsetting queries is one way to go, but I think it makes sense for all of the dependencies to be in one place in the meanwhile. I would actually be very interested in scheduling the generated queries explicitly instead of via a wrapper script, but I don't know how that would affect airflow performance.

jklukas: > I would actually be very interested in scheduling the generated queries explicitly instead of via a wrapper script, but I don't know how that would affect airflow performance.

I agree this would be ideal, but I also worry about how this might affect Airflow performance. It might be fine. If we want to seriously pursue that path, we should consult with @haroldwoo.

> If I'm understanding correctly, there would be three DAGs involved with the ETL

Yep, your details match what I was thinking.

> Sensors or offsetting queries is one way to go, but I think it makes sense for all of the dependencies to be in one place in the meanwhile

Yes, I think the best thing to do at the moment is document why this query is included in this DAG.

jklukas (approved):
r+ as long as we add some code comments to discuss why this query is scheduled in this DAG rather than in bqetl_core.

acmiyaguchi:
I'm going to go ahead and reinitialize the entire table from scratch, since it is now a month or two out of date.

acmiyaguchi:
It was actually over a year out of date, the job id for this query is `moz-fx-data-shared-prod:US.bquxjob_7f45bd01_178a3aef1b8`.

PR #1277, opened 2021-04-06
Amplitude export should not write to derived-datasets

jklukas:
We've been cleaning out `moz-fx-data-derived-datasets`, so it's no longer appropriate to be writing temporary data there.

### DIFF
--- a/dags/utils/amplitude.py
+++ b/dags/utils/amplitude.py
@@ -89,9 +89,11 @@ def export_to_amplitude(
             recreate_view >> wait_for_data
 
         # Create the table with yesterday's data
-        project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
         temp_table_name = table_or_view + '_{{ ds_nodash }}'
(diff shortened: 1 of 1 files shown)


dags/utils/amplitude.py
```diff
@@ -89,9 +89,11 @@ def export_to_amplitude(
             recreate_view >> wait_for_data
 
         # Create the table with yesterday's data
-        project_id = GoogleCloudBaseHook(gcp_conn_id=gcp_conn_id).project_id
         temp_table_name = table_or_view + '_{{ ds_nodash }}'
-        fully_qualified_table_name = '{}.{}.{}'.format(project_id, dataset, temp_table_name)
+        fully_qualified_table_name = '{}.{}.{}'.format(
```
fbertsch, line 93: I always assumed there was a permissions reason we retrieved the project id from the gcp connection, but in retrospect that doesn't matter 🤷

fbertsch (approved):

PR #1279, opened 2021-04-08
Optimize copy_deduplicate parallelism

jklukas:
We have previously been unable to run copy_deduplicate as a single query
for main_v4 due to resource constraints in BigQuery causing queries to fail.
See https://github.com/mozilla/telemetry-airflow/pull/1224 as an example.

BQ engineering, though, has made some backend changes to improve efficiency
for cases like this, and recent testing has shown that running a single query
is effective. This should be somewhat more cost efficient than splitting the
query into slices, and it can potentially lead to more effective clustering
and less fragmentation.

We also increase parallelism for non-main_v4 queries. Most of these are small
and overall time of the job may be dominated by simply waiting for BQ per-query
overhead.

I will plan to immediately rerun a day once this is merged as a sanity check
that the configuration is correct. If we do see a failure in
coming days or weeks, we can revert to the previous configuration.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -75,8 +75,6 @@ with models.DAG(
         target_project_id="moz-fx-data-shared-prod",
         billing_projects=("moz-fx-data-shared-prod",),
         only_tables=["telemetry_live.main_v4"],
-        parallelism=24,
-        slices=100,
         owner="jklukas@mozilla.com",
         email=[
             "telemetry-alerts@mozilla.com",
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -27,6 +27,14 @@ A few immediate downstream tables are also included in this DAG.
 In early 2021, manual reruns of `copy_deduplicate` were leading to empty
 partitions, but the root cause has been fixed. See
 [bug 1690363](https://bugzilla.mozilla.org/show_bug.cgi?id=1690363).
+
+## Changelog
+
+In April 2021, `copy_deduplicate_main_ping` was moved from a 100-slice
+configuration to a single-query configuration, which will change the
+performance profile and is intended to be more efficient and slightly
+faster. See [telemetry-airflow#1279](
+https://github.com/mozilla/telemetry-airflow/pull/1279/files)
 """
 
 default_args = {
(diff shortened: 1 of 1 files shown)


jklukas:
Looking at copy_deduplicate_all, the following query is interesting to see how individual queries perform:

```
SELECT
 creation_time, job_id, destination_table, referenced_tables, total_bytes_billed, timestamp_diff(end_time, start_time, MINUTE) AS duration
FROM `region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
where query like '%distinct_document_ids%'
and destination_table.dataset_id != 'tmp'
and DATE(creation_time) = '2021-04-08'
order by duration DESC
```

`event_v4` is by far the biggest table and takes ~40 minutes to complete. There are about 10 tables that take between 2 and 10 minutes to complete. Then there are about 250 tables that each take less than a minute. With parallelism increased from 4 to 10, I expect we'll be able to run through the 250 small tables much more quickly.

dags/copy_deduplicate.py
```diff
@@ -75,16 +87,24 @@
...
+        parallelism=1,
         owner="jklukas@mozilla.com",
         email=[
             "telemetry-alerts@mozilla.com",
             "relud@mozilla.com",
             "jklukas@mozilla.com",
         ],
+    )
+
+    copy_deduplicate_event_ping = bigquery_etl_copy_deduplicate(
```
jklukas, line 100: I see no dependencies on `event_v4` in bigquery-etl (except for views on top of the live table) since the derived event tables are more appropriate for all downstream analysis, so I think we are safe to split out the event ping to its own task without having to make coordinated changes outside the scope of this DAG.

jklukas:
I cleared all these tasks for yesterday to test performance, and looks like it worked great for copy_deduplicate_all and the new copy_deduplicate_event_ping, which both finished in under 12 minutes.

The main ping task, though, unfortunately hit memory errors 40 minutes in (which we've seen before when trying to run a single task). Strangely, the task did not stop at that point, and it reports as still running. See [logs](https://workflow.telemetry.mozilla.org/log?task_id=copy_deduplicate_main_ping&dag_id=copy_deduplicate&execution_date=2021-04-07T01%3A00%3A00%2B00%3A00).

I'm going to go back to slices, but try a smaller number than 100.

jklukas:
> Strangely, the task did not stop at that point, and it reports as still running

I had forgotten that there's retry logic in there now. It's kicked off another query, and that's why it still appears to be running.

PR #1280, opened 2021-04-08
Use 20 slices for copy_deduplicate_main_ping

jklukas:
We are still running into memory errors when running with a single query
after merging https://github.com/mozilla/telemetry-airflow/pull/1279

This moves back to slices, but decreases the number from 100 to 20.


jklukas:
I temporarily suspended privs here to merge and seek retroactive review. I'll be kicking off this job again to validate.

jklukas:
This ran in 49 minutes, so may be performing a bit better than the old 100-slice configuration.

relud (commented):
+1

PR #1281, opened 2021-04-09
Add taar profiles deletion job

eu9ene:
The new TAAR DAG runs the job which does the following:

1. reads client IDs in telemetry deletion requests from BigQuery for the specified period of time (last 29 days)
2. deletes user profiles associated with clients IDs from BigTable

It should be synced with `shredder` job to not repopulate recently deleted user profiles from `telemetry.clients_last_seen` by `taar_weekly` job.

Another option would be to include it in `shredder` DAG, but I think it will be easier to maintain it separately and see who is responsible in case things fail.

The deletion job is added in this PR https://github.com/mozilla/taar_gcp_etl/pull/11, so final testing depends on its approval to upload the docker image.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1651794
TAAR data deletion reference: https://github.com/mozilla/taar#deleting-individual-user-data-from-all-taar-resources


dags/taar_monthly.py
```diff
@@ -0,0 +1,80 @@
...
+        "hwoo@mozilla.com",
+        "epavlov@mozilla.com",
+        "telemetry-alerts@mozilla.com"
+    ],
+    "depends_on_past": False,
+    # should be in sync with shredder DAG
+    "start_date": datetime(2020, 4, 7),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
```
haroldwoo, line 40: consider increasing

dags/taar_monthly.py
```diff
@@ -0,0 +1,80 @@
...
+
+delete_optout = GKEPodOperator(
+    task_id="delete_opt_out_users_from_bigtable",
+    name="delete_opt_out_users_from_bigtable",
+    image=TAAR_ETL_CONTAINER_IMAGE,
+    # Due to the nature of the container run, we set get_logs to False,
+    # To avoid urllib3.exceptions.ProtocolError: 'Connection broken: IncompleteRead(0 bytes read)' errors
+    # Where the pod continues to run, but airflow loses its connection and sets the status to Failed
+    # See: https://github.com/mozilla/telemetry-airflow/issues/844
+    get_logs=False,
```
haroldwoo, line 66: this is only needed when the container runs for long periods of time without any output or updates

haroldwoo (approved):

dags/taar_monthly.py
```diff
@@ -0,0 +1,80 @@
...
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=60),
+}
+
+taar_monthly = DAG(
+    "taar_weekly", default_args=default_args_monthly, schedule_interval=timedelta(days=SCHEDULE_DAYS)
+)
+
+wait_shredder = ExternalTaskSensor(
```
haroldwoo, line 48: i think shredder may move out of airflow in the future

jklukas: And I don't think there's any need to ensure the date ranges are in sync with shredder. The deletion ping tables are updated daily. As long as this DAG is running every 28 days, it doesn't matter where the boundaries are.

eu9ene: Yes, it makes sense. Actually, I have an even better idea: I can just attach this job to taar_weekly right after BigTable is filled and delete opt-out profiles for the last 28 days. Then even if shredder wasn't launched yet and deleted last week profiles were repopulated they will be deleted on the next step. It will allow reducing the delay of deletion after a user opts out. It's just backfilling will be a bit slower. I could also exclude those profiles on BigTable update step, but it overcomplicates things.

relud (approved):

PR #1284, opened 2021-04-16
Bug 1695687 Add mad_server DAG

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1695687 (restricted access).

### DIFF
--- a/dags/mad_server.py
+++ b/dags/mad_server.py
@@ -1,0 +1,40 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from utils.gcp import gke_command
+
+
+default_args = {
+    "owner": "jklukas@mozilla.com",
+    "depends_on_past": False,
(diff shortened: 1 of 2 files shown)


dags/mad_server.py
```diff
@@ -0,0 +1,38 @@
...
+
+with DAG("mad_server", default_args=default_args, schedule_interval="@daily") as dag:
+
+    mad_server_pull = gke_command(
+        task_id="mad_server_pull",
+        command=[
+            "bin/airflow_pull",
+        ],
+        docker_image="gcr.io/malicious-addons-detection/mad-server:latest",
+        gcp_conn_id="google_cloud_airflow_gke",
```
jklukas, line 24: Is using the new GKE cluster the right thing to do for this task?

jasonthomas (approved):

PR #1285, opened 2021-04-19
Specify cluster name for mad_server

jklukas:
Follow-up to #1284 

The DAG is merged and the initial task failed ([logs](https://workflow.telemetry.mozilla.org/log?task_id=mad_server_pull&dag_id=mad_server&execution_date=2021-04-18T00%3A00%3A00%2B00%3A00)) with a message that mentioned 'bq-load-gke-1', so I think the problem is simply that we were using a default argument for cluster_name.


jasonthomas (approved):

jasonthomas:
We may also need to specify `gke_location`.

PR #1286, opened 2021-04-19
Add gke_location to mad_server DAG

jklukas:
Follow-up to #1285


jasonthomas (approved):
LGTM. I tested locally to verify.

PR #1288, opened 2021-04-20
Update owners

jklukas:
Removes some folks who no longer are active on the assigned projects.

Also see https://github.com/mozilla/bigquery-etl/pull/1969

### DIFF
--- a/dags/attitudes_daily.py
+++ b/dags/attitudes_daily.py
@@ -7,9 +7,9 @@ from utils.gcp import bigquery_etl_query, gke_command
 
 
 default_args = {
-    "owner": "ssuh@mozilla.com",
+    "owner": "jklukas@mozilla.com",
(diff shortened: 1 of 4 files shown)


dags/attitudes_daily.py
```diff
@@ -7,9 +7,9 @@
 
 
 default_args = {
-    "owner": "ssuh@mozilla.com",
+    "owner": "jklukas@mozilla.com",
```
jklukas, line 10: I have no idea who would have more context on this, so marking myself as owner.

Reviewers: feel free to drop any knowledge/context here.

BenWu (approved):
This looks reasonable to me.  I wasn't even aware attitudes_daily existed

jklukas:
>  I wasn't even aware attitudes_daily existed

You're not the only one. I think general awareness of this dataset is not high.

PR #1296, opened 2021-04-27
Add staging job for lookml-generator

fbertsch:


dags/probe_scraper.py
```diff
@@ -101,11 +101,11 @@
         python_callable=lambda: time.sleep(60 * 30))
 
     gcp_gke_conn_id = "google_cloud_airflow_gke"
-    lookml_generator = GKEPodOperator(
+    lookml_generator_prod = GKEPodOperator(
         email=["frank@mozilla.com", "dataops+alerts@mozilla.com"],
         task_id="lookml_generator",
         name="lookml-generator-1",
-        image="gcr.io/moz-fx-data-airflow-prod-88e0/lookml-generator:latest",
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/lookml-generator:release",
```
haroldwoo, line 108: you're going to tag the release, 'release'?

haroldwoo: i thought you were going to version them, e.g. v1.0

dags/probe_scraper.py
```diff
@@ -125,4 +125,30 @@
...
     )
 
-    delay_python_task >> lookml_generator
+    delay_python_task >> lookml_generator_prod
+
+    lookml_generator_staging = GKEPodOperator(
+        email=["frank@mozilla.com", "dataops+alerts@mozilla.com"],
+        task_id="lookml_generator_staging",
+        name="lookml-generator-staging-1",
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/lookml-generator:latest",
```
haroldwoo, line 134: staging is using latest, prod is using release?

fbertsch: That's the plan

dags/probe_scraper.py
fbertsch: Hmm, if I version them, then I have to not only tag a release but update this job. Is that what you mean?

jasonthomas: I agree with @haroldwoo we should version and have unique tags per release so that we can compare between release and also rollback to previous release easily if needed.

jasonthomas (approved):

PR #1301, opened 2021-05-07
Fission experiment monitoring: cleanup

akkomar:
This is a follow up to https://github.com/mozilla/telemetry-airflow/pull/1300 where aggregation job was switched to use main_nightly and crash tables.

 Changes:
 * unschedule no longer used intermediate tables
 * make aggregation job dependent on main_nightly


dags/fission_experiment_monitoring.py
```diff
@@ -25,25 +25,17 @@
...
-    wait_for_copy_deduplicate_main_ping = ExternalTaskSensor(
-        task_id="wait_for_copy_deduplicate_main_ping",
-        external_dag_id="copy_deduplicate",
-        external_task_id="copy_deduplicate_main_ping",
-        execution_delta=datetime.timedelta(hours=1),
+    wait_for_main_nightly = ExternalTaskSensor(
+        task_id="wait_for_main_nightly",
+        external_dag_id="bqetl_main_summary",
+        external_task_id="telemetry_derived__main_nightly__v1",
+        execution_delta=datetime.timedelta(hours=2),
```
scholtzan, line 32:
```suggestion
```
`bqetl_main_summary` is also scheduled for `0 2 * * *` so `execution_delta` shouldn't be necessary here. It might however be good to schedule this DAG a little later, maybe `0 3 * * *`, to ensure `bqetl_main_summary` already finished. Otherwise, since both DAGs start at the same time, this one will have to wait.

scholtzan (approved):

dags/fission_experiment_monitoring.py
akkomar: Good idea, thanks!

dags/fission_experiment_monitoring.py
```diff
@@ -22,28 +22,20 @@
...
-        task_id="fission_monitoring_main_v1",
-        project_id="moz-fx-data-shared-prod",
-        destination_table="fission_monitoring_main_v1",
-        dataset_id="telemetry_derived",
-        arguments=('--schema_update_option=ALLOW_FIELD_ADDITION',),
-    )
-
     wait_for_copy_deduplicate_crash_ping = ExternalTaskSensor(
         task_id="wait_for_copy_deduplicate_crash_ping",
         external_dag_id="copy_deduplicate",
```
scholtzan, line 41: The `execution_delta` here would be 2 hours

scholtzan (approved):
👍

PR #1303, opened 2021-05-07
Remove allow_field_addition_on_date support

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/2011

This removes support for `allow_field_addition_on_date`, instead table schemas should be updated via `schema.yaml` files in bigquery-etl. 

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1706683


jklukas (approved):
r+

PR #1307, opened 2021-05-13
Use bqetl CLI for running baseline queries

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/2027

I think this and bigquery-etl are the only places the scripts are used?

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -168,25 +168,41 @@ with models.DAG(
     )
     baseline_args = [
         "--project-id=moz-fx-data-shared-prod",
-        "--date={{ ds }}",
         "--only=*_stable.baseline_v1",
+        "--start_date={{ ds }}",
+        "--end_date={{ ds }}"
     ]
     baseline_clients_first_seen = GKEPodOperator(
         task_id="baseline_clients_first_seen",
         name="baseline-clients-first-seen",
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -168,25 +168,41 @@
...
     baseline_clients_first_seen = GKEPodOperator(
         task_id="baseline_clients_first_seen",
         name="baseline-clients-first-seen",
-        arguments=["script/run_glean_baseline_clients_first_seen"] + baseline_args,
+        command=[
+            "bqetl",
+            "query",
+            "backfill",
+            "*.baseline_clients_first_seen_v1",
+        ] + baseline_args,
```
scholtzan, line 183: This seems to work for https://github.com/mozilla/telemetry-airflow/blob/2407ba20bf5e6fc35fd5ab3d3194d6d7302bc5a1/dags/stripe.py#L20

jklukas:
This looks good. Will hold off on approving until https://github.com/mozilla/bigquery-etl/pull/2027 is settled, in case we need a change to the CLI invocation here.

jklukas (approved):
r+wc

We should be careful in ordering of events when we merge this. I believe we should wait for https://github.com/mozilla/bigquery-etl/pull/2027 to be merged and verify that the `generated-sql` content matches our expectations before merging this one.

PR #1311, opened 2021-05-18
Remove --only parameter from baseline tasks

scholtzan:
`--only` is not a backfill option, it was used in the old scripts


dags/copy_deduplicate.py
```diff
@@ -159,7 +159,6 @@
 
     baseline_args = [
         "--project-id=moz-fx-data-shared-prod",
-        "--only=*_stable.baseline_v1",
```
scholtzan, line 162: I _think_ this can be ignored here?

jklukas (approved):

PR #1321, opened 2021-05-28
[Bug DS-1540] Remove Leanplum Export

scholtzan:
See https://jira.mozilla.com/browse/DS-1540

Leanplum is not being renewed, and all Leanplum keys have been revoked.


fbertsch (approved):
r+wc. Let's also drop https://github.com/mozilla/telemetry-airflow/blob/main/dags/utils/leanplum.py

PR #1322, opened 2021-06-03
Add operational monitoring dag.

emtwo:
Should land after https://github.com/mozilla/bigquery-etl/pull/2054

### DIFF
--- a/dags/operational_monitoring.py
+++ b/dags/operational_monitoring.py
@@ -1,0 +1,51 @@
+from airflow import DAG
+from datetime import timedelta, datetime
+from utils.gcp import gke_command
+
+from airflow.operators.sensors import ExternalTaskSensor
+
+
+default_args = {
+    "owner": "msamuel@mozilla.com",
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "msamuel@mozilla.com",
+    ],
+    "depends_on_past": False,
+    "start_date": datetime(2021, 6, 3),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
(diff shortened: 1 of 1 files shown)


dags/operational_monitoring.py
```diff
@@ -0,0 +1,31 @@
...
+        task_id="run_operational_monitoring",
+        command=[
+            "bqetl",
+            "opmon",
+            "run",
+            "--submission-date={{ ds }}",
+        ],
+        docker_image="mozilla/bigquery-etl:latest",
+        dag=dag,
+    )
```
scholtzan, line 31: This should run after `telemetry_derived__main_nightly__v1` (https://github.com/mozilla/bigquery-etl/blob/main/dags/bqetl_main_summary.py#L282). `telemetry_derived__main_nightly__v1` needs to be specified as an upstream dependency and the `schedule_interval` should be set to `0 3 * * *` (by then the upstream task should have finished)

dags/operational_monitoring.py
```diff
@@ -0,0 +1,51 @@
...
+        "msamuel@mozilla.com",
+    ],
+    "depends_on_past": False,
+    "start_date": datetime(2021, 6, 3),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+PROJECT_ID = "operational_monitoring"
```
scholtzan, line 21: This probably should be something like `DAG_NAME`. `PROJECT_ID` could be confusing here since it doesn't actually refer to a GCP project.

scholtzan (approved):

dags/operational_monitoring.py
emtwo: good point

PR #1328, opened 2021-06-22
Remove Amplitude export tasks

jklukas:
Noticed that these DAGs still exist while doing Airflow triage.

I need to get verification that these use cases are truly spun down.


fbertsch (changes requested):
We need to wait on this until the EOM. My plan was to disable at EOW, but we can keep this around to remove them after the wellness week.

GRAVEYARD.md
```diff
@@ -0,0 +1,7 @@
+# telemetry-airflow Code Graveyard
+
+This document records interesting code that we've deleted for the sake of discoverability for the future.
+
+## 2021-06-22 Amplitude export
```
fbertsch, line 5:
```suggestion
## 2021-07-05 Amplitude export
```

fbertsch (approved):
r+wc

PR #1329, opened 2021-06-22
Schedule partybal

scholtzan:
Fixes https://github.com/mozilla/jetstream/issues/790

Once this is merged we should kill the hala job

### DIFF
--- a/dags/partybal.py
+++ b/dags/partybal.py
@@ -1,0 +1,28 @@
+from airflow import DAG
+from datetime import timedelta, datetime
+from operators.gcp_container_operator import GKEPodOperator
+
+default_args = {
+    "owner": "ascholtz@mozilla.com",
+    "email": ["ascholtz@mozilla.com", "msamuel@mozilla.com",],
+    "depends_on_past": False,
+    "start_date": datetime(2021, 6, 21),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("partybal", default_args=default_args, schedule_interval="0 */3 * * *") as dag:
+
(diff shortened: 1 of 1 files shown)


dags/partybal.py
```diff
@@ -0,0 +1,28 @@
...
+    "email": ["ascholtz@mozilla.com", "msamuel@mozilla.com",],
+    "depends_on_past": False,
+    "start_date": datetime(2021, 6, 21),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+with DAG("partybal", default_args=default_args, schedule_interval="0 */3 * * *") as dag:
```
scholtzan, line 16: Does scheduling every 3 hours seem right?

dags/partybal.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # Built from repo https://github.com/mozilla/partybal
+    partybal_image = "gcr.io/partybal/partybal:latest"
```
scholtzan, line 19: I added the CircleCI config to the partybal repo. So the container has already been pushed

dags/partybal.py
tdsmith: sgtm 🙌 

Right now the docker image defaults to rebuilding everything that changed within the last 25 hours. Usually rebuilding an experiment is not expensive -- it's just time, and that's not critical here, but we could shorten the lookback interval if we wanted.

It does mean you might have to run partybal manually if it missed an update.

tdsmith (approved):
I don't feel qualified to review Airflow changes but this looks good to me.

dags/partybal.py
scholtzan: I guess the schedule can be easily adjusted if we see a need

PR #1332, opened 2021-07-06
Bug 1716802 - Dont dryrun mozagg trim db

fbertsch:


acmiyaguchi (approved):
Add a reference to the bug number in the squash (https://bugzilla.mozilla.org/show_bug.cgi?id=1716802), otherwise looks good to me.

PR #1333, opened 2021-07-08
Bug 1719188 - Set max_active_runs to 1

acmiyaguchi:
I'm backfilling [bug 1719188](https://bugzilla.mozilla.org/show_bug.cgi?id=1719188). These jobs are set up so there is no depends on past, but it is problematic if they run concurrently. This should allow for clearing of past tasks and allowing airflow to schedule the jobs appropriately.


dags/glam_fenix.py
```diff
@@ -18,6 +18,7 @@
     "email_on_retry": True,
     "retries": 1,
     "retry_delay": timedelta(minutes=30),
+    "max_active_runs": 1
```
scholtzan, line 21: I believe this needs to be passed as [parameter into `DAG(... , max_active_runs=1, ...)`](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dag/index.html#airflow.models.dag.DAG). See also: https://github.com/mozilla/telemetry-airflow/blob/89724725eef46d2a7f9c75f93d8af1da9eb7e990/dags/experiments_live.py#L22

scholtzan (approved):

PR #1340, opened 2021-07-23
Bug 1719188 - Add individual tasks for glam-fenix dag

acmiyaguchi:
[bug 1719188](https://bugzilla.mozilla.org/show_bug.cgi?id=1719188)

This sets the dependencies between all of the tasks in the dag. All of the tasks are in the same DAG, which ends up being quite a few. 

This is straight port of the [`run_glam_sql`](https://github.com/mozilla/bigquery-etl/blob/main/script/glam/run_glam_sql) script, and uses the functions directly to run the job.

image

The primary role of breaking all the tasks out like this is to be able to retry on the `clients_{scalar|histogram}_aggregates` job exactly once, since now scalar/histograms run in parallel and can cause issues.

See https://github.com/mozilla/bigquery-etl/pull/2208


dags/glam_fenix.py
```diff
@@ -84,17 +88,63 @@
...
     for dependency in LOGICAL_MAPPING.get(product, [product]):
-        mapping[dependency] >> query
+        mapping[dependency] >> cdsa
+        mapping[dependency] >> cdha
+
+    csa_init = init(task_name=f"{product}__clients_scalar_aggregates_v1")
+    csa = query(task_name=f"{product}__clients_scalar_aggregates_v1")
+
+    cdsa >> csa_init
+    lv >> csa_init
```
BenWu, line 120: If `cdsa >> lv` and `lv >> csa_init`, do we need `cdsa >> csa_init`?

dags/glam_fenix.py
```diff
@@ -84,17 +88,63 @@
...
+    csa_init >> csa
+
+    cha_init = init(task_name=f"{product}__clients_histogram_aggregates_v1")
+    cha = query(task_name=f"{product}__clients_histogram_aggregates_v1")
+
+    cdha >> cha_init
```
BenWu, line 126: Same comment as with `cdsa >> csa_init`

BenWu (approved):
The only thing I have to say is I think the acronym variable names are a little hard to follow at first and I would prefer long names.  Not a big deal though and feel free to disagree

dags/glam_fenix.py
acmiyaguchi: Technically no since the dependency is transitive, but its good practice to lay out the dependencies here. Airflow can run `lv` and `csa_init` at the same time here, instead of one after the other.

acmiyaguchi:
I'll see what this looks like with the longer names; the names are kind of excessively long though.

acmiyaguchi:
@BenWu Updated the tasks to use full names instead of acronyms, it made sense to split task declarations from dependencies because of length.

BenWu (approved):
😎

PR #1343, opened 2021-08-14
Update prio-processor v4 for minio gcs proxy

acmiyaguchi:
This makes a few changes to the way that containers are run in Kubernetes pods so the main container can access a minio proxy. The reason that this change was made is to accommodate https://github.com/mozilla/prio-processor/pull/117.

Normally when you run pods in a kubernetes cluster and you want them to talk to each other, you would [configure a service](https://kubernetes.io/docs/concepts/services-networking/service/). This isn't possible with the two operators that are available to us in Airflow (GKECreateClusterOperator and GKEPodOperator). Instead we use the `pod_mutation_hook` to modify the pod definition before it's sent to the cluster to run:

1. Allow containers to share the same process namespace and ports. Doing this allows one container to see the processes running in another container.
2. Add minio as a side-car container (available to other containers running on the same node)
3. Add third "reaper" container that watches the process namespace for a certain program name, then kill the minio process.
4. Run the main script and execute `exec -a <name> sleep 10` to notify the reaper container.

The reason for the complexity here is because the minio container will continue to run even after the main container finishes. This seems to work during testing, and I'm comfortable moving this to the actual airflow cluster.

There are a few things that may need to be done to mitigate failures:

1. A cluster may fail to start. In this case, clear the corresponding stop command.
2. A container may fail to shutdown. This can killed manually in the UI and the cluster shutdown.


acmiyaguchi:
This depends on https://mozilla-hub.atlassian.net/browse/DSRE-130 to run successfully.

dags/utils/gke.py
```diff
@@ -3,7 +3,7 @@ def create_gke_config(
     service_account,
     owner_label,
     team_label,
-    machine_type="n1-standard-1",
+    machine_type="n1-standard-2",
```
wlach, line 6: Did you mean to change these settings?

wlach (commented):
These changes look pretty reasonable to me in outline, though I don't know if I have enough experience with this type of Airflow job to be fully confident I'm not missing something. I think it would be a good idea to have someone from SRE take a look at this before landing.

acmiyaguchi:
Pulling in Harold to give the PR a look too.

dags/utils/gke.py
acmiyaguchi: Yes, there are now two containers running on the same node, so I want each of them to get their own core.

config/airflow_local_settings.py
```diff
@@ -1,11 +1,47 @@
...
+    "failed": "red",
+    "up_for_retry": "gold",
+    "up_for_reschedule": "turquoise",
+    "upstream_failed": "orange",
+    "skipped": "pink",
+    "scheduled": "tan",
 }
+
+
+def pod_mutation_hook(pod: V1Pod):
```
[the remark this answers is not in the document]
haroldwoo: I suspect because the upstream code lives outside the repo, and this is how you got things to work. If so I would normally say to import the upstream code, and modify the bits you want to change. But all of the k8s code for 2.0+ has gone under a lot of changes that I haven't been able to review yet, and a lot of work here will likely break when I upgrade.

I don't like this logic being put at a global place, which may require restarts/redeploys every time you need to modify something.

haroldwoo:
You may want to rethink this entire approach. From the airflow 2 docs (https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html):
--
Difference between KubernetesPodOperator and Kubernetes object spec
The KubernetesPodOperator can be considered a substitute for a Kubernetes object spec definition that is able to be run in the Airflow scheduler in the DAG context. If using the operator, there is no need to create the equivalent YAML/JSON object spec for the Pod you would like to run. The YAML file can still be provided with the pod_template_file or even the Pod Spec constructed in Python via the full_pod_spec parameter which requires a Kubernetes V1Pod.
--
I don't have the context of what exactly you're doing with minio/etc. but it may be easier to reason about a k8s pod operator where you fully define the pod spec, to include your sidecar or init container. This may remove your requirement to share process namespaces for your reaper functionality. You may also be able to rely on a large pod timeouts, or shared volume mount with semaphore file to remove your requirement of sharing a process namespace.

I would also suggest considering not trying to fit this whole workflow within airflow. Having a small long lived gke cluster will allow you to run a service, and you can use k8s cronjobs to perform batch work.

config/airflow_local_settings.py
```diff
@@ -1,11 +1,47 @@
...
+        minio_container.image = "minio/minio:RELEASE.2021-06-17T00-10-46Z"
+        minio_container.args = ["gateway", "gcs", "$(PROJECT_ID)"]
+        minio_container.name = "minio"
+        pod.spec.containers.append(minio_container)
+
+        # Search for a new process named `minio-done` and kill the minio
+        # container above. This can be done using `exec -a minio-done sleep 10`
+        # which will will create a process available in the shared namespace for
+        # 10 seconds.
+        pkill_container = deepcopy(pod.spec.containers[0])
```
haroldwoo, line 39: you are using this copy, modify, append logic more than once, it may justify writing a helper method (or use the kubernetes libraries to construct an object from scratch. There may be some metadata or other extra things that have implications

dags/prio/kubernetes.py
```diff
@@ -133,10 +129,19 @@ def container_subdag(
...
             ],
+        )
+
+        run_prio = GKEPodOperator(
+            task_id=f"processor_{server_id}",
+            name=f"processor_{server_id}",
+            cluster_name=cluster_name,
+            image=image,
+            arguments=arguments,
+            env_vars=dict(**env_vars, PYTHONUNBUFFERED="0"),
```
haroldwoo, line 140: ah the smart way

dags/prio_processor.py
```diff
@@ -67,6 +75,21 @@
     BUCKET_PRIVATE_B,
     "{{ var.value.prio_public_key_hex_internal }}",
     "{{ var.value.prio_public_key_hex_external }}",
+    env_vars={
+        # required for minio in the pod_mutation_hook
+        "PROJECT_ID": PROJECT_ADMIN,
+        "MINIO_ROOT_USER": username,
+        "MINIO_ROOT_PASSWORD": password,
```
haroldwoo, line 82: if you do go dedicated gke cluster route, a lot of these can be injected via secret to the containers

haroldwoo (commented):
Due to the complexity, we may want to meet to discuss other options. I'll also have a better idea of what the k8s landscape will look like in airflow2 in a few weeks

config/airflow_local_settings.py
```diff
@@ -1,11 +1,47 @@
...
-    "queued": 'gray',
-    "running": 'lime',
-    "success": '#0000FF',
-    "failed": 'red',
-    "up_for_retry": 'gold',
-    "up_for_reschedule": 'turquoise',
-    "upstream_failed": 'orange',
-    "skipped": 'pink',
-    "scheduled": 'tan',
+    "queued": "gray",
```
haroldwoo, line 7: is this some linting

acmiyaguchi:
I tried using the `pod_template_file` inside of the podoperator, but this functionality doesn't work in 1.10.15 (I tried it in [this commit](https://github.com/mozilla/telemetry-airflow/pull/1343/commits/70e5075e68272ced940d0d64931c3e3cd98817fb), see this issue https://github.com/apache/airflow/pull/6230#issuecomment-597176715 where the first version for this to hit is 2.x). 

Having a long lived kubernetes cluster with a minio service would be the easiest solution, using the podoperator in the intended usage.

config/airflow_local_settings.py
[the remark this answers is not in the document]
acmiyaguchi: This is how the `pod_mutation_hook` is supposed to be defined, as per the airflow documentation: https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html#pod-mutation-hook

config/airflow_local_settings.py
```diff
@@ -1,11 +1,47 @@
...
+
+    # check that we're running a prio-processor job, and spin up a side-car
+    # container for proxying gcs buckets.
+    if pod.metadata.labels["job-kind"] == "prio-processor":
+        pod.spec.share_process_namespace = True
+        # there is only one container within the pod, so lets append a few more
+
+        # for proxying gcs resource consistently across platforms
+        minio_container = deepcopy(pod.spec.containers[0])
+        minio_container.image = "minio/minio:RELEASE.2021-06-17T00-10-46Z"
```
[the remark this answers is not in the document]
acmiyaguchi: Probably, but the minio container is pinned because nince newer versions break behavior with the default port on the docker-compose version of the integration tests.

config/airflow_local_settings.py
acmiyaguchi: Yes, this is from black.

config/airflow_local_settings.py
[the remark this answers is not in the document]
haroldwoo: Ah, unfortunate. This effects all pods

PR #1347, opened 2021-08-19
Add note to operational_monitoring DAG

scholtzan:
The `operational_monitoring` DAG is currently still under heavy development and frequently failing. To prevent constantly bugs being filed during Airflow triage, I added a note to the DAG description that failures can be ignored for now. Once it's production ready we can update the docs there.

### DIFF
--- a/dags/operational_monitoring.py
+++ b/dags/operational_monitoring.py
@@ -5,6 +5,21 @@ from utils.gcp import gke_command
 from operators.task_sensor import ExternalTaskCompletedSensor
 
 
+docs = """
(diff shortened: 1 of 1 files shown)


emtwo:
Thanks Anna!

emtwo (approved):
lgtm but I'm not sure why CI is failing so pending that it's good

scholtzan:
I think the build failure is unrelated to this change. 
@haroldwoo do you know why build is failing? It looks like this is happening also for other CI build runs.

PR #1352, opened 2021-09-02
Add Fivetran operator, sensor, hook backport

scholtzan:
Backporting the Airflow Fivetran operators so we can use them in Airflow 1.10.15.
The code was basically just copied from https://github.com/fivetran/airflow-provider-fivetran

There is also an example DAG which I removed: https://github.com/mozilla/telemetry-airflow/pull/1352/commits/857a671701212dbf07704c19d70d2543484954b9


dags/operators/backport/fivetran/tests/test_hooks.py
```diff
@@ -0,0 +1,94 @@
+# Ported from https://github.com/fivetran/airflow-provider-fivetran/blob/main/tests/hooks/test_hooks.py
+"""
+Unittest module to test Fivetran Hooks.
+Requires the unittest, pytest, and requests-mock Python libraries.
+Run test:
+    python3 -m unittest operators.backport.fivetran.tests.test_hooks.TestFivetranHook
```
scholtzan, line 6: I manually executed these tests to see if everything is working.

haroldwoo (approved):
I would omit the tests dir

PR #1354, opened 2021-09-09
add sample counts to glam desktop dag

alekhyamoz:
The PR is to resolve the issue [#1240](https://github.com/mozilla/glam/issues/1240). 
As part of this change, I have added two tasks 
1. glam_sample_counts: This task aggregates (sums) on the histogram values [query](https://gist.github.com/alekhyamoz/031fd954469acd6020d844202c5f8ce4)
2. extract_sample_counts: This task extracts the data from bigquery table and pushed the data to GCS in csv format

The below is how the dag looks including the above two tasks:
Screen Shot 2021-09-09 at 11 43 36 AM

### DIFF
--- a/dags/glam_subdags/extract.py
+++ b/dags/glam_subdags/extract.py
@@ -160,3 +160,63 @@ def extract_user_counts(
     etl_query >> gcs_delete >> bq2gcs
 
     return dag
+
+def extract_sample_counts(
+    parent_dag_name,
+    child_dag_name,
+    default_args,
+    schedule_interval,
+    dataset_id
+):
+
+    dag = DAG(
+        dag_id="{}.{}".format(parent_dag_name, child_dag_name),
+        default_args=default_args,
+        schedule_interval=schedule_interval,
+    )
+
+    bq_extract_table = "glam_sample_counts_extract_v1"
+    etl_query = bigquery_etl_query(
+        task_id="glam_sample_counts_extract",
+        destination_table=bq_extract_table,
+        dataset_id=dataset_id,
+        project_id=project_id,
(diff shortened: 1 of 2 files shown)


dags/glam_subdags/extract.py
```diff
@@ -160,3 +160,63 @@ def extract_user_counts(
     etl_query >> gcs_delete >> bq2gcs
 
     return dag
+
+def extract_sample_counts(
```
emtwo, line 164: Since `extract_sample_counts()` is very similar to `extract_user_counts()`, can we refactor the function so that the same one can be used and just pass in the required params for each call?

dags/glam_subdags/extract.py
```diff
@@ -135,18 +136,18 @@ def extract_user_counts(
...
     )
 
-    gcs_destination = "gs://{}/glam-extract-firefox-counts.csv".format(
-        glam_bucket
+    gcs_destination = "gs://{}/glam-extract-firefox-{}.csv".format(
+        glam_bucket, file_prefix
     )
     bq2gcs = BigQueryToCloudStorageOperator(
-        task_id="glam_extract_user_counts_to_csv",
+        task_id="glam_extract_user_{}_to_csv".format(task_prefix),
```
emtwo, line 150: I think you can remove `user` from here? The `task_prefix` already contains `user`

emtwo (approved):
looks good! Just the one extra small change I mentioned below and it looks ready to go

dags/glam_subdags/extract.py
alekhyamoz: Yes I will remove it. Thank you Marina!

PR #1364, opened 2021-09-27
Bug 1729524 Add adm_export weekly DAG

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1729524

Uses new docker image from https://github.com/mozilla/docker-etl/pull/28

Depends on https://github.com/mozilla/bigquery-etl/pull/2373

### DIFF
--- a/dags/adm_export.py
+++ b/dags/adm_export.py
@@ -1,7 +1,8 @@
 import datetime
 
-from airflow import models
+from airflow import DAG
 from airflow.hooks.base_hook import BaseHook
+from airflow.contrib.hooks.gcp_api_base_hook import GoogleCloudBaseHook
 
 from operators.gcp_container_operator import GKEPodOperator
 from operators.task_sensor import ExternalTaskCompletedSensor
@@ -29,7 +30,7 @@ default_args = {
 
 dag_name = "adm_export"
 
-with models.DAG(
+with DAG(
     dag_name, schedule_interval="0 5 * * *", doc_md=DOCS, default_args=default_args
 ) as dag:
 
@@ -40,7 +41,7 @@ with models.DAG(
         name="adm_weekly_aggregates_to_sftp",
         # See https://github.com/mozilla/docker-etl/pull/28
         image="gcr.io/moz-fx-data-airflow-prod-88e0/bq2sftp_docker_etl:latest",
-        project_id='moz-fx-data-shared-prod',
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_gke_conn_id).project_id,
         gcp_conn_id="google_cloud_airflow_gke",
         cluster_name="workloads-prod-v1",
         location="us-west1",
--- a/dags/adm_export.py
+++ b/dags/adm_export.py
@@ -1,9 +1,11 @@
 import datetime
 
 from airflow import models
+from airflow.hooks.base_hook import BaseHook
 
 from operators.gcp_container_operator import GKEPodOperator
-from airflow.hooks.base_hook import BaseHook
+from operators.task_sensor import ExternalTaskCompletedSensor
+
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -52,6 +52,7 @@ services:
       - AIRFLOW__KUBERNETES__IN_CLUSTER=False
       - URL=http://localhost:8000
       - WEBSERVER_USE_RBAC=False
+      - 'AIRFLOW_CONN_ADM_SFTP=ftp://myname:mypassword@myhost.com:8000?known_hosts=myhost.com+AAAABBBBB'
```
jklukas, line 55: I'm adding dummy connection details for the "adm_sftp" connection as env var here to allow the dag parsing test to run successfully. This appears to be working, and may be quicker to load than using the CLI to add the connection in `bin/run`. AFAICT, there's overhead to every airflow CLI call that's dominating the startup time.

dags/adm_export.py
```diff
@@ -0,0 +1,72 @@
...
+) as dag:
+
+    conn = BaseHook.get_connection('adm_sftp')
+
+    adm_weekly_aggregates_to_sftp = GKEPodOperator(
+        task_id="adm_weekly_aggregates_to_sftp",
+        name="adm_weekly_aggregates_to_sftp",
+        # See https://github.com/mozilla/docker-etl/pull/28
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/bq2sftp_docker_etl:latest",
+        project_id='moz-fx-data-shared-prod',
```
haroldwoo, line 43: I think the project id and the gcp conn id need to correspond, else the operator will think the workloads-prod-v1 gke cluster exists in shared-prod.

docker-compose.yml
haroldwoo: This was an option i considered as well. It becomes a little more complex with gcp/aws type connections, and we can use a .env file to store all of them as well. 

https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html#storing-a-connection-in-environment-variables

Why is the whole k:v in single quotes? Shouldn't it be k='v'?

haroldwoo (approved):
lgtm. had some comments

docker-compose.yml
jklukas: From the yaml document's perspective, this is a list of strings. It's docker-compose that is expecting to parse these items as `k=v`. If we wrote `k='v'` here, then the value would literally contain the quotes.

That said, I don't think there are any characters here that are problematic, so we don't need any quotes at all.

dags/adm_export.py
```diff
@@ -0,0 +1,73 @@
...
+) as dag:
+
+    conn = BaseHook.get_connection('adm_sftp')
+
+    adm_weekly_aggregates_to_sftp = GKEPodOperator(
+        task_id="adm_weekly_aggregates_to_sftp",
+        name="adm_weekly_aggregates_to_sftp",
+        # See https://github.com/mozilla/docker-etl/pull/28
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/bq2sftp_docker_etl:latest",
+        project_id=GoogleCloudBaseHook(gcp_conn_id=gcp_gke_conn_id).project_id,
```
haroldwoo, line 44: i dont see gcp_gke_conn_id defined?

haroldwoo: if this needs to be revised, can you just hardcode it to 'moz-fx-data-airflow-gke-prod', since 2.0 will need it this way

jklukas: You're right. I copy/pasted this. I had set auto-merge and I was expecting this to not merge if the test-environment task failed. Apologies for the churn. I'm working on a follow-up immediately.

PR #1368, opened 2021-10-06
Schedule metrics queries

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/2409

Before merging we'll also have to create the tables and at some point backfill them.


scholtzan:
All tables got deployed. I'll trigger a test run once the DAG changes got deployed.

PR #1382, opened 2021-10-15
Burnham multi version config

jrbenny35:
This PR adds the configuration necessary to run a Burnham dag against multiple versions of Burnham. It adds an additional client based on the original clients. These new clients will pull the latest Major version of burnham, as specified within the DAG.

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -9,6 +9,8 @@ import logging
 import uuid
 import time
 
+from dataclasses import dataclass
+
 from airflow import DAG
 from airflow.operators.python import PythonOperator
 from operators.bq_sensor import BigQuerySQLSensorOperator
@@ -27,7 +29,7 @@ See https://github.com/mozilla/burnham
 """
 
 DAG_OWNER = "rpierzina@mozilla.com"
-DAG_EMAIL = ["glean-team@mozilla.com", "rpierzina@mozilla.com"]
+DAG_EMAIL = ["glean-team@mozilla.com", "rpierzina@mozilla.com", "bforehand@mozilla.com"]
 
 PROJECT_ID = "moz-fx-data-shared-prod"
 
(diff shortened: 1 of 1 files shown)


hackebrot (changes requested):
As discussed on Zoom today, a few of the expected query results need to be updated. Clearing this review request for now.

jrbenny35:
These commits are kinda weird but we can fix with a squash

dags/burnham.py
```diff
@@ -178,9 +180,9 @@
 """
 
 WANT_TEST_EXPERIMENTS = [
-    {"experiment": "no_experiments", "document_count": 2},
-    {"experiment": "spore_drive:tardigrade", "document_count": 2},
-    {"experiment": "spore_drive:tardigrade-dna", "document_count": 6},
+    {"experiment": "no_experiments", "document_count": 4},
+    {"experiment": "spore_drive:tardigrade", "document_count": 4},
```
hackebrot, line 184: From what I can tell we should expect `1` additional `discovery` ping, see https://github.com/mozilla/telemetry-airflow/pull/1382#pullrequestreview-784554475.

```suggestion
    {"experiment": "spore_drive:tardigrade", "document_count": 3},
```

hackebrot (changes requested):

dags/burnham.py
jrbenny35: Whoops, I had that but maybe it got lost when I had to do a rebase/merge. Just going to add it via github.

hackebrot (approved):
LGTM! 🚀

haroldwoo:
I would squash the commits so the log doesnt show 17 for this PR. Aftwards, feel free to hit the 'rebase and merge' button

jrbenny35:
I can't merge as I do not have write access.

haroldwoo:
Granted you access

PR #1391, opened 2021-10-20
Update notification email list of GLAM dags

Iinh:


kik-kik (approved):

kik-kik:
The change itself looks good @Iinh , but I can see there are some conflicts that need to be resolved before merging.

kik-kik (commented):
@Iinh Actually, just noticed. Is there a specific reason why `owner` and `email` fields are not just DAG level params instead of passing them to each task individually? That would help avoid duplication of those values.

If you'd like to keep it as it now the only suggestion I'd also have would be to define those values at the top of the module as constants and just reference them when passing them into a task. This way we can avoid having to make multiple changes in the DAG in case of email or owner changes.

Iinh:
> Actually, just noticed. Is there a specific reason why owner and email fields are not just DAG level params instead of passing them to each task individually? That would help avoid duplication of those values.

Hi @alekhyamoz, @kik-kik brought up a good point here. Could you elaborate more on why we decided to add these emails at the task level? Thanks.

alekhyamoz:
> @Iinh Actually, just noticed. Is there a specific reason why `owner` and `email` fields are not just DAG level params instead of passing them to each task individually? That would help avoid duplication of those values.
> 
> If you'd like to keep it as it now the only suggestion I'd also have would be to define those values at the top of the module as constants and just reference them when passing them into a task. This way we can avoid having to make multiple changes in the DAG in case of email or owner changes.

It's my understanding that whoever worked on this subdag might have anticipated tasks to be added (given this is subdag) which could have totally different owners.

Iinh:
> It's my understanding that whoever worked on this subdag might have anticipated tasks to be added (given this is subdag) which could have totally different owners.

That made sense, considering at one point multiple people were working on different tasks in the GLAM etl. 

I applied @kik-kik's suggestion and turned the email list to a constant to avoid having to make multiple changes in the future when/if the owners change again. If this looks good to you @alekhyamoz please approve the PR so we can merge. Thanks :)

alekhyamoz (approved):
looks good

PR #1413, opened 2021-11-01
Bug 1738045 Add clients_last_seen dep to taar_daily

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1738045


eu9ene (approved):
thank you for the PR!

PR #1416, opened 2021-11-10
[DSRE-205] Migrate Glam gke cronjob to airflow for better dependency timing

haroldwoo:
TODO
- [ ] After merging this PR, "Suspend" the Glam GKE cronjob and wait for these tasks to run via airflow. Ask glam dev team to verify.
- [ ] Once verified, manually remove GKE cronjob from stage/prod glam environments because simply removing the k8s yaml from the repo and redeploying via jenkins won't do so.

### DIFF
--- a/dags/glam.py
+++ b/dags/glam.py
@@ -11,6 +11,7 @@ in telemetry-airflow.
 from datetime import datetime, timedelta
 
 from airflow import DAG
+from operators.gcp_container_operator import GKEPodOperator
 from operators.task_sensor import ExternalTaskCompletedSensor
 from airflow.operators.subdag_operator import SubDagOperator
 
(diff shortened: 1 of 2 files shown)


robhudson (approved):
The commands look correct to me.

alekhyamoz (approved):
lgtm

dags/glam.py
```diff
@@ -353,3 +394,9 @@
 client_scalar_probe_counts >> extracts_per_channel
 scalar_percentiles >> extracts_per_channel
 histogram_percentiles >> extracts_per_channel
+
+extracts_per_channel >> glam_import_desktop_aggs_beta
+extracts_per_channel >> glam_import_desktop_aggs_nightly
+extracts_per_channel >> glam_import_desktop_aggs_release
+extracts_per_channel >> glam_import_user_counts
+extracts_per_channel >> glam_import_probes
```
alekhyamoz, line 402: @robhudson , Just wanted to check, is it fine if all the import jobs run parallel?
Currently it looks like it runs sequentially.

Iinh (approved):

dags/glam.py
robhudson: Good catch. Parallel would be better actually. Some can take quite a while and I believe they were parallel previously via the crons.

PR #1418, opened 2021-11-16
[DSRE-205] Add GKENatPodOperator, and modify glam dags use this

haroldwoo:
For import job that requires whitelist into postgres db


haroldwoo:
yolo

robhudson (approved):

PR #1419, opened 2021-11-18
Road-134 - Add the quicksuggest2bq DAG

Dexterp37:
This is the last bit required to complete [this ticket](https://mozilla-hub.atlassian.net/browse/ROAD-134).

This schedules a daily job that runs quicksuggest2bq, introduced by mozilla/docker-etl#29.


dags/quicksuggest2bq.py
```diff
@@ -0,0 +1,45 @@
...
+    "depends_on_past": False,
+    "start_date": datetime(2021, 11, 18),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+project_id = "TODO"
+table_id = "TODO"
```
Dexterp37, line 25: @jklukas what should we put here?

dags/quicksuggest2bq.py
```diff
@@ -0,0 +1,45 @@
...
+
+    quicksuggest2bq = gke_command(
+        task_id="quicksuggest2bq",
+        command=[
+            "python", "quicksuggest2bq/main.py",
+            "--destination-project", project_id,
+            "--destination-table-id", table_id,
+        ],
+        docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/quicksuggest2bq:latest",
+        gcp_conn_id="google_cloud_derived_datasets",
```
Dexterp37, line 40: @jklukas I cargo culted these from the play-store-export, are these correct for this job?

jklukas: Yes, this all looks fine.

dags/quicksuggest2bq.py
jklukas: Let's make this `moz-fx-data-shared-prod` and `search_terms_derived.remotesettings_suggestions_v1`.

dags/quicksuggest2bq.py
```diff
@@ -0,0 +1,45 @@
+"""
```
jklukas, line 1: How about we name this DAG a bit more generically? Perhaps `contextual_services_import`? There could be other import tasks we would want to run on this schedule.

dags/quicksuggest2bq.py
```diff
@@ -0,0 +1,45 @@
...
+Runs a Docker image that imports Quicksuggest suggestions
```
jklukas, line 2: Looks like this old version of the file still exists.

jklukas (changes requested):
Need to remove the old DAG.

dags/quicksuggest2bq.py
Dexterp37: Whoops.

jklukas (approved):

PR #1421, opened 2021-11-19
Attempted fix for burnham airflow errors.

jrbenny35:
Attempting a fix for the Burnham errors on Airflow. We forgot a mission and I raised the query limits to see if that helps.

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -143,7 +143,7 @@ TEST_EXPERIMENTS = f"""WITH {DISCOVERY_V1_DEDUPED},
   WHERE
     metrics.string.test_name = "{DEFAULT_TEST_NAME}"
   LIMIT
-    20 ),
+    40 ),
   experiment_counts AS (
   SELECT
     experiment,
@@ -202,7 +202,7 @@ WHERE
 ORDER BY
   metrics.string.mission_identifier
 LIMIT
-  20
+  40
 """
 
 WANT_TEST_GLEAN_ERROR_INVALID_OVERFLOW = [
@@ -337,6 +337,8 @@ ORDER BY
 
 WANT_TEST_DELETION_REQUEST_PING_CLIENT_ID = [
     {"mission_identifier": "MISSION B: TWO WARPS"},
+    {"mission_identifier": "MISSION B: TWO WARPS"},
+    {"mission_identifier": "MISSION C: ONE JUMP"},
     {"mission_identifier": "MISSION C: ONE JUMP"},
 ]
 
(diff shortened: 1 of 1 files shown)


dags/burnham.py
```diff
@@ -143,7 +143,7 @@
   WHERE
     metrics.string.test_name = "{DEFAULT_TEST_NAME}"
   LIMIT
-    20 ),
+    40 ),
```
jklukas, line 146: @hackebrot - Do you remember if there is a specific reason for the LIMIT clauses used here? If it's only there to reduce the potential for accidentally returning a large result, I think the likelihood of that causing issues is low enough that we can remove the limit clauses.

jklukas (approved):
These look like reasonable changes to me, but I've not gone back and looked carefully over the missions to verify these are the values we should expect.

dags/burnham.py
hackebrot: I believe we've only added the `LIMIT` clauses because it's best practice and we knew that we wouldn't be seeing more pings for a given test run. I'd be happy to increase that to something higher or if that's your recommendation remove that entirely.

jklukas: I'd be happy for us to remove it entirely. The danger of lost time due to debugging unexpected behavior seems bigger than any query cost or efficiency gain here.

hackebrot:
I won't be able to review this before Monday next week since it's EOD for me now.

Here are a few thoughts that I wanted to document before I log off.

For `2021-11-18T00` there were 3 Airflow tasks that failed:
1. `verify_discovery_data`
2. `verify_space_ship_ready`
3. `verify_deletion_request_data`

For `2021-11-17T00` there were also 3 Airflow tasks that failed, but different ones:
1. `verify_discovery_data`
2. `verify_deletion_request_data`
3. `verify_discovery_data_disable_upload`

For `2021-11-16T00` there were only 2 Airflow tasks that failed:
1. `verify_discovery_data`
2. `verify_deletion_request_data`

I did not have bandwidth to investigate in detail, but it seems that it's most likely not only the expected query results which are not matching the actual results. @jklukas, I've mentioned to @jrbenny35 that maybe the current `LIMIT` clauses mean that we don't see all of the documents for a given test scenario in a test run or that maybe the Airflow Sensors need to be adjusted to expect more documents given that there are now additional clients that send additional pings. What do you think about that? 🤔

jklukas:
> maybe the Airflow Sensors need to be adjusted to expect more documents given that there are now additional clients that send additional pings. What do you think about that?

The 20 minute wait should be sufficient to ensure data has had a chance to flow through, but this is definitely a possibility. It would be worth looking through the logic and validating that the min ping counts there match the number of pings we expect. If we're setting it too low, we do run the risk of proceeding when only partial data has been loaded.

jrbenny35:
I removed all limit statements and updated the query I saw that was wrong. I think this is good to merge

PR #1422, opened 2021-11-22
added the dag for firefox_desktop

alekhyamoz:
Add firefox desktop dag for glam.

The below is screenshot for the dev version: 
Screen Shot 2021-11-22 at 12 15 12 PM

### DIFF
--- a/dags/glam_firefox_desktop.py
+++ b/dags/glam_firefox_desktop.py
@@ -1,0 +1,190 @@
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from operators.task_sensor import ExternalTaskCompletedSensor
+from glam_subdags.generate_query import (
+    generate_and_run_glean_queries,
+    generate_and_run_glean_task,
+)
+from utils.gcp import gke_command
+from functools import partial
+
+default_args = {
+    "owner": "akommasani@mozilla.com",
+    "depends_on_past": False,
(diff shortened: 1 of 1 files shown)


dags/glam_firefox_desktop.py
```diff
@@ -0,0 +1,190 @@
...
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
+}
+
+PROJECT = "moz-fx-data-glam-prod-fca7"
+BUCKET = "moz-fx-data-glam-prod-fca7-etl-data"
+
+# Fenix as a product has a convoluted app_id history. The comments note the
```
emtwo, line 29: I think this comment doesn't apply.

emtwo (commented):
Similar to my comment on the [bqetl pr](https://github.com/mozilla/bigquery-etl/pull/2485#discussion_r754523042), I think there is a lot of duplicated code here that we should try to factor out so that future glean products can use the same code as well.

Can we factor out the duplicated code into a separate file that both desktop and fenix can use?

emtwo (approved):

dags/glam_firefox_desktop.py
alekhyamoz: That makes sense. I will go ahead and merge the PR while working on the changes needed for this comment

PR #1423, opened 2021-11-23
More changes to fix burnham dag errors.

jrbenny35:
Doubling the sleep timer. This may be a bit much but I don't think it will hurt since we are effectively doubling the amount of pings sent.

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -703,10 +703,10 @@ with DAG(
 
     client8.set_upstream(generate_burnham_test_run_uuid)
 
-    # We expect up to 40 minute latency for pings to get loaded to live tables
-    # in BigQuery, so we have a task that explicitly sleeps for 40 minutes
+    # We expect up to 20 minute latency for pings to get loaded to live tables
+    # in BigQuery, so we have a task that explicitly sleeps for 20 minutes
     # and make that a dependency for our tasks that need to read the BQ data.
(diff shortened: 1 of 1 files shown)


jklukas (approved):
> since we are effectively doubling the amount of pings sent

The number of pings we sent shouldn't have any bearing on the latency for loading, since burnham pings are only a small fraction of the load handled by the shared telemetry pipeline infrastructure. I don't expect that the 20-minute wait is a limiting factor here, and I worry that the longer we make that timer the harder it becomes to do iterative testing as we make changes. I think it's okay to try this for now to confirm whether it helps, but I'd like to make sure we reduce this back down to 20 minutes if we determine other mitigations were effective to get this all working.

jrbenny35:
> > since we are effectively doubling the amount of pings sent
> 
> The number of pings we sent shouldn't have any bearing on the latency for loading, since burnham pings are only a small fraction of the load handled by the shared telemetry pipeline infrastructure. I don't expect that the 20-minute wait is a limiting factor here, and I worry that the longer we make that timer the harder it becomes to do iterative testing as we make changes. I think it's okay to try this for now to confirm whether it helps, but I'd like to make sure we reduce this back down to 20 minutes if we determine other mitigations were effective to get this all working.

I'm going to try something else before merging this

jrbenny35:
I increased the number of rows needed for the sensor. Didn't quite understand this before but I get it now

jklukas (approved):

PR #1424, opened 2021-11-23
Fix burnham counters.

jrbenny35:
I added too many ```value_sum``` metrics, this reverts that.


jklukas (approved):

jklukas:
Verified the code is live in Airflow and cleared the DAG so that the new code can run.

PR #1440, opened 2022-01-04
[DSRE-360] Modify dags to use newer dataproc images for sec reasons

haroldwoo:
Newer dataproc images address Log4j vunerability
Needs to be tested (especially taar_daily because it used 1.3), or we can just YOLO and patch


willkg (approved):
Seems fine to me. I'll get an error if that was wrong and can figure out issues then.

akkomar (approved):
r+

eu9ene (approved):

PR #1444, opened 2022-01-07
added platform key for mysql container in docker-compose for m1 compatibility

kik-kik:
# added platform key for mysql container in docker-compose for m1 compatibility

I followed the instructions linked, I noticed that when trying to start Airflow locally with make up mysql image failed to be pulled with the following error:
`no matching manifest for linux/arm64/v8 in the manifest list entries`

I did a quick Google and was able to find a fix/get it working: https://stackoverflow.com/questions/65456814/docker-apple-silicon-m1-preview-mysql-no-matching-manifest-for-linux-arm64-v8

Seems to be mac M1 specific thing.


kik-kik:
okay, it looks like this won't work for now because docker-compose used by the `test-environment` step does not support `platform` key yet.

.circleci/config.yml
```diff
@@ -54,6 +54,8 @@ jobs:
       - checkout
       - run:
           command: |
+            sudo curl -L "https://github.com/docker/compose/releases/download/1.27.4/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+            sudo chmod +x /usr/local/bin/docker-compose
```
jklukas, line 58: So this was previously using a `docker-compose` present in the base `ubuntu-1604` image? Ubuntu 16 is past EOL, so we should probably use a newer ubuntu image anyway. Can you check if using the latest LTS ubuntu image gives us a working docker-compose out of the box?

docker-compose.yml
```diff
@@ -2,6 +2,9 @@ version: '2.1'
 
 services:
   db:
+    # platform key added for Macbook M1 compatibility, followed:
+    # https://stackoverflow.com/questions/65456814/docker-apple-silicon-m1-preview-mysql-no-matching-manifest-for-linux-arm64-v8
+    platform: linux/x86_64
```
jklukas, line 7: Thanks for including the commentary to document why this is here.

jklukas (approved):
r+wc

If we can get a working docker-compose by pulling a more recent ubuntu image, let's do that before merging this.

.circleci/config.yml
kik-kik: I can (will do on Monday), I didn't want to just upgrade the base image as I wasn't sure if it's used anywhere else in the project and would want to avoid different environment in prod and testing.

jklukas (approved):

PR #1446, opened 2022-01-10
Step to validate airflow dags as part of CI

### LINKED ISSUE #1443: Airflow DAG tag validation is missing
# Airflow DAG tag validation is missing

Our DAGs now include tags, however, we need a way to validate that all DAGs are tagged correctly. This means:
1. They contain all required tags
2. Tags specified have correct values.

We believe this check should occur in CircleCi before merging is allowed.

kik-kik:
# Step to validate airflow dags as part of CI

## Description
- `bin/test-dag-tags.py` script introduced to query `dag_tag` Airflow table to retrieve a list of all DAGs along with their tags and ensure they are correct.
- `test-environment` check now includes a command for executing a python script validating Airflow DAGs (part of `test-parse` script).
- installing mysql-client on airflow image to be able to query the DB for tag validation
- Added tag assignment to DAGs that were missing them

## Next step
- Integrate this check with `bigquery-etl` repos

resolves #1443

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,8 +58,9 @@ jobs:
             docker-compose build
             # now take ownership of the folder
             sudo chown -R 10001:10001 .
+
       - run:
-          name: Test if dag scripts can be parsed
+          name: Test if dag scripts can be parsed by Airflow and that tags have been correctly set
           command: bash bin/test-parse
 
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,7 @@ jobs:
 
       - run:
           name: Test if dag scripts can be parsed by Airflow and that tags have been correctly set
+          #  Valid DAG tags are defined in: `bin/test-dag-tags.py`
           command: bash bin/test-parse
 
   deploy:
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         apt-transport-https build-essential curl git libpq-dev python-dev \
         default-libmysqlclient-dev gettext sqlite3 libffi-dev libsasl2-dev \
-        lsb-release gnupg vim screen procps && \
+        lsb-release gnupg vim screen procps default-mysql-client && \
     CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" && \
     echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
     curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
--- a/bin/test-dag-tags.py
+++ b/bin/test-dag-tags.py
@@ -33,6 +33,7 @@ def _execute_shell_cmd(cmd: str, cmd_params: Dict[str, dict] = dict()) -> str:
         capture_output=True,
         text=True,
     )
+
     try:
         cmd_output.check_returncode()
     except subprocess.CalledProcessError as _err:
@@ -45,7 +46,7 @@ def _execute_shell_cmd(cmd: str, cmd_params: Dict[str, dict] = dict()) -> str:
     return cmd_output.stdout.strip().replace("\r", "").split("\n")
 
 
-def retrieve_existing_airflow_dags_from_db(pswd: str) -> Dict[str, str]:
+def get_loaded_airflow_dag_tags_from_db(pswd: str) -> Dict[str, str]:
     shell_cmd = "docker-compose exec web mysql -Ns -h db -u root -p{pswd} airflow -e 'SELECT dag_id, name FROM dag_tag;'"
     cmd_params = {
         "pswd": {
@@ -67,7 +68,7 @@ def retrieve_existing_airflow_dags_from_db(pswd: str) -> Dict[str, str]:
     return dags
 
 
-def get_number_of_dags_loaded(pswd) -> int:
+def get_loaded_dags_from_db(pswd) -> int:
     cmd_params = {
         "pswd": {
             "value": pswd,
@@ -82,10 +83,10 @@ if __name__ == "__main__":
     # Assumes the web and db containers are already running
     tag_errors = 0
 
-    dags = get_number_of_dags_loaded(db_pass)
+    dags = get_loaded_dags_from_db(db_pass)
     num_of_dags = len(dags)
 
-    dags_with_tags = retrieve_existing_airflow_dags_from_db(db_pass)
+    dags_with_tags = get_loaded_airflow_dag_tags_from_db(db_pass)
     num_of_dags_with_tags = len(dags_with_tags)
 
     if num_of_dags != num_of_dags_with_tags:
--- a/bin/test-dag-tags.py
+++ b/bin/test-dag-tags.py
@@ -11,71 +11,46 @@ class DagValidationError(Exception):
     pass
 
 
-def retrieve_existing_airflow_dags():
-    output = subprocess.run(
-        "docker-compose exec web mysql -Ns -h db -u root -psecret airflow -e 'SELECT dag_id, name FROM dag_tag;'",
+def retrieve_existing_airflow_dags_from_db(pswd):
+    mask_string = "###"
+    shell_cmd = "docker-compose exec web mysql -Ns -h db -u root -p{pswd} airflow -e 'SELECT dag_id, name FROM dag_tag;'"
+
+    print("Executing command: %s" % (shell_cmd.format(pswd=mask_string)))
+    cmd_output = subprocess.run(
+        shell_cmd.format(pswd=pswd),
         shell=True,
-        capture_output=True
-    ).stdout.decode('ascii').strip()
+        capture_output=True,
+        text=True,
+    )
+    try:
+        cmd_output.check_returncode()
+    except subprocess.CalledProcessError as _err:
+        print(cmd_output.stdout)
+        print(cmd_output.stderr)
+        raise _err
+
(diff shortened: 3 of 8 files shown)


haroldwoo (approved):

bin/test-dag-tags.py
```diff
@@ -0,0 +1,52 @@
...
+REQUIRED_TAG_TYPES = ("impact",)
+
+
+class DagValidationError(Exception):
+    pass
+
+
+def retrieve_existing_airflow_dags():
+    output = subprocess.run(
+        "docker-compose exec web mysql -Ns -h db -u root -psecret airflow -e 'SELECT dag_id, name FROM dag_tag;'",
```
haroldwoo, line 16: Worth exploring if the airflow command line is able to get you tag information instead.

kik-kik: I checked, unfortunately it is not possible.

Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         apt-transport-https build-essential curl git libpq-dev python-dev \
         default-libmysqlclient-dev gettext sqlite3 libffi-dev libsasl2-dev \
-        lsb-release gnupg vim screen procps && \
+        lsb-release gnupg vim screen procps default-mysql-client && \
```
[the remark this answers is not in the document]
kik-kik: that change was already there. I'm open to suggestions.

bin/test-dag-tags.py
haroldwoo: I would parse the python files on the filesystem rather than relying on a mysql table. The table may change over time, and also dag state for deleted dags, etc. is not always clear. Also we have multiple github repos

haroldwoo: Does the circleCi environment even have the dags loaded into mysql?

kik-kik: yes, it happens automatically when airflow spins up. I assume that's actually how airflow command for listing dags works. It most likely fetches them from the `dags` table.

Anyways, I added an alternative approach to checking dag tags.

kik-kik: Please let me know what you think and then I'll make final tweaks.

bin/test-dag-tags.py
```diff
@@ -0,0 +1,91 @@
...
+        "Tag.ImpactTier.tier_2": "impact/tier_2",
+        "Tag.ImpactTier.tier_3": "impact/tier_3",
+        "repo/telemetry-airflow": "impact/tier_3",
+        "repo/bigquery-etl": "repo/bigquery-etl",
+    }
+
+
+    dags = defaultdict(list)
+
+    for file in glob.glob("dag_dir/*.py"):
```
haroldwoo, line 49: does .glob do recursive lookups

haroldwoo: also this may pull in library files that we've written

bin/test-parse
```diff
@@ -88,6 +88,15 @@ function main() {
         echo "Unit tests passed!"
     fi
 
+    # Validate Airflow tags
+    # see Github Issue: https://github.com/mozilla/telemetry-airflow/issues/1443
+    python3 bin/test-dag-tags.py
```
haroldwoo, line 93: i would put this in a separate script, and a separate CircleCI test check as well

bin/test-dag-tags.py
haroldwoo: added comments. I still prefer the non sql approach. we can let klukas chime in also

haroldwoo: after some thought, i dont feel too strongly about this. If it breaks we can just repair

bin/test-parse
kik-kik: yes, if we decide to go with the parsing approach then I'd move it into it's own step in circleci

bin/test-dag-tags.py
kik-kik: I tried it locally and it seems to behave as expected. I can double check to make sure thought.

kik-kik:
@jklukas what are your thoughts?

bin/test-dag-tags.py
```diff
@@ -0,0 +1,91 @@
...
+    for dag in output.replace("\r", "").split("\n"):
+        dag_name, tag_name = dag.split("\t")
+
+        dags[dag_name] = dags[dag_name] + [tag_name]
+
+    return dags
+
+
+import glob, os
+import re
```
jklukas, line 32: Can these imports be at the top of the file?

bin/test-dag-tags.py
```diff
@@ -0,0 +1,91 @@
...
+
+# for file in glob.glob("*.txt"):
+#     print(file)
```
jklukas, line 35: Should remove this before merging.

bin/test-dag-tags.py
```diff
@@ -0,0 +1,91 @@
...
+
+def get_dag_tags(dag_dir: str):
+    tag_mapping = {
+        "Tag.ImpactTier.tier_1": "impact/tier_1",
+        "Tag.ImpactTier.tier_2": "impact/tier_2",
+        "Tag.ImpactTier.tier_3": "impact/tier_3",
+        "repo/telemetry-airflow": "impact/tier_3",
+        "repo/bigquery-etl": "repo/bigquery-etl",
+    }
```
jklukas, line 44: Maintaining this mapping here seems problematic. I think this should be generated from the enum rather than duplicated here.

bin/test-dag-tags.py
jklukas: I prefer the sql approach, as it involves working with parsed DAGs. The textual approach could have a lot of unexpected edge cases and will be brittle based on having to maintain mappings from enum constructs to text values.

haroldwoo: The sql approach is more correct. The parsing would be quicker, and less brittle imo. I don't feel strongly either way, and will rely on kik to be the go to person for ci issues regarding this in the future

kik-kik: 🙈 <- literally me right now 😅

haroldwoo: if you go sql, try catch your subprocess.run and print out error codes, exceptions etc.

bin/test-parse
kik-kik: Ok, I'd say I'll go with the SQL approach. As Jeff mentioned this way we are already interacting with parsed values by Airflow so we can be sure they are set correctly.

For the above reason, I will leave the script execution where it correctly is in order to avoid having to spin up Airflow locally twice in a row (docker-compose up) as it will just extend the time the pipeline takes to complete.

bin/test-dag-tags.py
kik-kik: So in this case I made a small tweak to the `.run` call, I believe there is no need for any additional error handling here for that specific command and output should be redirected to `STDOUT` and exception should be raised if `non-zero` return code is present for the shell command.

kik-kik: apologies, it looks like I actually have to manually output the errors. Adding that now.

kik-kik:
@haroldwoo @jklukas ok, I'd say this is the final version. Please let me know if you have any comments.

bin/test-dag-tags.py
```diff
@@ -0,0 +1,110 @@
...
+    return _execute_shell_cmd("docker-compose exec web mysql -Ns -h db -u root -p{pswd} airflow -e 'SELECT dag_id from dag WHERE is_subdag = 0;'", cmd_params)
+
+
+if __name__ == "__main__":
+    db_pass = "secret"
+    # Assumes the web and db containers are already running
+    tag_errors = 0
+
+    dags = get_number_of_dags_loaded(db_pass)
```
haroldwoo, line 85: nit: func name implies it returns length. change name to something like get_loaded_dags_from_sql

bin/test-dag-tags.py
```diff
@@ -0,0 +1,110 @@
...
+        key: val["value"]
+        for key, val in cmd_params.items()
+    }
+
+    cmd_output = subprocess.run(
+        cmd.format(**cmd_params_formatted),
+        shell=True,
+        capture_output=True,
+        text=True,
+    )
```
haroldwoo, line 35: nit: whitespace

.circleci/config.yml
```diff
@@ -58,8 +58,9 @@ jobs:
             docker-compose build
             # now take ownership of the folder
             sudo chown -R 10001:10001 .
+
       - run:
-          name: Test if dag scripts can be parsed
+          name: Test if dag scripts can be parsed by Airflow and that tags have been correctly set
```
haroldwoo, line 63: maybe add a comment noting that valid tags are defined in  bin/test-dag-tags.py

bin/test-dag-tags.py
```diff
@@ -0,0 +1,110 @@
...
+        print(cmd_output.stdout)
+        print(cmd_output.stderr)
+        raise _err
+
+    print("Command executed successfully, processing output...")
+
+    return cmd_output.stdout.strip().replace("\r", "").split("\n")
+
+
+def retrieve_existing_airflow_dags_from_db(pswd: str) -> Dict[str, str]:
```
haroldwoo, line 48: nit: name get_dags_and_tags_from_sql or similar to whatever you rename the func on line 70 to

haroldwoo (approved):
lgtm, just had some nits

bin/test-dag-tags.py
kik-kik: that's a fair point, I believe originally that's what it did then I changed it and forgot to update the fun name. Fixed that now.

bin/test-dag-tags.py
kik-kik: Not sure I got it, you mean add a newline between `cmd_output` and `try`?

PR #1449, opened 2022-01-14
Addressing issue with more dags in tags tbl than in dags table

kik-kik:
# Addressing issue with more dags in tags tbl than in dags table

In some cases it seems we can have dags with tags not in dags table, in this case print warning but don't fail check.

Not sure why we see this error, in dev branch have not happened before. This adds logic to handle this specific case and log which dag is "missing"


jklukas (approved):
One style comment.

PR #1450, opened 2022-01-19
Django dev testing

alekhyamoz:
As part of testing the glam import jobs in dev env for FOG. I have added the import tasks and modified the env variables to point to the dev environment. 

Once the testing is successful, will migrate the jobs to prod env.


scholtzan (approved):
lgtm

PR #1452, opened 2022-01-20
separating the glam glean imports function from the etl

alekhyamoz:
The glam_fenix and glam_firefox_desktop currently have "glam imports" as part of the etl run. 
The goal is to separate the imports from the run to avoid the redundancy of running the import jobs in both dags. 
As part of this PR:
1. created new dag glam_glean_imports.py which runs glean imports. This dag waits on glam_fenix and glam_firefox_desktop to complete. The parent dags are scheduled to run at 2 am UTC. I have set the child dags to run at 5 am UTC
2. Renames glam_firefox_desktop dag to glam_fog
3. Removed the dev variables added as part of the [PR](https://github.com/mozilla/telemetry-airflow/pull/1450) from bin/run


dags/glam_glean_imports.py
```diff
@@ -0,0 +1,123 @@
...
+
+glam_import_glean_counts = GKENatPodOperator(
+    task_id = 'glam_import_glean_counts',
+    name = 'glam_import_glean_counts',
+    image = glam_import_image,
+    arguments = base_docker_args + ['import_glean_counts'],
+    env_vars = env_vars,
+    dag=dag)
+
+wait_for_fenix >> wait_for_fog
```
kik-kik, line 119: Is this really the actual representation of the dependency?

All the tasks should wait until both sensors complete successfully, right? But both should be independent of each other. In this case the `wait_for_fog` sensor would only start running once `wait_for_fenix` completed.

if this is not the intention then I'd suggest doing something like instead:
```
[wait_for_fenix, wait_for_fog] >> glam_import_glean_aggs_beta
[wait_for_fenix, wait_for_fog] >> glam_import_glean_aggs_nightly
[wait_for_fenix, wait_for_fog] >> glam_import_glean_aggs_release
[wait_for_fenix, wait_for_fog] >> glam_import_glean_counts 
```

It just more explicitly shows that the two sensor tasks are independent of each other and should be executed in parallel.

dags/glam_glean_imports.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+# Make sure all the data for the given day has arrived before running.
+wait_for_fenix = ExternalTaskCompletedSensor(
+    task_id="wait_for_fenix",
+    external_dag_id="glam_fenix",
+    external_task_id="pre_import",
+    execution_delta=timedelta(hours=3),
+    check_existence=True,
+    mode="reschedule",
```
kik-kik, line 54: Does this task ever fail? I'm just wondering how do we know it's actually being executed instead of just constantly being rescheduled?

kik-kik (approved):
LGTM 👍 Just left a couple of small comments.

dags/glam_glean_imports.py
alekhyamoz: Makes sense.

PR #1461, opened 2022-02-04
attempting to see if this is enough to resolve bug: 1751206

kik-kik:
https://bugzilla.mozilla.org/show_bug.cgi?id=1751206

There seems to report of people seeing this error after upgrading spark. I'm not sure if this actually is why we started to see this problem. This is the simple potential "solution" I found and thought it's worth a try.


akkomar (approved):
Worth trying, however I think I once saw a similar error here to be temporary.

kik-kik:
Yeah, I don't really see any obvious changes to the environment. But this doesn't seem like it can hurt to try so thought why not. I look forward to our session later on to understand out spark set up a little bit better :)

PR #1462, opened 2022-02-04
Adding few extra links to the Mozilla menu button for Airflow

kik-kik:
Hopefully this is a step in making Airflow triage quicker and easier.

### DIFF
--- a/plugins/mozmenu.py
+++ b/plugins/mozmenu.py
@@ -19,6 +19,24 @@ wtmo_dev = {
     "href": "https://mana.mozilla.org/wiki/display/DOPS/WTMO+Developer+Guide"
 }
 
+airflow_triage_guide = {
+    "name": "Airflow Triage Guide",
+    "category": "Mozilla",
+    "href": "https://mana.mozilla.org/wiki/display/DATA/Airflow+Triage+Process",
+}
+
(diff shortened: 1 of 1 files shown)


plugins/mozmenu.py
```diff
@@ -19,11 +19,29 @@
     "href": "https://mana.mozilla.org/wiki/display/DOPS/WTMO+Developer+Guide"
 }
 
+airflow_triage_guide = {
+    "name": "Airflow Triage Guide",
+    "category": "Mozilla",
+    "href": "https://mana.mozilla.org/wiki/display/DATA/Airflow+Triage+Process",
+}
```
jklukas, line 26: I think of this Mana link as canonical for triage, so adding this one here looks good. The two others below I think are too prone to iteration to include here. I'd rather we just link to the Mana page, and the Mana page is the one place where we'd be maintaining these links for bugzilla interaction.

jklukas (changes requested):

plugins/mozmenu.py
kik-kik: that's fair, I'll remove the other two then.

jklukas (approved):

haroldwoo (approved):

PR #1467, opened 2022-02-14
Rename shredder tasks now that everything will be flat rate

relud:
and specify `--no-use-dml` for all tasks.

shredder tracks state by target, not by airflow task id, and is designed to gracefully handle restarts. It should handle renaming all of the tasks without issue, but the new tasks will need to be manually started in the current shredder run.

### DIFF
--- a/dags/shredder.py
+++ b/dags/shredder.py
@@ -64,34 +64,21 @@ base_command = [
     # race conditions with downstream tables and pings received shortly after a
     # deletion request.
     "--start-date={{macros.ds_add(ds, 27-28*2)}}",
+    # non-dml statements use LEFT JOIN instead of IN to filter rows, which takes about
+    # half as long as of 2022-02-14, and reduces cost by using less flat rate slot time
+    "--no-use-dml",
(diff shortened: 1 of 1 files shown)


relud:
I'm fine with losing airflow history, I prefer using the `mozdata.monitoring.shredder_progress` view for old runs.

we don't need to wait for current runs to finish, because the new tasks will take over the work of the old tasks without losing progress, and airflow is configured to delete the k8s pod when a task stops.

I expect this won't immediately kick off a new round of tasks, based on how new tasks have behaved in other dags, instead I will need to manually start the new tasks. If it does kick off new tasks immediately, all the better.

jklukas (approved):
> we don't need to wait for current runs to finish, because the new tasks will take over the work of the old tasks without losing progress, and airflow is configured to delete the k8s pod when a task stops.

As long as we can verify that the old tasks stop once this is merged. It seems possible that the DAG would no longer track the task, but the task would keep running, and we'd end up with two logically identical tasks running at the same time.

PR #1470, opened 2022-02-18
feat(DENG-17): Acoustic data in BigQuery

kik-kik:
# feat(DENG-17): Acoustic data in BigQuery

DAG to schedule Fivetran run.

### DIFF
--- a/dags/fivetran_acoustic.py
+++ b/dags/fivetran_acoustic.py
@@ -1,0 +1,118 @@
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.operators.python_operator import PythonOperator
+from airflow.hooks.base import BaseHook
+from operators.backport.fivetran.operator import FivetranOperator
+from operators.backport.fivetran.sensor import FivetranSensor
+from utils.tags import Tag
+from utils.acoustic.acoustic_client import AcousticClient
+
+
+def _generate_acoustic_report(conn_id, report_type, config, *args, **kwargs):
+    conn = BaseHook.get_connection(conn_id)
+
+    acoustic_connection = {
+        "base_url": conn.host,
+        "client_id": conn.login,
+        "client_secret": conn.password,
+        "refresh_token": conn.extra_dejson["refresh_token"],
+    }
+
+    acoustic_client = AcousticClient(**acoustic_connection)
+    acoustic_client.generate_report(request_template=config["request_template"], template_params=config["request_params"], report_type=report_type)
+
+    return
+
+
+DOCS = """
+### fivetran_acoustic
+
+#### Description
+
+DAG triggers acoustic report generation, then corresponding Fivetran connector to consume the generated report
+from Acoustic SFTP server and waits until its completion.
+
+#### Owner
+
+kignasiak@mozilla.com
+"""
+
+DAG_OWNER = "kignasiak@mozilla.com"
+
+ACOUSTIC_CONNECTION_ID = "acoustic"
+
+EXEC_START = '{{ macros.ds_format(ds, "%Y-%m-%d", "%m/%d/%Y %H:%M:%S") }}'
+EXEC_END = '{{ macros.ds_format(ds, "%Y-%m-%d", "%m/%d/%Y %H:%M:%S") }}'
+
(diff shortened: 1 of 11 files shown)


kik-kik:
@scholtzan I believe this should be it from the Airflow side, I will also add a doc to fivetran repo with steps to create the connectors in Fivetran using the `SFTP` connector.

Please let me know if you have any questions or comments.

kik-kik:
Note: have not been able to test end-to-end since I do not have API key for Fivetran to create the connection locally on my machine.

dags/fivetran_acoustic.py
```diff
@@ -0,0 +1,119 @@
...
+
+#### Owner
+
+kignasiak@mozilla.com
+"""
+
+DAG_OWNER = "kignasiak@mozilla.com"
+
+ACOUSTIC_CONNECTION_ID = "acoustic"
+# TODO: add airflow viariable containing fivetran connector ID
```
ANich, line 44: Is this a pre-merge TODO?

dags/fivetran_acoustic.py
```diff
@@ -0,0 +1,119 @@
...
+    ) as dag:
+        generate_report = PythonOperator(
+            task_id="generate_acoustic_report",
+            python_callable=_generate_acoustic_report,
+            op_args=[ACOUSTIC_CONNECTION_ID, report_type, _config],
+        )
+
+        sync_trigger = FivetranOperator(
+            task_id='trigger_fivetran_connector',
+            connector_id=f"{{{{ var.value.fivetran_acoustic_{report_type}_connector_id }}}}",
```
ANich, line 109: Do you need four `{}` here?

dags/acoustic/acoustic_client.py
```diff
@@ -0,0 +1,188 @@
+"""
```
ANich, line 1: Should this be in the `dags/` directory? No strong opinion but this seems like `utils/`

ANich (approved):
Don't have a ton of the Acoustic context, but if you've been able to test it 👍 Left a few other comments as well

requirements.txt
```diff
@@ -1,238 +1,830 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.8
```
ANich, line 2: This is a pretty big diff that seems to be mostly adding line breaks... Is this an editor issue?

dags/fivetran_acoustic.py
kik-kik: oh yes, thanks removed now.

dags/fivetran_acoustic.py
```diff
@@ -0,0 +1,119 @@
...
+    },
+    "contact_export": {
+        "request_template": f"{REQUEST_TEMPLATE_LOC}/export_database.xml.jinja",
+        "request_params": {
+            "list_id": "{{ var.value.fivetran_acoustic_contact_export_list_id }}",  # list_name: "Main Contact Table revision 3"
+            # "list_id": 1390189,  # dev
+            # "list_id": 1364939,  # prod
+            "export_type": "ALL",
+            "export_format": "CSV",
+            "visibility": 1,  # 0 (Private) or 1 (Shared)
```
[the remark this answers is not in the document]
kik-kik: It's a fixed value and specific to this config and won't be used anywhere else so I think it's fine like this.

dags/fivetran_acoustic.py
kik-kik: yes, it's because I'm combining Airflow macro with f'' string formatting. Double `{}` wouldn't work in this case.

requirements.txt
kik-kik: so looks like this is caused by the version of `pip-tools`. This is older version format (I tried `pip-tools==5.4.0`), the newer versions use a new format. We should update this with another PR.

dags/acoustic/acoustic_client.py
kik-kik: moved the folder inside existing folder: `dags/utils/`

ANich (approved):

PR #1473, opened 2022-03-09
separate glam etl from imports

alekhyamoz:
This is separate the glam etl runs from the imports. This will give us the flexibility to turn on/off the imports during deployments. 
The schedule_interval="0 7 * * *" and needs to be modified depending on glam dag run times

### DIFF
--- a/dags/glam_imports.py
+++ b/dags/glam_imports.py
@@ -1,0 +1,119 @@
+"""
+Desktop ETL for importing glam data into GLAM app
+"""
+
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from operators.gcp_container_operator import GKENatPodOperator
+from operators.task_sensor import ExternalTaskCompletedSensor
+from airflow.models import Variable
+from airflow.operators.subdag_operator import SubDagOperator
+
+from glam_subdags.extract import extracts_subdag, extract_user_counts
+from glam_subdags.histograms import histogram_aggregates_subdag
+from glam_subdags.general import repeated_subdag
+from glam_subdags.generate_query import generate_and_run_desktop_query
+from utils.gcp import bigquery_etl_query, gke_command
+from utils.tags import Tag
+
+default_args = {
+    "owner": "akommasani@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2019, 10, 22),
+    "email": [
+        "telemetry-alerts@mozilla.com",
+        "akommasani@mozilla.com",
+    ],
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
+}
+
+
+
+tags = [Tag.ImpactTier.tier_2]
+
+dag = DAG(
+    "glam_imports",
+    default_args=default_args,
+    schedule_interval="0 7 * * *",
+    doc_md=__doc__,
+    tags=tags,
+)
+
+
+# Make sure all the data for the given day has arrived before running.
+wait_for_glam = ExternalTaskCompletedSensor(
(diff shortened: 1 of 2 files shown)


dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+    tags=tags,
+)
+
+
+# Make sure all the data for the given day has arrived before running.
+wait_for_glam = ExternalTaskCompletedSensor(
+    task_id="wait_for_glam",
+    external_dag_id="glam",
+    external_task_id="pre_import",
+    execution_delta=timedelta(hours=3),
```
scholtzan, line 52:
```
    execution_delta=timedelta(hours=5),
```
The glam DAG is running at 2AM, this DAG is running at 7AM, so the delta is 5.

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+
+tags = [Tag.ImpactTier.tier_2]
+
+dag = DAG(
+    "glam_imports",
+    default_args=default_args,
+    schedule_interval="0 7 * * *",
+    doc_md=__doc__,
+    tags=tags,
+)
```
scholtzan, line 44: I guess there is the risk of the GLAM DAG actually running longer than 5 hours and data not being entirely up-to-date when this DAG gets executed. Might be worth adding a comment to add this caveat here and maybe explain why this needs to be done.

scholtzan (approved):

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+
+glam_import_probes = GKENatPodOperator(
+    task_id = 'glam_import_probes',
+    name = 'glam_import_probes',
+    image = glam_import_image,
+    arguments = base_docker_args + ['import_probes'],
+    env_vars = env_vars,
+    dag=dag)
+
+[wait_for_glam] >> glam_import_desktop_aggs_beta
```
kik-kik, line 115: I don't believe the `[]` is needed here.

kik-kik, line 115:
```suggestion
wait_for_glam >> glam_import_desktop_aggs_beta
```

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+[wait_for_glam] >> glam_import_desktop_aggs_nightly
```
kik-kik, line 116:
```suggestion
wait_for_glam >> glam_import_desktop_aggs_nightly
```

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+[wait_for_glam] >> glam_import_desktop_aggs_release
```
kik-kik, line 117:
```suggestion
wait_for_glam >> glam_import_desktop_aggs_release
```

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+[wait_for_glam] >> glam_import_user_counts
```
kik-kik, line 118:
```suggestion
wait_for_glam >> glam_import_user_counts
```

dags/glam_imports.py
```diff
@@ -0,0 +1,119 @@
...
+[wait_for_glam] >> glam_import_probes
```
kik-kik, line 119:
```suggestion
wait_for_glam >> glam_import_probes
```

kik-kik: I think you can reduce the task flow statement to a single line like so:

```python
wait_for_glam >> [glam_import_desktop_aggs_beta, glam_import_desktop_aggs_nightly, glam_import_desktop_aggs_release, glam_import_user_counts, glam_import_probes]
```

kik-kik (approved):

PR #1474, opened 2022-03-17
Call burnham with Airflow Task ID

hackebrot:
See https://github.com/mozilla/burnham/pull/387


jklukas (approved):

hackebrot:
Hi @jklukas! 👋🏻 

Thank you for the review. I released a new version of burnham that adds support for the Airflow task ID parameter and just pushed a commit here to use that version for the client tasks.

Do you mind giving this another look?

PR #1478, opened 2022-03-28
Schedule KPI forecasting

scholtzan:
Depends on https://github.com/mozilla/docker-etl/pull/36

### DIFF
--- a/dags/kpi_forecasting.py
+++ b/dags/kpi_forecasting.py
@@ -2,6 +2,10 @@
 See [kpi-forecasting in the docker-etl repository]
 (https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting).
 
+This DAG runs the forecast for year-end KPI values for Desktop QCDOU, Mobile CDOU and Pocket.
+The forecast is only updated once per week since values aren't expected to change significantly
+day-to-day. The output powers KPI dashboards.
+
 This DAG is low priority.
 """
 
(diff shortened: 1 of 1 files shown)


dags/kpi_forecasting.py
```diff
@@ -0,0 +1,93 @@
...
+    "start_date": datetime(2022, 3, 28),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+tags = [Tag.ImpactTier.tier_3]
+
+with DAG("kpi_forecasting", default_args=default_args, schedule_interval="0 4 * * SAT", doc_md=__doc__, tags=tags,) as dag:
```
scholtzan, line 28: I changed the schedule slightly. The unified_metrics and mobile_usage queries are run at 3am UTC, since the KPI forecasting depends on them we can schedule it after they are likely done (4am).

scholtzan:
@perrymcmanis144 Please have a look at the DAG, just to make sure everything looks right.

perrymcmanis144:
> @perrymcmanis144 Please have a look at the DAG, just to make sure everything looks right.

LGTM & TYVM!

dags/kpi_forecasting.py
```diff
@@ -0,0 +1,93 @@
+"""
+See [kpi-forecasting in the docker-etl repository]
+(https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting).
```
jklukas, line 3: Can we add some more context to these docs? Particularly, I'd like to know why it only runs once per week, what are the outputs of this DAG, and what is it powering dashboards that stakeholders see?

dags/kpi_forecasting.py
```diff
@@ -0,0 +1,93 @@
...
+
+with DAG("kpi_forecasting", default_args=default_args, schedule_interval="0 4 * * SAT", doc_md=__doc__, tags=tags,) as dag:
+    dataset_yamls = ["yaml/mobile.yaml", "yaml/desktop.yaml"]
+
+    kpi_forecasting = gke_command(
+        task_id="kpi_forecasting",
+        command=[
+            "python", "kpi-forecasting/kpi-forecasting.py",
+            "-c",
+        ] + dataset_yamls,
```
jklukas, line 36: Can we split these out to separate tasks for desktop and mobile? I would expect we may want to be able to rerun separately if we fix upstream issues that are specific to desktop vs. mobile.

dags/kpi_forecasting.py
```diff
@@ -0,0 +1,93 @@
...
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+        email_on_retry=False,
+        dag=dag,
+    )
+
+    [
+        wait_for_clients_daily, 
+        wait_for_main_ping, 
+        wait_for_mobile_usage, 
+        wait_for_unified_metrics
```
jklukas, line 92: If `unified_metrics` already depends on all the above, then we may want to consider removing the other dependencies. My main concern is that each `ExternalTaskCompletedSensor` introduces some brittleness if we change scheduling of the upstream tasks, since we have to make sure to adjust all these execution deltas. A code comment about the implicit dependencies may be sufficient here.

But I also see some knowledge benefit to having these all explicitly declared and visible in the DAG.

jklukas (changes requested):
The only blocking thing here is more complete documentation about what this DAG is doing. Other suggestions are open for discussion.

dags/kpi_forecasting.py
scholtzan: yes, that makes sense

dags/kpi_forecasting.py
```diff
@@ -0,0 +1,101 @@
...
+        external_dag_id="copy_deduplicate",
+        external_task_id="copy_deduplicate_main_ping",
+        execution_delta=timedelta(hours=2),
+        check_existence=True,
+        mode="reschedule",
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+        email_on_retry=False,
+        dag=dag,
+    )
+
```
jklukas, line 75:
```suggestion
```

These definitions are no longer used.

jklukas (approved):
r+wc

PR #1480, opened 2022-03-30
[DSRE-596] add sample pubsub and slack operators

haroldwoo:
This will require a new airflow deployment to pick up library changes.

- PR adds example code for writing to pubsub (the topic must already exist)
- PR adds example code to show how to post to slack when a dag task fails
- PR also clears out airflow-worker.pid when user runs `make clean`

todo
- [x] test pubsub operator again because rebase and re pip-compiling requirements changes lib dependencies
- [x] test slack operator once bot and token are created
- [x] add failure hook to slack operator example and make it a reusable func
- [x] add connection and variable to wtmo
- [x] add connection and variable to sops file for record keeping
- [ ] cut release after merging this PR and deploy

Slack alerts will go to #airflow-alerts channel currently and will look similar to:
Screen Shot 2022-04-01 at 3 30 40 PM

### DIFF
--- a/dags/utils/slack.py
+++ b/dags/utils/slack.py
@@ -1,0 +1,22 @@
+from airflow.models import Variable
+from airflow.providers.slack.operators.slack import SlackAPIPostOperator
+
+SLACK_CHANNEL = "#airflow-alerts"
+
+def if_task_fails_alert_slack(context):
+    failed_alert = SlackAPIPostOperator(
+        task_id='slack_failed',
+        channel=SLACK_CHANNEL,
+        token=Variable.get("slack_secret_token"),
+        text="""
+            :red_circle: Task Failed.
+            *Task*: {task}
+            *Dag*: {dag}
+            *Date*: {ds}
+            """.format(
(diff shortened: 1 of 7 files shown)


dags/utils/slack.py
```diff
@@ -0,0 +1,22 @@
+from airflow.models import Variable
+from airflow.providers.slack.operators.slack import SlackAPIPostOperator
+
+SLACK_CHANNEL = "#airflow-alerts"
```
whd, line 4: This looks like the recommended way of interfacing with slack. I think this channel should be configurable in case specific job owners want to direct their slack notifications to a non-default channel.

whd (approved):
LGTM

dags/utils/slack.py
haroldwoo: hmm unfortunately when using a callback func it only expects a single argument for the context. We could change this to pull from Variables

haroldwoo: ill leave this as a future decision

PR #1481, opened 2022-04-04
Added DAG for static publish script

ANich:
Closes https://github.com/mozilla/bigquery-etl/issues/2836

### DIFF
--- a/dags/publish_bqetl_static.py
+++ b/dags/publish_bqetl_static.py
@@ -1,0 +1,54 @@
+"""
+Daily deployment of static bigquery-etl data to various projects.
+
+See the publish script [here](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/publish_static.py).
+"""
+
+from airflow import DAG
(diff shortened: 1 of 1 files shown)


dags/publish_bqetl_static.py
```diff
@@ -0,0 +1,56 @@
...
+) as dag:
+
+    publish_static_mozdata = gke_command(
+        task_id="publish_static_mozdata",
+        command=[
+            "script/publish_static",
+            "--project_id", "mozdata"
+        ],
+        docker_image=IMAGE,
+        dag=dag,
```
jklukas, line 45: When a DAG is used as a context manager, the `dag` param for tasks is set implicitly, so no `dag` argument is needed. We do have a few places in existing code where that could be cleaned up.

jklukas (approved):
r+wc

dags/publish_bqetl_static.py
ANich: Ty, will update to make it one fewer!

PR #1483, opened 2022-04-06
Bug 1740724 - Burnham: Run the sensor check against a deduped table

badboy:
This implements both suggestions I made in bug 1740724:

1. Deduping ping rows in the sensor query
2. Include the number of rows included in the calculation into the answer

2 is only done for one of the test queries. 
With 1. I would expect this to never be a mismatch again, so we could also leave it out.

### DIFF
--- a/dags/burnham.py
+++ b/dags/burnham.py
@@ -341,16 +341,25 @@ WANT_TEST_DELETION_REQUEST_PING_CLIENT_ID = [
 # the test scenarios to JSON and b64-encode them to ensure we can safely pass
 # this information to the burnham-bigquery Docker container. We can use
 # string-formatting here, because Airflow executes the query directly.
+# This tries to deduplicate based on the document_id, to avoid miscounting.
 SENSOR_TEMPLATE = """
+WITH {table}_numbered AS (
+  SELECT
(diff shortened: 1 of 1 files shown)


dags/burnham.py
```diff
@@ -82,23 +82,35 @@
...
   metrics.string.test_name = "{DEFAULT_TEST_NAME}"
-GROUP BY
-  technology_space_travel.key
 ORDER BY
-  technology_space_travel.key
+  key
 """
 
 WANT_TEST_LABELED_COUNTER_METRICS = [
+    {"key": "row_count", "value_sum": 20},
```
hackebrot, line 113: I imagine this may be useful to the other test scenarios as well. 🤔

hackebrot (commented):
The changes to the queries look good to me, but I'll defer @jklukas to approve that. 👨🏻‍🔬 

I've added an inline comment about replicating the `row_count` piece for other test scenarios.

dags/burnham.py
badboy: I would think so, just didn't spend the time on it before we agreed that it would.

jklukas (approved):

hackebrot (approved):

jrbenny35 (approved):

PR #1487, opened 2022-04-12
fixed a bug where we tried to generate report for 0 second time window

kik-kik:
fixed a bug where we tried to generate report for 0 second time window

### DIFF
--- a/dags/fivetran_acoustic.py
+++ b/dags/fivetran_acoustic.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from typing import Any, Dict
 
 from airflow import DAG
 from airflow.operators.python_operator import PythonOperator
@@ -9,9 +10,14 @@ from utils.tags import Tag
 from utils.acoustic.acoustic_client import AcousticClient
 
 
-def _generate_acoustic_report(conn_id, report_type, config, *args, **kwargs):
+def _generate_acoustic_report(conn_id: str, report_type: str, config: Dict[Any, Any], *args, **kwargs):
+    """
+    A wrapper function for retrieving Acoustic connection details from Airflow instantiating AcousticClient and generating report.
+    """
+
     if config["request_params"]["date_start"] == config["request_params"]["date_end"]:
-        raise ValueError("It appears start and end date are exactly the same.")
+        err_msg = "It appears start and end date are exactly the same. This is undesired and will result in data being generated for 0 second time range."
+        raise ValueError(err_msg)
 
     conn = BaseHook.get_connection(conn_id)
 
--- a/dags/fivetran_acoustic.py
+++ b/dags/fivetran_acoustic.py
@@ -10,6 +10,9 @@ from utils.acoustic.acoustic_client import AcousticClient
 
 
 def _generate_acoustic_report(conn_id, report_type, config, *args, **kwargs):
+    if config["request_params"]["date_start"] == config["request_params"]["date_end"]:
+        raise ValueError("It appears start and end date are exactly the same.")
+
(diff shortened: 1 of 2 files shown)


dags/fivetran_acoustic.py
```diff
@@ -10,6 +10,9 @@
 
 
 def _generate_acoustic_report(conn_id, report_type, config, *args, **kwargs):
```
lucia-vargas-a, line 12: Docstrings? Or what is our current practice for code documentation: functions & arguments data types and descriptions?

dags/fivetran_acoustic.py
```diff
@@ -10,6 +10,9 @@
...
+    if config["request_params"]["date_start"] == config["request_params"]["date_end"]:
+        raise ValueError("It appears start and end date are exactly the same.")
```
lucia-vargas-a, line 14: You might consider a more explanatory sentence of the expected behaviour, for example: "The start and end date cannot be the same, please check."

dags/utils/acoustic/acoustic_client.py
```diff
@@ -152,7 +152,8 @@ def generate_report(
 
         supported_report_types = ("raw_recipient_export", "contact_export",)
         if report_type not in supported_report_types:
-            pass  # TODO: raise with options
+            err_msg = f"{report_type} is not a valid option, support report types: {supported_report_types}"
```
lucia-vargas-a, line 155: "supported report types:"

dags/utils/acoustic/acoustic_client.py
```diff
@@ -152,7 +152,8 @@ def generate_report(
 
         supported_report_types = ("raw_recipient_export", "contact_export",)
         if report_type not in supported_report_types:
-            pass  # TODO: raise with options
```
lucia-vargas-a, line 155: You could omit "pass" here with the same result?

dags/utils/acoustic/acoustic_client.py
```diff
@@ -152,7 +152,8 @@ def generate_report(
...
+            err_msg = f"{report_type} is not a valid option, support report types: {supported_report_types}"
+            raise AttributeError(err_msg)
 
         with open(request_template, "r") as _file:
```
lucia-vargas-a, line 158: Ideally we could use more explicit names for variables, so that if they are used later in the code is easier to know what they refer to. In this case for example: _template_file or _request_file better than _file.

dags/utils/acoustic/acoustic_client.py
kik-kik: this line is being removed with this PR

dags/fivetran_acoustic.py
```diff
@@ -10,6 +10,9 @@
 
 
 def _generate_acoustic_report(conn_id, report_type, config, *args, **kwargs):
+    if config["request_params"]["date_start"] == config["request_params"]["date_end"]:
```
lucia-vargas-a, line 13: Do you want to raise an exception or use a default in case any of the "date_start" or "date_end" keys is missing? Or is it maybe handled already in the previous step?

dags/utils/acoustic/acoustic_client.py
kik-kik: This is just file object being loaded into memory and used only on the next line.

dags/fivetran_acoustic.py
lucia-vargas-a: You want a complete failure or a specific exception raised with the info?

lucia-vargas-a (approved):

PR #1488, opened 2022-04-14
tweak/acoustic load

kik-kik:
tweak/acoustic load

Once DAG for creating a derived dataset is merged, additional tasks for triggering it and waiting until completion need to be added.

### DIFF
--- a/dags/utils/acoustic/acoustic_client.py
+++ b/dags/utils/acoustic/acoustic_client.py
@@ -145,9 +145,6 @@ class AcousticClient:
         :param template_params: values that should be used to render the request template provided
 
         :return: Returns back a list of rows of data returned by the API call as a dictionary
-
-        More information about the specific API call:
-        https://developer.goacoustic.com/acoustic-campaign/reference/reporting#getaggregatetrackingfororg
         """
 
         supported_report_types = ("raw_recipient_export", "contact_export",)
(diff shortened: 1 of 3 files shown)


dags/utils/acoustic/acoustic_client.py
```diff
@@ -187,3 +184,53 @@ def generate_report(
         logging.info(f"{report_type} generation complete. Report location: {report_loc}. Time taken: {datetime.now() - start}")
 
         return
+
+
+if __name__ == "__main__":
+
+    EXEC_START = "01/04/2022 00:00:00"
+    EXEC_END = "02/04/2022 00:00:00"
```
ANich, line 192: This is for testing correct? Since this client is primarily used as a module? Might be worth mentioning that here (or putting this in a test?)

dags/utils/acoustic/acoustic_client.py
```diff
@@ -145,9 +145,6 @@ def generate_report(
         :param template_params: values that should be used to render the request template provided
 
         :return: Returns back a list of rows of data returned by the API call as a dictionary
-
-        More information about the specific API call:
-        https://developer.goacoustic.com/acoustic-campaign/reference/reporting#getaggregatetrackingfororg
```
ANich, line 150: Moved to a readme or something?

ANich (approved):
Don't have the acoustic context but if you've tested LGTM

dags/utils/acoustic/acoustic_client.py
kik-kik: no, this is no longer valid.

dags/utils/acoustic/acoustic_client.py
kik-kik: yeah, to be honest this was just for me generating a report locally so that I can check it worked without having to go through a PR process.

kik-kik: will remove later.

PR #1490, opened 2022-04-19
Update madserver DAG to use GCS instead of S3 for model and dataset storage.

gleonard-m:
- Moves storage of AMO datasets used for model training to GCS
- Adds task to train a malicious addon detector (MAD) model and store the model in GCS.  The task also evaluates the model and stores the report in a GCS bucket, to be available for viewing from protosaur.dev.
- Adds task to evaluate the new AMO data set and stores the report in a GCS bucket, to be available for viewing from protosaur.dev


dags/mad_server.py
```diff
@@ -62,18 +42,83 @@
...
+        gke_project_id="moz-fx-data-airflow-gke-prod",
+        gke_cluster_name="workloads-prod-v1",
+        gke_location='us-west1',
+        env_vars=dict(
+            GCS_BUCKET="mad-resources-training",
+            GCS_ROOT_TRAINING="datasets",
+            CUSTOMS_TRAINING_ALLOW_OVERWRITE="True",
+            CLOUD_SERVICE="GCS",
+            GCLOUD_PROJECT='mad-model-training',
+            GCS_REPORT_BUCKET='mad-reports',
```
jklukas, line 115: There's a good bit of duplication of configuration between these 3 tasks. It may be worth factoring out some of this to reusable constants.

jklukas (approved):
r+ with suggestion for reducing code duplication here.

dags/mad_server.py
gleonard-m: Update and re-rested

jklukas (approved):
Looks great!

jklukas:
If you have write privs here, feel free to merge. Otherwise, let me know and I can push the button.

jklukas:
This has propagated to WTMO. I went ahead and cleared the most recent run to exercise the full DAG and see if any issues crop up in prod: https://workflow.telemetry.mozilla.org/tree?dag_id=mad_server

PR #1492, opened 2022-04-21
Bug/acoustic debugging

kik-kik:
Bug/acoustic debugging

Acoustic generation seems wrong, trying to determine if the request body is correct.


dags/utils/acoustic/acoustic_client.py
```diff
@@ -157,6 +157,8 @@ def generate_report(
 
         request_body = request_body_template.render(**template_params)
 
+        print(f"DEBUG: {request_body}")
```
akkomar, line 160: Since `logging` is imported here, this can be:
```suggestion
        logging.debug(f"Request body: {request_body}")
```

akkomar (approved):
r+ to not block you; some nits:
* seems like this could be tested locally
* https://cbea.ms/git-commit/

dags/utils/acoustic/acoustic_client.py
kik-kik: yes, the problem is that locally it's behaving as expected. I want to check in our airflow deployment to make sure old code is not running.

PR #1494, opened 2022-04-21
Start Intacct Fivetran sensors before Fivetran syncs start (bug 1765180)

sean-rose:
When investigating [bug 1765180](https://bugzilla.mozilla.org/show_bug.cgi?id=1765180) I discovered a race condition with how `FivetranSensor` was set as downstream of the associated `FivetranOperator`:
> The [Fivetran sensor](https://github.com/mozilla/telemetry-airflow/blob/main/dags/operators/backport/fivetran/sensor.py) appears to work by reading the last connector run timestamp when it starts and then looking for a change in that. However, if the sensor task is delayed for some reason it might start after the Fivetran sync already completed, in which case it will end up waiting for the next sync, which will be the next daily sync triggered by Airflow.
> 
> That appears to have been the case here where the `intacct-sensor-china_vie` task didn't start until 2022-04-19 02:05:34, while the associated `sage_intacct_china_vie` Fivetran connector sync had already completed at 2022-04-19 02:05:27.

This PR attempts to avoid those race conditions by having the sensors start _before_ their associated Fivetran syncs.  Unfortunately there's no straightforward way to do that in Airflow using normal task dependencies, so it removes the dependency between the `FivetranSensor` and `FivetranOperator` tasks, and uses a parent DummyOperator to still group the tasks together.

This also fixes the timeout for `FivetranSensor`s (which run in `poke` mode by default and thus wouldn't respect the `timeout` argument which apparently only applies in `reschedule` mode), and sets the sensors to not retry (otherwise the retried sensors would just wait indefinitely for another sync to happen because the sync doesn't get retried).

### DIFF
--- a/dags/intacct_historical.py
+++ b/dags/intacct_historical.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 
 from airflow import DAG
+from airflow.utils.task_group import TaskGroup
 from operators.backport.fivetran.operator import FivetranOperator
 from operators.backport.fivetran.sensor import FivetranSensor
 from airflow.operators.dummy import DummyOperator
@@ -64,21 +65,23 @@ with DAG(
     )
 
     for location, connector_id in list_of_connectors.items():
-
-        fivetran_sync_start = FivetranOperator(
+        with TaskGroup(f'intacct-{location}', prefix_group_id=False):
+            fivetran_sync_start = FivetranOperator(
                 task_id='intacct-task-{}'.format(location),
                 fivetran_conn_id='fivetran',
                 connector_id='{}'.format(connector_id)
-        )
+            )
 
-        fivetran_sync_wait = FivetranSensor(
+            # It's best if the sensor starts before the Fivetran sync is triggered to avoid any
+            # chance of it missing the Fivetran sync happening, so we give it a higher priority and
+            # don't set it as downstream of the sync start operator.
+            fivetran_sync_wait = FivetranSensor(
                 task_id='intacct-sensor-{}'.format(location),
                 fivetran_conn_id='fivetran',
                 connector_id='{}'.format(connector_id),
                 poke_interval=5,
                 timeout=3*60*60,  # Timeout of 3 hours
-        )
+                priority_weight=fivetran_sync_start.priority_weight + 1,
+            )
 
-        fivetran_sync_start >> fivetran_sync_wait >> fivetran_sensors_complete
-
-
+            fivetran_sync_wait >> fivetran_sensors_complete
--- a/dags/intacct_historical.py
+++ b/dags/intacct_historical.py
@@ -67,18 +67,18 @@ with DAG(
     for location, connector_id in list_of_connectors.items():
         with TaskGroup(f'intacct-{location}', prefix_group_id=False):
             fivetran_sync_start = FivetranOperator(
-                task_id='intacct-task-{}'.format(location),
+                task_id=f'intacct-task-{location}',
                 fivetran_conn_id='fivetran',
-                connector_id='{}'.format(connector_id)
+                connector_id=connector_id,
             )
 
             # It's best if the sensor starts before the Fivetran sync is triggered to avoid any
             # chance of it missing the Fivetran sync happening, so we give it a higher priority and
             # don't set it as downstream of the sync start operator.
             fivetran_sync_wait = FivetranSensor(
-                task_id='intacct-sensor-{}'.format(location),
+                task_id=f'intacct-sensor-{location}',
                 fivetran_conn_id='fivetran',
-                connector_id='{}'.format(connector_id),
+                connector_id=connector_id,
                 poke_interval=5,
                 timeout=3*60*60,  # Timeout of 3 hours
                 priority_weight=fivetran_sync_start.priority_weight + 1,
--- a/dags/intacct_historical.py
+++ b/dags/intacct_historical.py
@@ -83,6 +83,7 @@ with DAG(
             connector_id=connector_id,
             poke_interval=5,
             execution_timeout=timedelta(hours=3),
+            retries=0,
             priority_weight=fivetran_sync_start.priority_weight + 1,
         )
         fivetran_sync >> fivetran_sync_wait >> fivetran_sensors_complete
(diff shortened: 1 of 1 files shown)


dags/intacct_historical.py
```diff
@@ -64,21 +65,23 @@
...
+            # It's best if the sensor starts before the Fivetran sync is triggered to avoid any
+            # chance of it missing the Fivetran sync happening, so we give it a higher priority and
+            # don't set it as downstream of the sync start operator.
+            fivetran_sync_wait = FivetranSensor(
+                task_id=f'intacct-sensor-{location}',
                 fivetran_conn_id='fivetran',
-                connector_id='{}'.format(connector_id),
+                connector_id=connector_id,
                 poke_interval=5,
-                timeout=3*60*60,  # Timeout of 3 hours
```
kik-kik, line 79: Are you sure we don't want to fail after 3 hours? I had a case in the past where the sensor task was stuck for a day and only realised after noticing manually.

kik-kik (approved):

scholtzan (approved):

sean-rose:
I'm having second thoughts about this, as it will make rescheduling failed Intacct Fivetran syncs more complicated and error prone:  You'll have to first clear the sensor, then clear the main sync task, and I doubt the two tasks would be near each other in the default tree view (I think TaskGroups only help with graph view).

A less robust but simpler way to try to mitigate this problem is to keep the existing task dependencies but give the sensors a much higher `priority_weight` (e.g. 100) to try to have them start ASAP after the Fivetran sync is triggered.

@scholtzan & @kik-kik what do you think?

kik-kik:
not sure I got it right, but why do you think re-running tasks would be more difficult? You can just clear the upstream task and make downstream tasks by default get also cleared? You still have the dependency specified within the task group.

You could also just add a dummy task at the beginning and use that to clear the whole run?

dags/intacct_historical.py
```diff
@@ -64,21 +65,23 @@
...
-        )
-
-        fivetran_sync_start >> fivetran_sync_wait >> fivetran_sensors_complete
-
+                execution_timeout=timedelta(hours=3),
+                priority_weight=fivetran_sync_start.priority_weight + 1,
+            )
 
+            fivetran_sync_wait >> fivetran_sensors_complete
```
kik-kik, line 87: I believe this should stil be:

```python
fivetran_sync_start >> fivetran_sync_wait >> fivetran_sensors_complete
```

dags/intacct_historical.py
```diff
@@ -64,21 +65,23 @@
...
                 fivetran_conn_id='fivetran',
-                connector_id='{}'.format(connector_id)
-        )
-
-        fivetran_sync_wait = FivetranSensor(
-                task_id='intacct-sensor-{}'.format(location),
+                connector_id=connector_id,
+            )
+
+            # It's best if the sensor starts before the Fivetran sync is triggered to avoid any
```
kik-kik, line 75: But if there is no active sync would this just immediately be marked as successful?

dags/intacct_historical.py
sean-rose: As originally envisioned, this change is the point of the PR, because if you start the Fivetran sync and only then start the sensor there's no guarantee how quickly Airflow will schedule it, and in some cases the sensor is starting minutes later (Airflow busy with other stuff) after the Fivetran sync has completed.

kik-kik: would in this case the sensor just say I don't see an active sync in progress in Fivetran and exit successfully? In this case this would be fine?

dags/intacct_historical.py
sean-rose: No, when the sensor starts running it looks at the last sync that happened, then sits there waiting for another sync to happen.  The sensor only completes when a sync happens during the time the sensor is running.

sean-rose:
> You could also just add a dummy task at the beginning and use that to clear the whole run?

This is probably a better idea than using TaskGroups.  It would avoid the issue of manual rescheduling being more complicated/error prone.

dags/intacct_historical.py
kik-kik: hmm, maybe it would be worth contributing to the community and adding an ability to check if the most recent run in Fivetran occurred within x amount of time of time for cases like this.

kik-kik: https://github.com/fivetran/airflow-provider-fivetran/blob/c65bb58130951a90bdcefea63880d2d609f61034/fivetran_provider/hooks/fivetran.py#L284-L290

Looks like somewhere around here we could add a date if after which a success occurred we just return True.

kik-kik: But thanks for raising this, let's hope our sync takes longer than it takes for Airflow to schedule the sensor task.

dags/intacct_historical.py
sean-rose: > would in this case the sensor just say I don't see an active sync in progress in Fivetran and exit successfully? In this case this would be fine?

No (see other thread).

dags/intacct_historical.py
sean-rose: > hmm, maybe it would be worth contributing to the community and adding an ability to check if the most recent run in Fivetran occurred within x amount of time of time for cases like this.

> https://github.com/fivetran/airflow-provider-fivetran/blob/c65bb58130951a90bdcefea63880d2d609f61034/fivetran_provider/hooks/fivetran.py#L284-L290
> 
> Looks like somewhere around here we could add a date if after which a success occurred we just return True.

A similar idea I had was using XCom to pass the sync start timestamp to the sensor.

Or another idea would be to have a single custom operator that starts the sync and waits for it to complete.  A big benefit of that would be making retries of failed syncs straightforward.  As it is, failed syncs won't get retried unless we find a way to have the sensor task retries also trigger the associated sync task to retry (seems possible with using `on_retry_callback` and/or sub-DAGs, but complicated).

However, for this PR I'm trying to keep things simple and mitigate the problems we're seeing without committing to more custom Fivetran Airflow operator/sensor code (partly because I think it'd be good if we can use Fivetran's provided packages rather than our customized copy of it).

scholtzan (approved):
That looks like a more reasonable workaround

PR #1496, opened 2022-04-26
[DSRE-658] Add publish to pubsub for glam import migration.

haroldwoo:
Next step is to modify the import job container to subscribe to the topic and verify msgs prior to importing.

This should be done in dev/stage first with separate subscriptions. Once this is working, we can replace the Airflow GkePodOperators that run the migrate.py <args> with the new GkeCronJob||GkeDeployment in Glam

### DIFF
--- a/dags/glam_glean_imports.py
+++ b/dags/glam_glean_imports.py
@@ -2,7 +2,8 @@
 Desktop ETL for importing glean data into GLAM app
 """
 
-from datetime import datetime, timedelta
+import datetime
+from datetime import timedelta
 
 from airflow import DAG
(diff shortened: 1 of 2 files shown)


dags/glam_glean_imports.py
```diff
@@ -2,13 +2,15 @@
 Desktop ETL for importing glean data into GLAM app
 """
 
-from datetime import datetime, timedelta
+import datetime
```
kik-kik, line 5: Is there a specific reason why you decided to change the import? I saw you use `datetime.date`, could also just add `date` to the existing import?

PR #1498, opened 2022-04-27
renamed xml request template file to a new name and updated the logic

kik-kik:
renamed xml request template file to a new name and updated the logic

This is because the production deployment keep executing with the old version of the template. Trying to rename the file.


dags/utils/acoustic/acoustic_client.py
```diff
@@ -188,90 +186,3 @@ def generate_report(
         logging.info(f"{report_type} generation complete. Report location: {report_loc}. Time taken: {datetime.now() - start}")
 
         return
-
```
lucia-vargas-a, line 191: Are you sure this file isn't needed anywhere else?

kik-kik: hey, no it's not. This is the only place.

lucia-vargas-a (approved):

PR #1502, opened 2022-05-02
Increase `copy_deduplicate` DAG task retries from 1 to 2

sean-rose:
To better tolerate transient GCP issues, like how the `copy_deduplicate_main_ping` task failed twice and stopped on [2022-04-30](https://workflow.telemetry.mozilla.org/log?dag_id=copy_deduplicate&task_id=copy_deduplicate_main_ping&execution_date=2022-04-29T01%3A00%3A00%2B00%3A00).

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -47,8 +47,8 @@ default_args = {
     "email_on_failure": True,
     "email_on_retry": True,
     "depends_on_past": False,
-    # If a task fails, retry it once after waiting at least 5 minutes
-    "retries": 1,
+    # If a task fails, retry it twice after waiting at least 5 minutes
+    "retries": 2,
     "retry_delay": datetime.timedelta(minutes=5),
 }
 


dags/copy_deduplicate.py
```diff
@@ -47,8 +47,8 @@
     "email_on_failure": True,
     "email_on_retry": True,
     "depends_on_past": False,
-    # If a task fails, retry it once after waiting at least 5 minutes
-    "retries": 1,
+    # If a task fails, retry it twice after waiting at least 5 minutes
+    "retries": 2,
     "retry_delay": datetime.timedelta(minutes=5),
```
scholtzan, line 52: I wonder if we should increase the delay here at little too? Since often re-running it almost right away still results in the same error, but waiting a little longer usually works.

scholtzan: I guess it depends what the actual reason for the error is. Maybe the support ticket will give more insight

jklukas (approved):

dags/copy_deduplicate.py
jklukas: In practice, these failures usually happen more than an hour into processing anyway, so there's natural delay built in that way.

sean-rose: Yeah, for the incident on 2022-04-30 retrying after 5 minutes did work and copy jobs were succeeding, but then a couple hours later one of the copy jobs failed again.

I wonder if it would be worth adding try/catch/retry logic within `copy_deduplicate` for specific steps, because when the `copy_deduplicate_main_ping` task was getting re-run on 2022-04-30 it was spending hours redoing work that had already been done in previous tasks that had gotten pretty far and failed toward the end.

PR #1504, opened 2022-05-10
Update Glam dag ownership

akkomar:


kik-kik (approved):

dags/glam.py
```diff
@@ -75,7 +75,7 @@
     destination_table="latest_versions",
     dataset_id=dataset_id,
     project_id=project_id,
-    owner="msamuel@mozilla.com",
+    owner="akommasani@mozilla.com",
```
kik-kik, line 78: Are these actually necessary? Should this information be inherited from `default_args`?

akkomar: Good call, I'll clean that up and hold on with merging until today's runs complete.

PR #1505, opened 2022-05-10
[Bug 1762430] ExternalTaskMarker for graphics_telemetry dashboard

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1762430

This PR is for testing the new `ExternalTaskSensor` and `ExternalTaskMarker` that are built-in to Airflow 2.


dags/graphics_telemetry.py
```diff
@@ -61,13 +62,15 @@
...
-    wait_for_main_ping = ExternalTaskCompletedSensor(
+    wait_for_main_ping = ExternalTaskSensor(
         task_id="wait_for_main_ping",
         external_dag_id="copy_deduplicate",
         external_task_id="copy_deduplicate_main_ping",
         execution_delta=datetime.timedelta(hours=2),
         check_existence=True,
         mode="reschedule",
+        allowed_states=['success'],
+        failed_states=['failed', 'upstream_failed'],
```
scholtzan, line 73: `failed_states` are by default `None`, so they'll need to be listed here explicitly
https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/sensors/external_task/index.html#airflow.sensors.external_task.ExternalTaskSensor

dags/graphics_telemetry.py
```diff
@@ -12,8 +12,9 @@
 import os
 
 from airflow import DAG
+from airflow.utils.task_group import TaskGroup
```
sean-rose, line 15: Don't need `TaskGroup` here.
```suggestion
```

dags/graphics_telemetry.py
```diff
@@ -12,8 +12,9 @@
...
+from airflow.sensors.external_task import ExternalTaskSensor
```
sean-rose, line 16: These should go after `from airflow.providers...`

dags/copy_deduplicate.py
```diff
@@ -105,6 +105,21 @@
         ],
     )
 
+    with TaskGroup(group_id='main_ping_downstream_external') as main_ping_downstream_external:
```
sean-rose, line 108: Could be more succinct.
```suggestion
    with TaskGroup('main_ping_downstream_external') as main_ping_downstream_external:
```

dags/copy_deduplicate.py
```diff
@@ -105,6 +105,21 @@
...
+        )
+
+        ExternalTaskMarker(
+            task_id="graphics_telemetry_graphics_dashboard",
+            external_dag_id="graphics_telemetry",
+            external_task_id="graphics_dashboard",
+        )
+
+    copy_deduplicate_main_ping >> main_ping_downstream_external
+
```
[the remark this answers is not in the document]
sean-rose: I'm curious if/how this will show up in the default tree view, as the [docs](https://airflow.apache.org/docs/apache-airflow/2.1.4/concepts/dags.html#taskgroups) only mention it affecting the _graph_ view.

TaskGroups also automatically prefix the task IDs with `{group_id}.` which is handy in this case I think.

dags/graphics_telemetry.py
```diff
@@ -61,13 +62,15 @@
...
 
-    wait_for_main_ping = ExternalTaskCompletedSensor(
+    wait_for_main_ping = ExternalTaskSensor(
         task_id="wait_for_main_ping",
         external_dag_id="copy_deduplicate",
         external_task_id="copy_deduplicate_main_ping",
         execution_delta=datetime.timedelta(hours=2),
         check_existence=True,
         mode="reschedule",
+        allowed_states=['success'],
```
sean-rose, line 72: The allowed states default to "success", so this isn't technically necessary, but I'm not opposed to being explicit.

sean-rose (approved):
r+wc

dags/copy_deduplicate.py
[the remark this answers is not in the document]
scholtzan: Yeah, I'm also curious. We use `SubDag` in some places which only shows one box in the tree view and lets you expand into the sub tasks. However the docs say `SubDag` is deprecated and to use task groups instead.

dags/graphics_telemetry.py
scholtzan: I'll look into using a global variable once we decide to migrate

jklukas:
Am I correct in understanding that the intention of this PR is to introduce these concepts for a single use case and gain operational knowledge around it before doing a wider refactor to adopt these concepts?

dags/copy_deduplicate.py
```diff
@@ -105,6 +105,21 @@
         ],
     )
 
+    with TaskGroup('main_ping_downstream_external') as main_ping_downstream_external:
```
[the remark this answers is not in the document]
sean-rose: As it stands, yes, "this is the way".  I believe the idea is to have such code be automatically inserted during DAG generation, so there shouldn't be any additional dev code, just output/generated code.  And I think the benefits of being able to easily & accurately re-run tasks and all their downstream dependencies across DAGs will be worth it.

dags/graphics_telemetry.py
sean-rose: Sounds good.  At that point I'd also suggest either using a similar global variable for default allowed states or removing the explicit `allowed_states` parameter (since the default behavior for `allowed_states` is what we want anyway).

scholtzan:
> Am I correct in understanding that the intention of this PR is to introduce these concepts for a single use case and gain operational knowledge around it before doing a wider refactor to adopt these concepts?

Yes, this is just to test this concept on a few DAGs and then we can decide whether to adopt it across all DAGs and implement it in bigquery-etl

dags/copy_deduplicate.py
```diff
@@ -105,6 +105,21 @@
...
+            task_id="graphics_telemetry__graphics_trends",
+            external_dag_id="graphics_telemetry",
+            external_task_id="graphics_trends",
+        )
+
+        ExternalTaskMarker(
+            task_id="graphics_telemetry__graphics_dashboard",
+            external_dag_id="graphics_telemetry",
+            external_task_id="graphics_dashboard",
+        )
```
sean-rose, line 119: I know this already got merged, but I just belatedly realized that the `ExternalTaskMarker`s should probably be targeting the `ExternalTaskSensor`s rather than the actual tasks doing the work, otherwise when those dependent tasks get cleared they'll immediately start running without waiting for the upstream task to complete first.
```suggestion
        ExternalTaskMarker(
            task_id="graphics_telemetry__wait_for_main_ping",
            external_dag_id="graphics_telemetry",
            external_task_id="wait_for_main_ping",
        )
```

PR #1509, opened 2022-05-11
Set execution_date for ExternalTaskMarker

scholtzan:
So the current `ExternalTaskMarker` setup didn't actually work. Recursively clearing the task didn't clear the downstream tasks.
I did some local testing and the issue is that the DAGs have different execution times. To workaround this issue, the `execution_date` parameter needs to be set to the current data. This worked at least for my local setup.


dags/copy_deduplicate.py
```diff
@@ -110,6 +110,7 @@
             task_id="graphics_telemetry__wait_for_main_ping",
             external_dag_id="graphics_telemetry",
             external_task_id="wait_for_main_ping",
+            execution_date="{{ ds }}",
```
sean-rose, line 113: So the difference is this will be just the date string like `2022-05-10` instead of the [default ISO datetime string](https://github.com/apache/airflow/blob/614858fb7d443880451e6111b27fdaf942f563a4/airflow/sensors/external_task.py#L279) like `2022-05-10T00:00:00+00:00`.

I don't understand why that would matter unless the specific scheduled times are involved (e.g. this DAG is scheduled at 01:00 and the `graphics_telemetry` DAG is scheduled at 03:00), in which case the [execution date filter involved](https://github.com/apache/airflow/blob/614858fb7d443880451e6111b27fdaf942f563a4/airflow/models/dag.py#L1291) wouldn't match (and I suppose we'd have to construct the appropriate datetime here to match the target task's 03:00 scheduled time).

However, if you've tested locally and this now works as intended I guess I can live with not understanding why it works. 🙃

sean-rose (approved):

scholtzan:
Got deployed and worked 🎉

dags/copy_deduplicate.py
```diff
@@ -110,6 +110,7 @@
             task_id="graphics_telemetry__wait_for_main_ping",
             external_dag_id="graphics_telemetry",
             external_task_id="wait_for_main_ping",
+            execution_date="{{ (execution_date + macros.timedelta(hours=2)).isoformat() }}",
```
sean-rose, line 113: When we implement this more broadly I'd suggest hard-coding the time part rather than having it be an offset from the upstream DAG's schedule, because this will break if either the upstream DAG's schedule or downstream DAG's schedule changes.

With the external task execution time hard-coded it will only break if the downstream DAG's schedule changes, and it should be easier to implement (just read the downstream schedule rather than having to calculate the delta to use).

scholtzan: Good point!

PR #1510, opened 2022-05-12
[Bug 1762430] Use ExternalTaskMarker and ExternalTaskSensor in favour of ExternalTaskCompletedSensor

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1762430

This PR replaces all usages of `ExternalTaskCompletedSensor` with `ExternalTaskSensor` and adds `ExternalTaskMarker`s to define downstream dependencies.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -87,14 +87,29 @@ with models.DAG(
         resources=resources,
     )
 
-    with TaskGroup('copy_deduplicate_all_downstream_external') as copy_deduplicate_all_downstream_external:
+    with TaskGroup('copy_deduplicate_all_external') as copy_deduplicate_all_external:
         ExternalTaskMarker(
             task_id="bhr_collection__wait_for_bhr_ping",
             external_dag_id="bhr_collection",
             external_task_id="wait_for_bhr_ping",
+            execution_date="{{ execution_date.replace(hour=5, minute=0).isoformat() }}",
         )
 
-        copy_deduplicate_all >> copy_deduplicate_all_downstream_external
+        ExternalTaskMarker(
+            task_id="glam_fenix__wait_for_copy_deduplicate",
+            external_dag_id="glam_fenix",
+            external_task_id="wait_for_copy_deduplicate",
+            execution_date="{{ execution_date.replace(hour=2, minute=0).isoformat() }}",
+        )
+
+        ExternalTaskMarker(
+            task_id="glam_fog__wait_for_copy_deduplicate",
+            external_dag_id="glam_fog",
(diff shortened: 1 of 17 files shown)


dags/copy_deduplicate.py
```diff
@@ -87,6 +87,30 @@
         resources=resources,
     )
 
+    with TaskGroup('copy_deduplicate_all_downstream_external') as copy_deduplicate_all_downstream_external:
+        ExternalTaskMarker(
+            task_id="bhr_collection__wait_for_bhr_ping",
+            external_dag_id="bhr_collection",
+            external_task_id="wait_for_bhr_ping",
+            execution_date="{{ (execution_date + macros.timedelta(hours=4)).isoformat() }}",
```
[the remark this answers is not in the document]
sean-rose: A couple possibilities:
```suggestion
            execution_date="{{ execution_date.replace(hour=5, minute=0).isoformat() }}",
```
```suggestion
            execution_date="{{ ds }}T05:00:00+00:00",
```

dags/copy_deduplicate.py
```diff
@@ -87,6 +87,30 @@
         resources=resources,
     )
 
+    with TaskGroup('copy_deduplicate_all_downstream_external') as copy_deduplicate_all_downstream_external:
```
sean-rose, line 90: I like that the `_downstream_external` suffix is precise/unambiguous, but it is a bit lengthy, and in the default tree view that will push the task-specific part of the task ID further behind the run results grid where it'll get more hidden/cut-off.

Maybe just `_downstream` _or_ `_external` would still be clear enough, while being more succinct?

dags/glam_fenix.py
```diff
@@ -67,153 +71,164 @@
...
+        external_dag_id="copy_deduplicate",
+        external_task_id="copy_deduplicate_all",
+        execution_delta=timedelta(hours=1),
+        check_existence=True,
+        mode="reschedule",
+        allowed_states=ALLOWED_STATES,
+        failed_states=FAILED_STATES,
+        pool="DATA_ENG_EXTERNALTASKSENSOR",
+        email_on_retry=False,
         dag=dag,
```
sean-rose, line 93: Do we still need to pass in the DAG via the `dag` parameters everywhere?

I thought one reason for using the context manager approach is that the DAG then automatically gets applied to everything in that context scope and doesn't need to be passed in.

sean-rose (approved):
r+wc

dags/copy_deduplicate.py
scholtzan: `_external` is probably good enough.

PR #1511, opened 2022-05-16
Wait for search aggregates before starting search alert task

akkomar:
Search alert needs to wait for search aggregates to complete since the script queries this table (https://github.com/mozilla/docker-etl/blob/7c0f31290a0642b785f879bad43234e267221a4e/jobs/search-alert/search_alert/main.py#L53).


ANich (approved):
LGTM

PR #1513, opened 2022-05-31
Bug 1770814 Move adm_export to daily rather than weekly

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1770814

Depends on https://github.com/mozilla/bigquery-etl/pull/2999

### DIFF
--- a/dags/adm_export.py
+++ b/dags/adm_export.py
@@ -11,7 +11,7 @@ from utils.tags import Tag
 
 
 DOCS = """\
-Weekly data exports of contextual services data aggregates to adMarketplace.
+Daily data exports of contextual services data aggregates to adMarketplace.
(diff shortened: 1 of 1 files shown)


dags/adm_export.py
```diff
@@ -34,14 +34,14 @@
 tags = [Tag.ImpactTier.tier_3]
 
 with DAG(
-    dag_name, schedule_interval="0 5 * * MON", doc_md=DOCS, default_args=default_args, tags=tags,
+    dag_name, schedule_interval="0 5 * * *", doc_md=DOCS, default_args=default_args, tags=tags,
```
jklukas, line 37: Changing `MON` to `*` here makes it run every day rather than just on Monday.

PR #1518, opened 2022-06-15
Bug 1774415: upgrade instance type for crash_symbolication dag

willkg:
The crash_symbolication.top_signatures_correlations task is running out
of memory. This upgrades the instance type to make more available.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1774415


jklukas (approved):
LGTM

PR #1520, opened 2022-06-21
Add task markers for bqetl_ DAG  downstream tasks

scholtzan:
Add task markers for downstream dependencies that are part of `bqetl_` DAGs


jklukas (approved):
Is this list going to have to be manually maintained as new tasks are added in bigquery-etl? What is the effect if new tasks are added that aren't represented here?

scholtzan:
> Is this list going to have to be manually maintained as new tasks are added in bigquery-etl?

For now, yes. We should however think about how we can possibly automate some of this.

> What is the effect if new tasks are added that aren't represented here?

The only downside is that new tasks won't get cleared automatically. Also if the execution delta is configured incorrectly or the task is removed in the bqetl DAG, but still here, nothing will break. They will simply get ignored when clearing.

PR #1529, opened 2022-07-22
Use ./bqetl query run-multipart instead of old script

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/3105


akkomar (approved):
👍

PR #1530, opened 2022-07-25
Update Fivetran connector to v1.1.2

alekhyamoz:
The PR would resolve the issue where the syncs are completed quickly after being started by the FivetranOperator. 
More details here: https://github.com/fivetran/airflow-provider-fivetran/issues/60

The Fivetran connector v1.1.2 will help us avoid the task priorities set in private_bqetl_revenue DAG: https://github.com/mozilla/private-bigquery-etl/blob/main/dags/private_bqetl_revenue.py#L152-L154

There would be a subsequent PR to make the necessary changes to the revenue DAG

### DIFF
--- a/dags/operators/backport/fivetran/hook.py
+++ b/dags/operators/backport/fivetran/hook.py
@@ -29,27 +29,27 @@ class FivetranHook(BaseHook):
     :type retry_delay: float
     """
 
-    conn_name_attr = 'fivetran_conn_id'
-    default_conn_name = 'fivetran_default'
-    conn_type = 'fivetran'
-    hook_name = 'Fivetran'
-    api_user_agent = 'airflow_provider_fivetran/1.0.1'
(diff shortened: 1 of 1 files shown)


akkomar (changes requested):
Please rewrite the commit message so it is clear what and why is changing here.
Take a look at this post which contains some great examples and rationale: https://cbea.ms/git-commit/

That might be a question to @haroldwoo, but do we still need to backport this connector by copying the code here? Or maybe we can just install `airflow-provider-fivetran==1.1.2`?

haroldwoo:
Can you create a DSRE Jira for installing the fivetran airflow provider 1.1.2 on the wtmo cluster? I can add this to a list of needs for when I upgrade airflow again.

alekhyamoz:
Created a Jira ticket :https://mozilla-hub.atlassian.net/browse/DSRE-921
>

PR #1533, opened 2022-07-27
[DSRE-882] Create dev GLAM ETL Pipeline via WTMO

haroldwoo:
Do not commit until:
- [x] whitelisting NAT IP to dev cluster
- [x] add secrets to sops/wtmo

- Moves all of the glam/glean import jobs into one file, for easier updating.
- Uses TaskGroups to group Dev and Prod tasks into separate groups (viewable from Graph view)

Before:
Screen Shot 2022-07-26 at 5 26 00 PM
After:
Screen Shot 2022-07-26 at 5 26 57 PM

Note that the graph view isn't perfect, and doesn't display the fan in nature of the glean imports until you click on a taskgroup to expand it:
 
Screen Shot 2022-07-26 at 5 28 07 PM

- Allows devs to change the image version for Dev only, so they can test changes to ETL without hitting prod
- This was set up in a way that adding a Stage env taskgroup is trivial (for the dag code at least, but requires infra changes another db, ip whitelist, sops secrets, etc.)

Open to any suggestions on how to do this differently

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -171,15 +171,19 @@ init_variables() {
     airflow variables set "glean_dictionary_netlify_build_webhook_id" "status/200"
     airflow variables set "lookml_generator_release_str" "v0.0.0"
 
-    airflow variables set "glam_secret__database_url" "glam_secret__database_url"
-    airflow variables set "glam_secret__django_secret_key" "glam_secret__django_secret_key"
-
     airflow variables set "fivetran_acoustic_raw_recipient_export_connector_id" "dummy_connector_id"
 
     airflow variables set "fivetran_acoustic_contact_export_connector_id" "dummy_connector_id"
     airflow variables set "fivetran_acoustic_contact_export_list_id" "00000"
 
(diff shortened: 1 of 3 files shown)


alekhyamoz:
@haroldwoo, this LGTM. But just needed clarification on why dev imports are dependent on the prod ETLs to complete. Can the dev imports happen in parallel while the prod ETL is running.

haroldwoo:
Screen Shot 2022-07-28 at 4 10 02 PM

The tasks should still run in parallel but depending on certain factors(pool, resources, configuration) some tasks may need to wait for others to finish.

alekhyamoz (approved):

akkomar:
@haroldwoo please squash commits before merging

haroldwoo:
Curious, was this inconvenient because you were looking at commit history when investigating glam dev dag tasks failing?

akkomar:
@haroldwoo not in this case but it's generally inconvenient when you're browsing the commit log and see messages like `rm todo comments` :)

PR #1535, opened 2022-08-04
Add note about disabling test suite to burnham DAG

hackebrot:
I will be sending out an announcement email about this today.


akkomar (approved):
LGTM, you can also add a link to a bug if there is one.

hackebrot:
Thanks for the review @akkomar!

We don't have a bug for this, but here's the corresponding change in burnham: https://github.com/mozilla/burnham/pull/447

PR #1539, opened 2022-08-15
Remove outdated forecasts.

skahmann3:
Expected forecast output:
1.  QDAU for desktop
2. DAU for mobile

No CDOU or QCDOU forecasts.

### DIFF
--- a/dags/kpi_forecasting.py
+++ b/dags/kpi_forecasting.py
@@ -2,11 +2,11 @@
 See [kpi-forecasting in the docker-etl repository]
 (https://github.com/mozilla/docker-etl/blob/main/jobs/kpi-forecasting).
 
-This DAG runs the forecast for year-end KPI values for Desktop QCDOU, Mobile CDOU and Pocket.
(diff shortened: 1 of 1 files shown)


skahmann3:
@perrymcmanis144 will you please review? After @irrationalagent updated the forecasting job with the correct forecasts, this set of PRs ([see docker PR too](https://github.com/mozilla/docker-etl/pull/74)) removes the outdated forecasts. QCDOU and CDOU should not be forecasted directly, so this simplifies the DAG to only require `unified_metrics`

dags/kpi_forecasting.py
```diff
@@ -31,26 +31,7 @@
 tags = [Tag.ImpactTier.tier_3]
```
scholtzan, line 31: Increasing priority, since you mention it is at least high priority for some part of the month.
```suggestion
tags = [Tag.ImpactTier.tier_1]
```

scholtzan (approved):
lgtm

PR #1543, opened 2022-08-23
[DSRE-922] Migrate jobs off of gke derived datasets

haroldwoo:
This PR makes all WTMO dags run on the newer GKE Cluster as part of a larger effort to deprecate the moz-fx-data-derived-datasets project. The dags that import utils/gcp's export_to_parquet (running dataproc) in this derived-datasets project will likely need some testing and fixing, but all other dags can be reverted or fixed if they fail (would likely be due to permissions issues, or reliance on artifacts, views, etc. in this derived-datasets project)

~Note that the new GKE cluster is behind a NAT, and if the dags talk to an external service, theres a slight chance they may reject the IP if the dag logic executes anything looking like a DDOS attempt~
Correction: If wanting to send traffic through the NAT we need to set configuration in GCP to do so

TODO
- [x] modify GKEPodOperator defaults 
- [x] modify utils/gcp.py defaults

One off config changes to point to new cluster
- [x] app_store_analytics.py
- [x] burnham.py
- [x] contextual_services_import.py
- [x] glam_fenix.py
- [x] glam_fog.py
- [x] glam_subdags/generate_query.py
- [x] play_store_export.py
- [x] search_alert.py
- [x] socorro_import.py

Permissions
- [x] double check derived_datasets IAM SA permissions with the airflow dataproc prod SA for bq in shared-prod

related infra changes - https://github.com/mozilla-services/cloudops-infra/pull/4400

For utils/gcp.py, make sure export_to_parquet will still work since this was running dataproc in derived-datasets and relies on some gcs buckets to be there, and potentially requires more permissions or refactoring to make work.

Dataproc changes for dag https://github.com/mozilla/telemetry-airflow/blob/main/dags/parquet_export.py
- [x] create equivalent of 'moz-fx-data-derived-datasets-parquet' bucket in airflow_dataproc project
(gcs bucket = airflow-dataproc-bq-parquet-exports)
- [x] modify any downstream consumers of 'moz-fx-data-derived-datasets-parquet' bucket to use new bucket (TAAR/mozagg)
- [x] make sure the google SA used by airflow (google_cloud_airflow_dataproc) has permissions to read shared-prod bq tables telemetry_derived.main_summary_v4, clients_daily_v6
- [x] make sure google SA used by airflow (google_cloud_airflow_dataproc) also has permissions to new bucket created above
- [x] modify utils/gcp.py lines 36/40 to use new bucket
- [x] test GKEPodOperator with new defaults. (Will not test taar/and the export to parquet and fix these as needed after merging this pr)
- [ ] migrate contents of old buckets to new right before merging this PR from (moz-fx-data-derived-datasets-parquet, moz-fx-data-derived-datasets-parquet-tmp) to (airflow-dataproc-bq-parquet-exports, airflow-dataproc-bq-parquet-exports-tmp)
- [x] rm 1 clean_gke_pods dag since they will be redundant


dags/clean_gke_pods.py
```diff
@@ -36,13 +36,16 @@
 
 dag = DAG("clean-gke-pods", default_args=default_args, schedule_interval="@daily", doc_md = docs, tags=tags,)
 
-docker_image='us-west1-docker.pkg.dev/moz-fx-data-airflow-prod-88e0/data-science-artifacts/gke-pod-clean:1.3'
-gke_cluster_name='bq-load-gke-1'
-gke_location='us-central1-a'
+#docker_image = 'us-west1-docker.pkg.dev/moz-fx-data-airflow-prod-88e0/data-science-artifacts/gke-pod-clean:1.3'
```
haroldwoo, line 39: leaving this ref as a note to when we start using artifact registry

whd (approved):

akkomar (approved):

haroldwoo:
Update: 
Copying over the entire buckets would take days (700TB + 256TB for the tmp bucket), so did more digging.

The mozaggregator dags have been rm'd from the prod instance so the 'airflow-dataproc-bq-parquet-exports-tmp' and 'moz-fx-data-derived-datasets-parquet-tmp' buckets do not need to be synced.

Spoke with evgeny and the taar jobs only need 'moz-fx-data-derived-datasets-parquet/clients_daily/v6/submission_date_s3=202209* and 202210* (last month of data) for only the clients_daily parquet files. So this is being copied over now.

PR #1548, opened 2022-08-26
Use copy_deduplicate CLI command

scholtzan:
Depends on https://github.com/mozilla/bigquery-etl/pull/3186

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -294,11 +294,14 @@ def bigquery_etl_copy_deduplicate(
     kwargs["name"] = kwargs.get("name", task_id.replace("_", "-"))
     table_qualifiers = []
     if only_tables:
-        table_qualifiers.append('--only')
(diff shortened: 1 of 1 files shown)


dags/utils/gcp.py
```diff
@@ -294,11 +294,14 @@ def bigquery_etl_copy_deduplicate(
     kwargs["name"] = kwargs.get("name", task_id.replace("_", "-"))
     table_qualifiers = []
     if only_tables:
-        table_qualifiers.append('--only')
-        table_qualifiers += only_tables
+        # bqetl expects multiple args as --only ... --only ...
+        for only_table in only_tables:
+            table_qualifiers.append('--only')
+            table_qualifiers.append(only_table)
```
scholtzan, line 300: Click doesn't allow multiple arguments for parameters, instead the parameter needs to be listed multiple times

akkomar (approved):

PR #1552, opened 2022-09-08
Move probe scraper to airflow-gke-prod

relud:


dags/probe_scraper.py
```diff
@@ -94,12 +100,19 @@
...
+            "hwoo@mozilla.com",
+            "relud@mozilla.com",
+        ],
         env_vars={
             "AWS_ACCESS_KEY_ID": aws_access_key,
             "AWS_SECRET_ACCESS_KEY": aws_secret_key
         },
-        dag=dag)
+        dag=dag,
+        **airflow_gke_prod_kwargs,
```
relud, line 114: I figure it's safe to move the currently-still-prod `probe_scraper` task while we're at it because it isn't accessing any gcp resources (other than gke). Added benefit is that then the "find logs" instructions are consistent between this and the `probe-scraper-moz-central` task

whd (approved):

PR #1553, opened 2022-09-08
Fix gsutil auth for probe_scraper tasks

relud:
this causes gsutil to use credentials from gke, and fixes exceptions like: `ServiceException: 401 Anonymous caller does not have storage.objects.list access to the Google Cloud Storage bucket.`


fbertsch (approved):
LGTM but I might stick it in a variable for future copy/paste errors

kik-kik (approved):

PR #1554, opened 2022-09-08
Fix highmem nodepool label name

relud:
~temporary solution intended to be reverted after issues with highmem-pool-v1 are fixed.~


haroldwoo:
sorry when derived-datasets gke cluster was set up, there was a ton of random stuff i tried to get the node scaling up to work, and other work to get scaling down to work. All of this wasn't codified in tf when we set up the new cluster. going to add the label to the highmem node pool now

haroldwoo (approved):
ill make the label nodepoool:highmem to match derived-datasets cluster

PR #1557, opened 2022-09-12
Change location from which to run data validation job

chelseatroy:
Here is my thinking:

- No matter where we cd into, for some reason CI seems to want the root directory to be usr/app
- That's really fine, because pytest will run properly from there anyway
- I don't want to flatten the file structure artificially to get our scheduling tools to work properly
- So I am now running the job in airflow from outside the src location


perrymcmanis144 (approved):
👍

PR #1561, opened 2022-09-20
Shift taar addons ranking date 1 day back

eu9ene:
It turned out the TAAR lite API has been returning an empty recommendations list for a while. The issue is that [this job](https://github.com/mozilla/taar_gcp_etl/blob/master/taar_etl/taar_lite_guid_ranking.py) writes an empty file to GCS even though it works on testing. I think it's because there is no data in `telemetry.addons` when it's being executed. A shift 1 day back should help. It is not critical to have a number of installations super precise.

### DIFF
--- a/dags/taar_daily.py
+++ b/dags/taar_daily.py
@@ -240,7 +240,7 @@ with DAG("taar_daily", default_args=default_args, schedule_interval="0 4 * * *",
         # This uses a circleci built docker image from github.com/mozilla/taar_gcp_etl
         image=TAAR_ETL_CONTAINER_IMAGE,
         arguments=["-m", "taar_etl.taar_lite_guid_ranking",
-                "--date", "{{ ds }}",
+                "--date", "{{ macros.ds_add(ds, -1) }}",
                 "--prefix", "taar/lite",
                 "--bucket", TAAR_ETL_MODEL_STORAGE_BUCKET],
     )


haroldwoo (approved):

bobsilverberg:
Thanks for the fix @eu9ene. When will this be active in prod?

eu9ene:
> Thanks for the fix @eu9ene. When will this be active in prod?

@bobsilverberg I see that the fix for the file was applied a week ago and it's not empty anymore but the API still returns empty results. I'm investigating why, probably we need to restart it to update the caches.

bobsilverberg:
@eu9ene Any news on this?

eu9ene:
@bobsilverberg Jason restarted the service and cache to make sure the fix was applied. Now we can see that it returns results for some addons but not for others. For example:
```
curl https://prod.taar.prod.dataops.mozgcp.net/taarlite/api/v1/addon_recommendations/@contain-facebook/
{"results": ["private-relay@firefox.com", "woop-NoopscooPsnSXQ@jetpack", "{242af0bb-db11-4734-b7a0-61cb8a9b20fb}", "{aff8af88-06a9-4eee-b383-3af08c47b8c8}"]}
```
It seems like there is another bug in logic, probably related to the whitelist https://github.com/mozilla/taar_gcp_etl/blob/master/taar_etl/taar_amowhitelist.py 

It requires deeper investigation but I already spent two days on this. I'm at Pocket now and have higher priority tasks at the moment, so I'm not sure when I'll be able to take a look again. Probably the addons team or folks from data org who are familiar with the service should take ownership to maintain it properly.

bobsilverberg:
Thanks for the update @eu9ene. I will follow up with the rest of the add-ons team to see where we can go with this.

PR #1563, opened 2022-10-05
Update probe_scraper DAG docs

akkomar:
This adds a direct link to the last 12h of probe_scraper logs in Logs Explorer (retrieved via "Share link option in the UI). In case the link expires I'm keeping the instructions on how to get there from scratch.


fbertsch (approved):
Wow, never been so impressed by a link in my life 💀

PR #1564, opened 2022-10-13
[DSRE-1056] Fix airflow dags after GKE migration

haroldwoo:
This PR fixes the dags:
1. parquet_export - by fixing the project_id references in dags/utils/gcp.py for dataprocOperators to change from derived-datasets to airflow-dataproc
2. socorro_import - by changing usage of BigQueryDeleteTableOperator to GKEPodOperator so that we can piggyback on existing permissions. Previously the airflow-access@derived-datasets SA had broader shared-prod permissions, so that workloads happening on the airflow worker pod could do things like BigQueryDeleteTableOperator to delete a socorro_crash partition. Now that we've switched to GKEPodOperator and the newer GKE airflow-gke cluster, the worker pod will use the airflow-gke-access SA to init a Pod on the airflow-gke cluster, and that Pod's container will have access defined by the default-workloads SA (which can perform shared-prod ops)

### DIFF
--- a/dags/socorro_import.py
+++ b/dags/socorro_import.py
@@ -1,8 +1,5 @@
 from airflow import DAG
 from airflow.operators.subdag_operator import SubDagOperator
-from airflow.providers.google.cloud.operators.bigquery import (
-    BigQueryDeleteTableOperator,
-)
 
 from airflow.providers.google.cloud.operators.cloud_storage_transfer_service import (
     CloudDataTransferServiceS3ToGCSOperator,
@@ -117,7 +114,6 @@ with DAG("socorro_import", default_args=default_args, schedule_interval="@daily"
 
 
     bq_gcp_conn_id = "google_cloud_airflow_gke"
-    bq_project_id = "moz-fx-data-derived-datasets"
 
     dest_s3_key = "s3://telemetry-parquet"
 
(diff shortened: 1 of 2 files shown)


dags/socorro_import.py
```diff
@@ -143,19 +142,20 @@
...
+
+    remove_bq_table_partition = GKEPodOperator(
+        task_id="remove_socorro_crash_bq_table_partition",
         gcp_conn_id=bq_gcp_conn_id,
-        deletion_dataset_table="{}.{}.{}${{{{ds_nodash}}}}".format(
-            "{{ var.value.gcp_shared_prod_project }}", bq_dataset, bq_table_name
-        ),
-        ignore_if_missing=True,
+        name="remove_socorro_crash_bq_table_partition",
+        image="gcr.io/google.com/cloudsdktool/google-cloud-cli:alpine",
```
whd, line 152: I think we should avoid relying on an external image, bqetl has this installed:
```suggestion
        image="gcr.io/moz-fx-data-airflow-prod-88e0/bigquery-etl:latest",
```
Or mirror this image onto our infra.

whd (approved):

dags/socorro_import.py
haroldwoo: testing now

PR #1568, opened 2022-10-24
Upgrade backported fivetran connector to v1.1.2

curtismorales:
This replaces the backported fivetran connector with the newer version 1.1.2, to resolve issues that occur when FivetranOperator syncs finish quickly.

See #1530 -- this is the same as that except that I pulled over the operator and sensor code as well and added a comment to point to the exact commit.

For testing, I've run things locally and confirm that dags still parse; is there anything else I ought to do?


scholtzan (approved):
Looks good to me as a temporary workaround until we can install `airflow-provider-fivetran==1.1.2`.
The best way for testing this is probably to merge this, wait until it got deployed and run one of the DAGs that uses this connector to make sure everything is working properly. If something got broken then we can revert this PR

PR #1571, opened 2022-10-28
Add a probe scraper job to check expiry and fog expiry

relud:
requires https://github.com/mozilla/probe-scraper/pull/511


chutten (approved):
I don't know if I'm the right person to review this as I don't speak airflow... but it looks good to my untrained eye.

relud:
leaving this until monday to merge, just in case 🤞

PR #1572, opened 2022-10-31
Fix branching in probe_scraper dag

relud:
tested locally to confirm branching and trigger rule work as expected

### DIFF
--- a/dags/probe_scraper.py
+++ b/dags/probe_scraper.py
@@ -4,9 +4,11 @@ from datetime import datetime, timedelta
 from airflow import DAG
 from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
 from airflow.models import Variable
+from airflow.operators.dummy import DummyOperator
 from airflow.operators.http_operator import SimpleHttpOperator
 from airflow.operators.python_operator import PythonOperator
-from airflow.sensors.weekday import DayOfWeekSensor
+from airflow.operators.branch_operator import BaseBranchOperator
+from airflow.utils.weekday import WeekDay
 from operators.gcp_container_operator import GKEPodOperator
 from utils.tags import Tag
 
@@ -229,22 +231,15 @@ with DAG('probe_scraper',
         **airflow_gke_prod_kwargs,
     )
 
-    probe_scraper_checks = []
-    for check_name, week_day in [("check-expiry", "Monday"), ("check-fog-expiry", "Wednesday")]:
-        day_check = DayOfWeekSensor(
-            task_id=f"week_day_check_{week_day.lower()}",
-            week_day=week_day,
-            use_task_execution_day=True,
-            dag=dag,
-        )
-        check_task = GKEPodOperator(
+    probe_scraper_checks = [
+        GKEPodOperator(
             task_id=f"probe_scraper_{check_name.replace('-', '_')}",
             name=f"probe-scraper-{check_name}",
             image=probe_scraper_image,
             arguments=(
                 probe_scraper_base_arguments
                 + [
-                    "--{check_name}",
+                    f"--{check_name}",
                     "--bugzilla-api-key={{ var.value.bugzilla_probe_expiry_bot_api_key }}"
                     # don't write any generated files, this job is for emails only
                     "--env=dev",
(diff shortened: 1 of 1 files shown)


dags/probe_scraper.py
```diff
@@ -229,22 +231,15 @@
...
-        check_task = GKEPodOperator(
+    probe_scraper_checks = [
+        GKEPodOperator(
             task_id=f"probe_scraper_{check_name.replace('-', '_')}",
             name=f"probe-scraper-{check_name}",
             image=probe_scraper_image,
             arguments=(
                 probe_scraper_base_arguments
                 + [
-                    "--{check_name}",
```
fbertsch, line 247: Was this not working before?

dags/probe_scraper.py
```diff
@@ -265,17 +260,40 @@
...
         )
-        day_check >> check_task
-        probe_scraper_glean_repositories >> check_task
-        for glean_task in probe_scraper_glean:
-            glean_task >> check_task
-        probe_scraper_checks.append(check_task)
+        for check_name in ("check-expiry", "check-fog-expiry")
+    ]
+    dummy_branch = DummyOperator(
+        task_id=f"dummy_branch",
```
chutten, line 269: f-string with no substitutions?

dags/probe_scraper.py
```diff
@@ -265,17 +260,40 @@
...
+        dag=dag,
+    )
+
+
+    class CheckBranchOperator(BaseBranchOperator):
+        def choose_branch(self, context):
+            """
+            Run an extra branch on the first day of the month
```
chutten, line 277: This comment appears to be incorrect

dags/probe_scraper.py
```diff
@@ -265,17 +260,40 @@
...
+        for check_name in ("check-expiry", "check-fog-expiry")
+    ]
+    dummy_branch = DummyOperator(
+        task_id=f"dummy_branch",
+        dag=dag,
+    )
+
+
+    class CheckBranchOperator(BaseBranchOperator):
+        def choose_branch(self, context):
```
fbertsch, line 275: Some docs here would be helpful, e.g.
```suggestion
    class CheckBranchOperator(BaseBranchOperator):
        """
        The `choose_branch` method returns an array of task_ids
        to be executed. These tasks must be downstream of the
        branch task.
        """
        def choose_branch(self, context):
```

chutten (commented):
Once more I do not speak Airflow (at all, or Python well), but I have found a couple of nits to pick.

dags/probe_scraper.py
```diff
@@ -265,17 +260,40 @@
...
+
+    check_branch = CheckBranchOperator(
+        task_id="probe_scraper_check_branch",
+        # wait for upstream, but ignore upstream failures
+        trigger_rule="all_done",
+        dag=dag,
+    )
+    check_branch >> [*probe_scraper_checks, dummy_branch]
+    check_branch << probe_scraper_glean
+    check_branch << probe_scraper_glean_repositories
```
fbertsch, line 296: What do you think about sticking with `>>` operators? There's already a lot of new context in this change and we may be able to avoid some of it

fbertsch (approved):
r+wc

dags/probe_scraper.py
relud: it was not (was supposed to be an f-string)

dags/probe_scraper.py
relud: as i understand it, you cannot do `[task, ...] >> task`, it has to be `task << [task, ...]` or `task.set_upstream([task, ...]`, and `probe_scraper_glean` is a list

dags/probe_scraper.py
relud:
```suggestion
        task_id="dummy_branch",
```

dags/probe_scraper.py
relud: yes, but I think the docs should go directly on the `choose_branch` method

dags/probe_scraper.py
fbertsch: In retrospect, that makes total sense. You could iterate through the list but that may be ugly. 👍 to leave as-is as well.

dags/probe_scraper.py
relud:
```suggestion
            Return an array of task_ids to be executed. These tasks must be
            downstream of the branch task.
```

dags/probe_scraper.py
relud: turns out i am wrong
```suggestion
    probe_scraper_glean >> check_branch
    probe_scraper_glean_repositories >> check_branch
```

PR #1580, opened 2022-11-10
feat: add backfill dag

mikaeld:
# Description
This PR adds a DAG to initiate backfills. This DAG should support the use cases enabled by the Backfill (alpha) UI plugin. The main advantage is that the backfill job runs in a worker node instead of running in the webserver. Also, logs are accessible at each step, empowering users with useful information for debugging.

### Features
- Input parameters validation
  - **date range validation**: validate `start_date` and `end_date`
  - **regex validation**: validate `task_regex` compiles as a regex pattern
  - **dag name validation**: validate `dag_name` exists in Airflow
- Graph view indicates if backfill job configured as `dry_run` and/or `clear` through branching operators
- Get exact backfill command as output of task `generate_backfill_command`
- Documentation, tips and tricks in DAG doc

image


dags/backfill.py
```diff
@@ -0,0 +1,130 @@
...
+    this workaround will no longer be required in Airflow >=2.2, see link below for future implementation
+    https://airflow.apache.org/docs/apache-airflow/stable/concepts/params.html
+    """
+    params_parsed = ast.literal_eval(string_params)
+    return BackfillParams(**params_parsed)
+
+
+def dry_run_branch_callable(params: str) -> str:
+    backfill_params = __parse_string_params(params)
+    return "dry_run" if backfill_params.dry_run else "real_deal"
```
haroldwoo, line 29: make dry_run and real_deal const vars or enums

dags/backfill.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def clear_branch_callable(params: str) -> str:
+    backfill_params = __parse_string_params(params)
+    return "clear_tasks" if backfill_params.clear else "do_not_clear_tasks"
```
haroldwoo, line 34: same comment here, we want to have vars here to compare instead of hardcoded strings

dags/backfill.py
```diff
@@ -0,0 +1,130 @@
...
+    real_deal_task = DummyOperator(task_id="real_deal")
+
+    clear_branch_task = BranchPythonOperator(
+        task_id="clear_parameter",
+        python_callable=clear_branch_callable,
+        op_kwargs={"params": "{{ dag_run.conf }}"},
+        trigger_rule=TriggerRule.ONE_SUCCESS,
+    )
+
+    clear_tasks_task = DummyOperator(task_id="clear_tasks")
```
haroldwoo, line 111: Just curious, why do we use DummyOperators here?

plugins/backfill/main.py
```diff
@@ -176,3 +162,40 @@ def background(self):
     def history(self):
         """ Outputs recent user request history """
         return flask.Response(file_ops('r'), mimetype='text/txt')
+
+
+def generate_backfill_command(dag_name: str,
```
haroldwoo, line 167: i'm confused as to why you refactored this to use a helper function generate_backfill_command. Originally I thought it was because you wanted to re-use the same function for the backfill dag, but I see that this function is also defined in dags/utils/backfill.

haroldwoo (approved):
lgtm, some minor things.

It is sort of hard to test this locally, I can help with that or you can merge and fix in follow up prs if need be.

plugins/backfill/main.py
mikaeld: So it's testable. I use the helper function in `test_generate_backfill_command` to make sure we match the `BackfillParams` method.

dags/backfill.py
mikaeld: Just for visual representation. This way you can see at a glance if you're doing a dry run and clearing tasks. The parent task is a branch operator, and the dummy tasks displays the parameter value.

plugins/backfill/main.py
mikaeld: Actually, importing the helper function in a test run has the side effect of loading a bunch of flask stuff. Behaviour is inconsistent from local to CI. Now that I've validated the outputs to be the same between the refactored version and the old version, I'll switch to the new version in the backfill plugin.

plugins/backfill/main.py
```diff
@@ -73,37 +77,21 @@ def stream(self):
...
+        backfill_params = BackfillParams(
+            clear=clear,
+            dry_run=dry_run,
+            task_regex=task_regex,
+            dag_name=dag_name,
+            start_date=start_date,
+            end_date=end_date)
 
-        cmd.extend(['-s', str(start_date), '-e', str(end_date), str(dag_name)])
+        cmd = backfill_params.generate_backfill_command()
```
mikaeld, line 95: @haroldwoo do you mind giving this section a second look? Specifically the translation between the string request args and their python type.

haroldwoo: what is the reason for adding the .lower == true

mikaeld: these variables are `str` instead of `bool`. `.lower()` simplifies the number of strings to compare (`"True"`, `"TRUE", `"true"``) and `== "true"` will convert to boolean.

PR #1583, opened 2022-11-14
feat: Adds Airflow DAG for running some of the DIM checks

kik-kik:
# feat: Adds Airflow DAG for running some of the DIM checks

dim link: https://github.com/mozilla/dim

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -95,6 +95,7 @@ init_connections() {
       "google_cloud_derived_datasets"
       "google_cloud_glam_fenix_dev"
       "google_cloud_shared_prod"
+      "google_cloud_gke_sandbox"
     )
     for conn_id in "${gcp_conn[@]}"; do
         airflow connections delete "${conn_id}"
--- a/dags/data_monitoring.py
+++ b/dags/data_monitoring.py
@@ -1,0 +1,45 @@
+from airflow import DAG
+from datetime import datetime, timedelta
+from airflow.models import Variable
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.gcp import bigquery_etl_query, gke_command
+
+DOCS = """\
+
+This DAG is related to data monitoring project and is still under development. Please ignore any alerts related to this DAG 
+"""
+
+default_args = {
+    "owner": "akommasani@mozilla.com",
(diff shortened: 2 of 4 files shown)


dags/data_monitoring.py
```diff
@@ -6,40 +6,63 @@
...
+    schedule_interval="@daily",
+    doc_md=DOCS,
+    tags=TAGS,
+    catchup=True,
+) as dag:
+
+    for target_dataset in TARGET_DATASETS:
+        project_id, dataset, table = target_dataset.split(".")
+
+        task_name = f"rim_run_check_{target_dataset}"
```
alekhyamoz, line 49: nit: dim instead on rim

dags/data_monitoring.py
```diff
@@ -6,40 +6,63 @@
...
+                f"--project_id={project_id}",
+                f"--dataset={dataset}",
+                f"--table={table}"
+                "--date_partition_parameter={{ ds }}"
+            ],
+            env_vars=dict(
+                SLACK_BOT_TOKEN="{{ var.value.dim_slack_secret_token }}"),
+            gcp_conn_id='google_cloud_gke_sandbox',
+            project_id='moz-fx-data-gke-sandbox',
+            cluster_name='akommasani-gke-sandbox',
```
alekhyamoz, line 66: shd this be changed to prod cluster?

alekhyamoz (approved):

dags/data_monitoring.py
kik-kik: well spotted. Thanks!

bin/run
```diff
@@ -95,6 +95,7 @@ init_connections() {
       "google_cloud_derived_datasets"
       "google_cloud_glam_fenix_dev"
       "google_cloud_shared_prod"
+      "google_cloud_gke_sandbox"
```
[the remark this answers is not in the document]
kik-kik: I think this is used by local dev environments only anyways. So I think this should not be a problem.

dags/data_monitoring.py
kik-kik: good point, I didn't notice this. Changed now.

PR #1586, opened 2022-11-16
bug: corrected data param and passing in ds -1

kik-kik:
# bug: corrected data param and passing in ds -1

- cli param for date is now called `date`
- some tables have two day lag, to compensate for this passing ds - 1 day as the `--date` option value.


alekhyamoz (approved):
lgtm

PR #1589, opened 2022-11-22
add hotlist to glam-dev

edugfilho:
glam-dev after this change:
Screen Shot 2022-11-21 at 7 54 47 PM


akkomar (approved):
LGTM.
Does `update_probe_hotlist` depend on `latest_versions`? If not then they could run in parallel with a dummy downstream task.

edugfilho:
> LGTM. Does `update_probe_hotlist` depend on `latest_versions`? If not then they could run in parallel with a dummy downstream task.

Yes during backfilling it pulls the last 10 versions

PR #1594, opened 2022-11-28
DENG-470 adding overwatch DAG

gleonard-m:
- Added overwatch DAG
- Errors can be ignored for now since Overwatch is still in active development.
- 2 Airflow variables are required
1. `overwatch_image_version` set to `latest`.  Can be used to roll back release.
2. `overwatch_slack_token` used to publish reports to a Slack channel

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -182,9 +182,6 @@ init_variables() {
     airflow variables set "Prod_glam_secret__django_secret_key" "Prod_glam_secret__django_secret_key"
     airflow variables set "Dev_glam_project" "Dev_glam_project"
     airflow variables set "Prod_glam_project" "Prod_glam_project"
-    
-    airflow variables set "overwatch_image_version" "overwatch_image_version"
-    airflow variables set "overwatch_slack_token" "overwatch_slack_token"
 }
 
 [ $# -lt 1 ] && usage
--- a/bin/run
+++ b/bin/run
@@ -182,6 +182,9 @@ init_variables() {
     airflow variables set "Prod_glam_secret__django_secret_key" "Prod_glam_secret__django_secret_key"
     airflow variables set "Dev_glam_project" "Dev_glam_project"
     airflow variables set "Prod_glam_project" "Prod_glam_project"
+    
+    airflow variables set "overwatch_image_version" "overwatch_image_version"
+    airflow variables set "overwatch_slack_token" "overwatch_slack_token"
 }
 
 [ $# -lt 1 ] && usage
--- a/dags/overwatch.py
+++ b/dags/overwatch.py
@@ -2,21 +2,17 @@
 Overwatch
 Runs daily at 0700 UTC
 
-Source code is [overwatch repository](https://github.com/mozilla/overwatch-mvp/).
-"""
+Source code is [overwatch-mvp repository](https://github.com/mozilla/overwatch-mvp/).
 
-from datetime import datetime, timedelta
-
-from airflow import DAG
-from airflow.models import Variable
-from operators.gcp_container_operator import GKEPodOperator
-
-docs = """\
 This DAG is executes the Overwatch monitoring system.
 All alerts related to this DAG can be ignored.
 
-(for more info on dim see: https://github.com/mozilla/overwatch-mvp)
 """
+
+from datetime import datetime
+
+from airflow import DAG
+from operators.gcp_container_operator import GKEPodOperator
 
 
 default_args = {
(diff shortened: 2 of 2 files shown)


dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
+"""
+Overwatch
+Runs daily at 0700 UTC
+
+Source code is [overwatch repository](https://github.com/mozilla/overwatch/).
```
kik-kik, line 5: The repo is empty for now, should we just update this to `overwatch-mvp` for now?

gleonard-m: Oops yes - to many repos

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.models import Variable
+from operators.gcp_container_operator import GKEPodOperator
+
+docs = """\
+This DAG is executes the Overwatch monitoring system.
```
kik-kik, line 15: I'd move all of this to the module doc string above and just just the docs in the DAG object to `__doc__`

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+All alerts related to this DAG can be ignored.
+
+(for more info on dim see: https://github.com/mozilla/overwatch-mvp)
```
alekhyamoz, line 18: nit: refers to dim

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+    ],
+    "start_date": datetime(2022, 11, 1),
+    "depends_on_past": False,
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 0,
+}
+
+tags = ["repo/telemetry-airflow", "impact/tier_3", ]
+image = "gcr.io/moz-fx-data-airflow-prod-88e0/overwatch:" + Variable.get("overwatch_image_version")
```
kik-kik, line 36: nit:

```suggestion
image = "gcr.io/moz-fx-data-airflow-prod-88e0/overwatch:{{ var.value.overwatch_image_version }}"
```

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+    schedule_interval="0 7 * * *",
+    doc_md=docs,
+    tags=tags,
+) as dag:
+    run_analysis = GKEPodOperator(
+        task_id="run_analysis",
+        name="run_analysis",
+        image=image,
+        dag=dag,
+        env_vars={"SLACK_BOT_TOKEN": Variable.get("overwatch_slack_token")},
```
kik-kik, line 51:
```suggestion
        env_vars={"SLACK_BOT_TOKEN": "{{ var.value.overwatch_slack_token }}",
```

bin/run
```diff
@@ -182,6 +182,9 @@ init_variables() {
     airflow variables set "Prod_glam_secret__django_secret_key" "Prod_glam_secret__django_secret_key"
     airflow variables set "Dev_glam_project" "Dev_glam_project"
     airflow variables set "Prod_glam_project" "Prod_glam_project"
+    
+    airflow variables set "overwatch_image_version" "overwatch_image_version"
```
alekhyamoz, line 186: Not sure if it is necessary to set the variables in this file

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+Overwatch
+Runs daily at 0700 UTC
+
+Source code is [overwatch repository](https://github.com/mozilla/overwatch/).
+"""
+
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.models import Variable
```
kik-kik, line 11:
```suggestion
```

dags/overwatch.py
kik-kik: If I recall correctly. Every time the DAG is parsed this logic would be executed. If using the macro instead the value should only be retrieved when the task is initiated.

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+    default_args=default_args,
+    schedule_interval="0 7 * * *",
+    doc_md=docs,
+    tags=tags,
+) as dag:
+    run_analysis = GKEPodOperator(
+        task_id="run_analysis",
+        name="run_analysis",
+        image=image,
+        dag=dag,
```
kik-kik, line 50:
```suggestion
```

This is not needed since we're using the DAG context manager.

kik-kik (approved):

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+
+default_args = {
+    "owner": "gleonard@mozilla.com",
+    "email": [
+        "gleonard@mozilla.com",
+        "wichan@mozilla.com",
+    ],
+    "start_date": datetime(2022, 11, 1),
```
kik-kik, line 28: Do you want the DAG to run for each date since the 1st of Nov or it just needs to run for the most recent date?

dags/overwatch.py
kik-kik: https://stackoverflow.com/a/7791595

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
...
```
alekhyamoz, line 28: This will run on entire month's worth of data

dags/overwatch.py
```diff
@@ -0,0 +1,53 @@
+"""
+Overwatch
+Runs daily at 0700 UTC
+
+Source code is [overwatch repository](https://github.com/mozilla/overwatch/).
+"""
+
+from datetime import datetime, timedelta
```
alekhyamoz, line 8: nit: timedelta is not being used

alekhyamoz (approved):

bin/run
gleonard-m: It is needed, `./bin/test-parse` will fail without these added.

dags/overwatch.py
kik-kik: Not unless `catchup=True` is passed into the DAG. Here's an example:
https://github.com/mozilla/telemetry-airflow/blob/main/dags/data_monitoring.py#L51

kik-kik: I just quickly researched this. So apparently in Airflow the default is `True`. However, checking our config it appears we are defaulting to `False`. You can check our Airflow config by going to WTMO -> Admin -> Configuration and searching for `catchup_by_default`

gleonard-m: I've updated the DAG to start `2022-11-23` with `catchup=True` to capture the changes in metrics over the holiday in the US.

bin/run
gleonard-m: Since switching to using macros for the variables I'll remove the changes to `bin/run`

PR #1600, opened 2022-12-06
glam-dev:add missing param to bucket counts to avoid dupe args

edugfilho:
Since we manually add `parameters=("submission_date:DATE:{{ds}}",)`, without `date_partition_parameter=None,`  `bigquery_etl_query` will also add `submission_date` to the query parameters and cause an error. For the same reason the other etl tasks use it


Iinh (approved):

edugfilho:
will merge this after airflow update is finished

PR #1605, opened 2022-12-07
Prevent using cloud function when deploying artifacts

scholtzan:


dags/bqetl_artifact_deployment.py
```diff
@@ -48,8 +48,8 @@
         cmds=["bash", "-c"],
         command=[
             "script/bqetl generate all && "
-            "script/bqetl query schema update '*' &&"
-            "script/bqetl query schema deploy '*' --skip-existing"
+            "script/bqetl query schema update '*' --use-cloud-function=false &&"
+            "script/bqetl query schema deploy '*' --skip-existing --use-cloud-function=false --force --ignore-dryrun-skip"
```
scholtzan, line 52: `--force` will skip checking schema compatibilities which we can skip here to make the deploy go faster: https://github.com/mozilla/bigquery-etl/blob/0873a6029b2d8dca08883161ae6953fda0f3ff1e/bigquery_etl/cli/query.py#L1590
Also since the schema generation ran right before it might be good to assume that schemas are up to date anyway

scholtzan: `--ignore-dryrun-skip` to ensure `baseline_clients_first_seen` etc isn't skipped from deploying here

PR #1628, opened 2022-12-29
Use the `airflow-provider-fivetran` package instead of our backport

sean-rose:
And use the [`airflow-provider-fivetran` package](https://github.com/fivetran/airflow-provider-fivetran/blob/main/fivetran_provider/hooks/fivetran.py)'s new feature to pass the timestamp of the last sync from the Fivetran operator to the Fivetran sensor via XCom so the sensor doesn't miss syncs that finish before it can check.

`airflow-provider-fivetran` 1.1.2 was recently installed ([DSRE-921](https://mozilla-hub.atlassian.net/browse/DSRE-921)), making this possible.

This should wait to be merged until after mozilla/bigquery-etl#3479 and mozilla/private-bigquery-etl#172 are merged, because this removes the backport code.

CCing other DAG owners for visibility: @scholtzan, @ANich, @kik-kik, @spatilmoz


scholtzan (approved):
Let's try it out! Last time when we tried the 1.1.2 version (although we just copied the code files back then), we had to revert the changes because it broke some stuff: https://github.com/mozilla/telemetry-airflow/pull/1570

sean-rose:
I tested a Fivetran task+sensor after deploying this and they ran fine with the XCom value being passed as configured.

PR #1629, opened 2022-12-29
Retry Fivetran syncs when they fail

sean-rose:
We use an [Airflow operator & sensor provided by Fivetran](https://github.com/fivetran/airflow-provider-fivetran), where the operator starts a Fivetran sync and the sensor monitors the sync until it completes.  The problem is when a Fivetran sync fails for some reason the Airflow sensor task will fail and possibly retry, but unless the Airflow operator task is also retried the Fivetran sync won't be restarted so the retried sensor won't find a sync and will ultimately fail again.

This PR adds a retry callback to Fivetran sensor tasks to also clear the associated Fivetran operator task when the sensor is being retried.

This should wait until #1628 is merged (and this PR will be re-targeted to the `main` branch at that point).

CCing other DAG owners for visibility: @scholtzan, @ANich, @kik-kik, @spatilmoz


dags/utils/callbacks.py
```diff
@@ -0,0 +1,26 @@
...
+from sqlalchemy.orm.session import Session
+
+
+@provide_session
+def retry_tasks_callback(context: Context, session: Optional[Session] = None) -> None:
+    """Callback to clear tasks specified by the `retry_tasks` task param.
+    
+    Intended to be used to as an `on_retry_callback` to also retry other tasks when a task fails.
+    """
+    retry_task_ids: list[str] = context['params'].get('retry_tasks', [])
```
sean-rose, line 16: I also considered having the task IDs passed in as a function argument, but that would require slightly more complex usage with something like [`functools.partial()`](https://docs.python.org/3.8/library/functools.html?highlight=partial#functools.partial) (and having to import that to use it), and a side benefit of doing it with task params is those will get captured as task instance metadata so we can see what the `retry_tasks` param values were after the fact.

scholtzan (approved):

dags/utils/callbacks.py
sean-rose: > ... a side benefit of doing it with task params is those will get captured as task instance metadata so we can see what the `retry_tasks` param values were after the fact.

This turned out to be untrue, as Airflow unhelpfully displays the `params` attribute value as a Python object reference (e.g. `<airflow.models.param.ParamsDict object at 0x7f15cfd69cd0>`).

dags/intacct_historical.py
```diff
@@ -75,7 +76,8 @@
             connector_id=connector_id,
             poke_interval=30,
             execution_timeout=timedelta(hours=6),
-            retries=0,
```
kik-kik, line 78: I know I'm a bit late to the party, but would it still make sense to have a fixed number of retries to avoid the task silently failing? Unless I misunderstood the change.

sean-rose: All tasks in this DAG have `retries=2` by default, as configured in the `default_args`.

kik-kik: Thanks for confirming! 🙇

PR #1644, opened 2023-02-15
Add merino jobs to Airflow

quiiver:
Add's the initial DAG for `merino-jobs`. This will be the primary dag definition for running merino jobs, starting with the wikipedia indexer.

Permissions are being discussed in the following ticket: https://mozilla-hub.atlassian.net/browse/DSRE-1198


dags/merino_jobs.py
```diff
@@ -0,0 +1,76 @@
...
+        ],
+
+    )
+
+default_args = {
+    "owner": "wstuckey@mozilla.com",
+    "start_date": datetime.datetime(2019, 7, 25),
+    "email": ["wstuckey@mozilla.com"],
+    "email_on_failure": True,
+    "email_on_retry": True,
```
mtbottle, line 40: I just want to confirm that we're OK with having all failure alerts go to `wstuckey@mozilla.com`? If DISCO team will be owning this, then maybe it would make more sense to have these sent to a mailing list with team members (including Wil). Tagging @ncloudioj for his thoughts on this.

dags/merino_jobs.py
```diff
@@ -0,0 +1,76 @@
...
+    "depends_on_past": False,
+    # If a task fails, retry it once after waiting at least 5 minutes
+    "retries": 1,
+    "retry_delay": datetime.timedelta(minutes=5),
+}
+
+tags = [Tag.ImpactTier.tier_3.__str__()]
+
+# Run weekly on mondays at 5am UTC
```
mtbottle, line 49: nit:
```suggestion
# Run weekly on Mondays at 5am UTC
```

dags/merino_jobs.py
quiiver: i only have myself here in case there's permission errors as we stand this up. in that case i didn't want to spam a big list. we can have a follow up to add a broader team mailing list once it's running.

mtbottle: > we can have a follow up to add a broader team mailing list once it's running.

OK! That sounds good too!

mtbottle (approved):

dags/merino_jobs.py
ncloudioj: We can use `conserv-alert at mozilla.com` for this one.

dags/merino_jobs.py
```diff
@@ -0,0 +1,76 @@
...
+        env_vars=env_vars,
+        email=[
+            "wstuckey@mozilla.com",
+        ],
+
+    )
+
+default_args = {
+    "owner": "wstuckey@mozilla.com",
+    "start_date": datetime.datetime(2019, 7, 25),
```
haroldwoo, line 37: is the intent here to have this dag backpopulate data from 2019?

haroldwoo (approved):

PR #1654, opened 2023-02-22
Add all cli arguments to the index option

quiiver:
I keep messing this up 🤦🏻 So i'm adding all cli arguments


ncloudioj (approved):
Yep, add em' all!

PR #1661, opened 2023-03-07
Delay bqetl artifact timing

fbertsch:
We have three operations that need to happen before this:
1. Probe scraper DAG generates new schemas
2. Those schemas get deployed to BQ
3. Rebuild the bq-etl container to cache those stable schemas

Once that's done, the container we run to update artifacts should be up-to-date with the current state of things.

Associated nightly bq-etl build: https://github.com/mozilla/bigquery-etl/pull/3641

### DIFF
--- a/dags/bqetl_artifact_deployment.py
+++ b/dags/bqetl_artifact_deployment.py
@@ -8,9 +8,10 @@ is successful the job can be considered healthy. This means previous failed DAG 
 can be ignored or marked as successful.
 """
 
+from datetime import datetime, timedelta
+
 from airflow import DAG
 from airflow.utils.trigger_rule import TriggerRule
-from datetime import timedelta, datetime
 from utils.gcp import gke_command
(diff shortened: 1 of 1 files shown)


relud (approved):
r+ because i see nothing wrong with adjusting when this runs.

> 3\. Rebuild the bq-etl container to cache those stable schemas

the commands run in these dags actually regenerate sql where relevant to avoid this issue.

fbertsch:
> r+ because i see nothing wrong with adjusting when this runs.
> 
> > 3. Rebuild the bq-etl container to cache those stable schemas
> 
> the commands run in these dags actually regenerate sql where relevant to avoid this issue.

Hmm, something got cached yesterday though - we had to rebuild the container in order to remove views that had been deleted.

relud:
that's a bug, probably stemming from generating sql in the container that already has generated sql. we probably need to add a command to remove previously generated sql before regenerating

PR #1663, opened 2023-03-09
Change time of Merino job to Tuesdays

mtbottle:
The job keeps failing on Monday because the upstream Wikimedia export is not ready when we try to download the file. So changing the date to Tuesday to improve our chances of download.


quiiver (approved):
lgtm!

PR #1664, opened 2023-03-10
fix: Swap CloudID for URL in the argument list. Move api key to ENV var.

quiiver:
Will do an adhoc run to verify


mtbottle (approved):
LGTM!

PR #1670, opened 2023-03-24
Add production step for wikipedia indexing

chelseatroy:
Merino needs the wikipedia index in both staging and production! This PR updates the Airflow configuration to add the production task.

The flow is that after the wikipedia index file gets copied over to GCS, both staging AND prod indexing can happen in parallel.

If someone close to the project could please check my math on which arguments needed to be switched to what for the staging job relative to prod, I'd be eternally grateful!

### DIFF
--- a/dags/merino_jobs.py
+++ b/dags/merino_jobs.py
@@ -62,7 +62,8 @@ with DAG(
     default_args=default_args,
     tags=tags,
 ) as dag:
-    conn = BaseHook.get_connection("merino_elasticsearch")
+    prod_connection = BaseHook.get_connection("merino_elasticsearch_prod")
+    staging_connection = BaseHook.get_connection("merino_elasticsearch_stage")
 
     wikipedia_indexer_copy_export = merino_job(
         "wikipedia_indexer_copy_export",
@@ -72,14 +73,14 @@ with DAG(
             "--gcs-path",
             "moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports",
             "--gcp-project",
-            "moz-fx-data-shared-prod",
+            "moz-fx-data-shared-nonprod",
         ],
         env_vars={
-            "MERINO_ENV": "production",
+            "MERINO_ENV": "staging",
         },
     )
 
-    wikipedia_indexer_build_index = merino_job(
+    wikipedia_indexer_build_index_for_staging = merino_job(
         "wikipedia_indexer_build_index",
         arguments=[
             "wikipedia-indexer",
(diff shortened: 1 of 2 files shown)


dags/merino_jobs.py
```diff
@@ -72,14 +73,37 @@ def merino_job(
             "--gcs-path",
             "moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports",
             "--gcp-project",
-            "moz-fx-data-shared-prod",
+            "moz-fx-data-shared-nonprod",
```
quiiver, line 76: this should stay using the "`prod`" project as that's what we've defined as a storage for this external data.

dags/merino_jobs.py
```diff
@@ -72,14 +73,37 @@ def merino_job(
...
         ],
         env_vars={
-            "MERINO_ENV": "production",
+            "MERINO_ENV": "staging",
```
quiiver, line 79: I think this can also stay using 'production'. This environment variable tells Merino how to layer its config files ([defined here](https://github.com/mozilla-services/merino-py/tree/main/merino/configs)) which in turn configures the logging to use the mozlog logging format. I believe we use `production` to mean anything running out of dev or testing

dags/merino_jobs.py
```diff
@@ -72,14 +73,37 @@ def merino_job(
...
+            "6600000",  # Estimate of the total number of documents in wikipedia index
+            "--elasticsearch-url",
+            str(staging_connection.host),
+            "--gcs-path",
+            "moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports",
+            "--gcp-project",
+            "moz-fx-data-shared-nonprod",
+        ],
+        env_vars={
+            "MERINO_ENV": "staging",
```
quiiver, line 100: s/staging/production/

dags/merino_jobs.py
```diff
@@ -72,14 +73,37 @@ def merino_job(
...
+            "moz-fx-data-shared-nonprod",
         ],
         env_vars={
-            "MERINO_ENV": "production",
+            "MERINO_ENV": "staging",
+        },
+    )
+
+    wikipedia_indexer_build_index_for_staging = merino_job(
+        "wikipedia_indexer_build_index",
```
quiiver, line 84: I would recommend changing the name of the tasks so they can easily be differentiated in the airflow UI

dags/merino_jobs.py
```diff
@@ -72,14 +73,37 @@ def merino_job(
...
+        env_vars={
+            "MERINO_ENV": "staging",
+            # Using the API key in the argument list leaks the sensitive data into the airflow UI.
+            "MERINO_JOBS__WIKIPEDIA_INDEXER__ES_API_KEY": staging_connection.password,
         },
     )
 
-    wikipedia_indexer_build_index = merino_job(
+    wikipedia_indexer_build_index_for_prod = merino_job(
         "wikipedia_indexer_build_index",
```
quiiver, line 107: Same as previous comment on task names

quiiver (approved):
Excellent! I have a couple small pieces of feedback regarding staging vs prod in various places but otherwise looks good. ty

dags/merino_jobs.py
```diff
@@ -89,7 +90,30 @@ def merino_job(
             "--total-docs",
             "6600000",  # Estimate of the total number of documents in wikipedia index
             "--elasticsearch-url",
-            str(conn.host),
+            str(staging_connection.host),
+            "--gcs-path",
+            "moz-fx-data-prod-external-data/contextual-services/merino-jobs/wikipedia-exports",
+            "--gcp-project",
+            "moz-fx-data-shared-nonprod",
```
quiiver, line 97: sorry i missed this originally. this refers to what gcp project the file was downloaded to. it's should be the same across all tasks eg `moz-fx-data-shared-prod`

PR #1675, opened 2023-03-31
glam-etl test improvements with partitioned tbls

edugfilho:
I'm deploying changes to `glam` etl that have a good chance to improve its execution time according to tests done in `glam-dev`. 
- `clients_histogram_aggregates`: 
  - Replace `submission_date` partitioned table, that holds the 7 previous days of data, with `sample_id` partitioned table, that keeps only the last day, since there's no need to keep more than that amount of data. New table is `clients_histogram_aggregates_v2`
  - Replace a 5-step Subdag with a single step job  - this was tested a few times in dev and ran well.

- `clients_histogram_bucket_counts`:
  - Reduce number of subdag steps from 25 to 10. This job will probably benefit from `clients_histogram_aggregates_v2` partitions since the substeps are based on `sample_id`s. BQ console tests also show around 20% improvements in execution time.

All code changes are being applied temporarily by using the `glam-dev-bigquery-etl` docker image into the steps. This allows for a faster rollback  - just switching back to the original bigquery-etl docker image - in case things go wrong.

When we decide we want to keep the changes I will make a PR to `bigquery-etl` with them so they can be incorporated into the original image, and remove `glam-dev-bigquery-etl` from the steps in this project.


relud (approved):

relud:
r+ with lint

PR #1680, opened 2023-04-04
glam - bucket_counts 10 substeps for another execution

edugfilho:
Upon discussion we want to see this thing running again with 10 substeps for result consistency sake.


Iinh (approved):
lgtm!

PR #1681, opened 2023-04-07
Add partition templating to the task generator for bqetl

quiiver:
This is so we can support monthly table partitions in airflow dags.

See https://github.com/mozilla/bigquery-etl/issues/3303

### DIFF
--- a/dags/utils/gcp.py
+++ b/dags/utils/gcp.py
@@ -139,7 +139,7 @@ def export_to_parquet(
                 if value
             ]
             + (["--static-partitions"] if static_partitions else [])
-            + static_partitions
+            + (static_partitions if static_partitions else [])
             + arguments,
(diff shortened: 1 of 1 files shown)


quiiver:
Dependency for https://github.com/mozilla/bigquery-etl/pull/3710

dags/utils/gcp.py
```diff
@@ -91,6 +89,9 @@ def export_to_parquet(
 
     params = get_dataproc_parameters("google_cloud_airflow_dataproc")
 
+    if arguments is None:
+        arguments = []
+
```
scholtzan, line 94: We'll need to do the same for `static_partitions`, which is concatenated to a list on line 142

scholtzan (approved):
r+wc

PR #1688, opened 2023-04-17
DENG-830 Add domain metadata task to merino_jobs dag.

quiiver:
This adds the domain_metadata cli command as a task to the merino_jobs dag.

Right now i'm keeping this as part of the main merino_jobs dag which is scheduled for every tues @ 5 am UTC. Its probably fine to keep this navigational domain task as part of that but when we have separate scheduling requirements we should think about having multiple dags.

### DIFF
--- a/dags/merino_jobs.py
+++ b/dags/merino_jobs.py
@@ -62,8 +62,8 @@ with DAG(
     default_args=default_args,
     tags=tags,
 ) as dag:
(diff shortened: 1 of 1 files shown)


quiiver:
depends on https://github.com/mozilla-services/merino-py/pull/260 before merging

abhi-agg (approved):
Lgtm 👍🏾

PR #1691, opened 2023-04-20
feat(DENG-794): updated active_users_aggregates project and dim run command

kik-kik:
feat(DENG-794): updated active_users_aggregates project and dim run command


dags/dim_active_users_aggregates.py
```diff
@@ -73,6 +73,7 @@
             name=f"{app_name}_{APP_NAME_POSTFIX}_dim_check",
             image=IMAGE,
             arguments=[
+                "dim",
```
kik-kik, line 76: updated the dim project Dockerfile, so this change is required for the command to work.

alekhyamoz (approved):

PR #1696, opened 2023-05-01
[Bug 1830729] Specify project for view deploys

scholtzan:
To fix view deploys


whd (approved):
I take it this isn't completely redundant with `--target-project` for some reason related to a lack of fully qualified references, though the upstream bug doesn't say anything about this. Ideally we'd fully qualify everything and run airflow jobs from the main projects.

scholtzan:
`--project_id` specifies which project directory views should be published from. (default is shared-prod) `--target_project_id` specifies the GCP project these views should be published in.

whd:
> `--project_id` specifies which project directory views should be published from. (default is shared-prod) `--target_project_id` specifies the GCP project these views should be published in.

I take it the only time these would differ is `mozdata` due to its special handling?

scholtzan:
yep

PR #1700, opened 2023-05-04
Add xcom push/pull to nav suggestions tasks

quiiver:
https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html#how-does-xcom-work
https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html

### DIFF
--- a/dags/merino_jobs.py
+++ b/dags/merino_jobs.py
@@ -172,11 +172,11 @@ with DAG(
         task_id="email_on_domain_success",
         to="wstuckey@mozilla.com",
         subject="Navigational Suggestions Domain Metadata job successful",
-        html_content=f"""
+        html_content="""
         Job completed. Download the new top picks json file on GCS.
         
-        Prod Top Pick JSON file: {{ task_instance.xcom_pull('nav_suggestions_prepare_domain_metadata_prod')['top_pick_url']}}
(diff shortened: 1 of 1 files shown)


quiiver:
https://github.com/mozilla-services/merino-py/pull/285

Allows us to output the public urls of the json file in the success email

dags/merino_jobs.py
```diff
@@ -171,6 +174,9 @@ def merino_job(
         subject="Navigational Suggestions Domain Metadata job successful",
         html_content="""
```
ncloudioj, line 175:
```suggestion
        html_content=f"""
```

dags/merino_jobs.py
```diff
@@ -171,6 +174,9 @@ def merino_job(
...
         Job completed. Download the new top picks json file on GCS.
+        
+        Prod Top Pick JSON file: {{ task_instance.xcom_pull('nav_suggestions_prepare_domain_metadata_prod')['top_pick_url']}}
```
ncloudioj, line 178:
```suggestion
        Prod Top Pick JSON file: { task_instance.xcom_pull("nav_suggestions_prepare_domain_metadata_prod")["top_pick_url"]}
```

ncloudioj (approved):
Looks good. r+wc

dags/merino_jobs.py
quiiver: i committed and then reverted because i remembered that this is a jinja template and not a fstring :)

PR #1701, opened 2023-05-05
Bugzilla-1830328: Resolve blurry top pick favicons

abhi-agg:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1830328

Added cli option "--min-favicon-width" and set it to 52 for both staging and prod environment


quiiver (approved):
this is fine although since it doesn't change from the default defined in the config its not 100% necessary

abhi-agg:
In that case, I will close this PR as it becomes an unnecessary change. Thanks for pointing out 👍🏾

PR #1703, opened 2023-05-16
Fix update_orphaning job

akkomar:
This uses `mozfun.norm.truncate_version` to extract major version number and use it in subsequent version comparisons.

This is a legacy job that used the longitudinal dataset and back in the day was ported to BQ with a shim function. The approach taken here seems to be the simplest one allowing to minimize risk of breaking things. I have tested this in a sandbox cluster and confirmed that as far as I can tell reasonable statistics are now generated.

### DIFF
--- a/jobs/update_orphaning_dashboard_etl.py
+++ b/jobs/update_orphaning_dashboard_etl.py
@@ -129,7 +129,7 @@ def longitudinal_shim_aggregate(date_from, date_to, destination_project, destina
         SELECT
             submission_timestamp,
             client_id,
-            struct(cast(environment.build.version as STRING) as version, environment.build.application_name as application_name) as build,
+            struct(cast(environment.build.version as STRING) as version, mozfun.norm.truncate_version(environment.build.version, 'major') as major_version, environment.build.application_name as application_name) as build,
             struct(struct(environment.settings.update.channel as channel, environment.settings.update.enabled as enabled) as update) as settings,
             payload.info.session_length,
             payload.info.profile_subsession_counter,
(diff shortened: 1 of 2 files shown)


jobs/update_orphaning_dashboard_etl.py
```diff
@@ -373,66 +478,72 @@ def latest_version_on_date(date, major_releases):
...
-    "COUNT(CASE WHEN build.version[0] < '{}.' AND build.version[0] >= '{}.' THEN 1 END) AS versionOutOfDate, "
-    "COUNT(CASE WHEN build.version[0] < '{}.' THEN 1 END) AS versionTooLow, "
-    "COUNT(CASE WHEN build.version[0] > '{}.' THEN 1 END) AS versionTooHigh, "
-    "COUNT(CASE WHEN NOT build.version[0] > '0' THEN 1 END) AS versionMissing "
-"{} "
-"WHERE "
+summary_sql = (
+    ""
+    "SELECT "
+    "COUNT(CASE WHEN build.major_version[0] >= '{}.' AND build.major_version[0] < '{}.' THEN 1 END) AS versionUpToDate, "
```
relud, line 526: should these still be comparing to `'{}.'` now that there definitely won't be a `.`?

relud: and shouldn't these use an integer comparison instead of string?

akkomar: I agree, I think these got automatically casted to integers during my test run.

relud (approved):

PR #1712, opened 2023-05-24
fix: disable catch up and change start dates.

mikaeld:
# Description
As discussed in the workgroup meeting, we need to disable catch up in preparation of the Airflow migration.

Re: `data_monitoring.py`, from  @kik-kik 
> The DAG operates on specific data partitions in target tables, and checks that they adhere to certain rules. So it is important that we guarantee that there is a run for each day

So I've updated the start_date to today.

Side effect of disabling `catchup` is some linting business.


gleonard-m (approved):
Approved but I only looked at the overwatch.py file.

PR #1725, opened 2023-06-09
chore(dags): Remove example DAGs.

mikaeld:
# Description
Clean up part of the DAGs migration. Remove unused DAGs.


haroldwoo (approved):
These were added to give DE examples of their options. But since you're taking over wtmo feel free.

PR #1728, opened 2023-06-13
DSRE 951

whd:
One of ~5. Original approach used sensors but I've switched to adding a sync to the end of each DAG.

I'm testing this out on the stage bucket because prod hasn't been provisioned yet.


dags/update_orphaning_dashboard_etl.py
```diff
@@ -100,3 +101,26 @@
         gcp_conn_id=gcp_conn_id,
     ),
 )
+
+# TODO remove this when the job writes to GCS directly
+gcs_sync = GKEPodOperator(
+    task_id="aws_s3_sync",
```
mikaeld, line 107: nit: consider renaming the step so it's clear we are sync'ing between s3 and GCS (like the variable name)

mikaeld (approved):
approved with small suggestion

PR #1732, opened 2023-06-15
DSRE 951 2

whd:
https://github.com/mozilla/telemetry-airflow/pull/1728 worked so these should as well.

### DIFF
--- a/dags/firefox_public_data_report.py
+++ b/dags/firefox_public_data_report.py
@@ -206,7 +206,7 @@ ensemble_transposer = GKEPodOperator(
 gcs_sync = GKEPodOperator(
     task_id="s3_gcs_sync",
     name="s3-gcs-sync",
-    image="google/cloud-sdk:slim",
+    image="google/cloud-sdk:435.0.1-alpine",
     arguments=[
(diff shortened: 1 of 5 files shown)


dags/firefox_public_data_report.py
```diff
@@ -202,6 +202,29 @@
...
+        "AWS_ACCESS_KEY_ID": aws_access_key,
+        "AWS_SECRET_ACCESS_KEY": aws_secret_key,
+    },
+    dag=dag,
+)
+
+wait_for_main_ping >> hardware_report >> gcs_sync
 wait_for_clients_last_seen >> user_activity >> user_activity_usage_behavior_export
-annotations_export >> ensemble_transposer
+user_activity_usage_behavior_export >> gcs_sync
```
whd, line 228: I fixed the deps here slightly since there's another export: https://prod.telemetry-airflow.prod.dataservices.mozgcp.net/dags/firefox_public_data_report/graph

haroldwoo (approved):
lgtm, only nit would be to consider adding an image tag to pin the google/cloud-sdk:slim's version but I see both ways in the repo

PR #1745, opened 2023-06-28
Test dataset-aware scheduling in bhr

scholtzan:
Testing out the new dataset-aware scheduling. We might want to keep using `ExternalTaskSensor` for the `bqetl_` generated DAGs though, due to the following shortcoming:
* it is not possible to trigger a specific task, but only an entire DAG to run. So if a DAG has multiple tasks that depend on different upstream tasks, all of them have to wait for all upstream dependencies to finish. If an upstream dependency fails then none of the tasks will run (even if only one of them actually depends on the failed upstream dependency). So this could potentially result in more tasks not getting executed that when using ExternalTaskSensor

### DIFF
--- a/dags/bhr_collection.py
+++ b/dags/bhr_collection.py
@@ -13,11 +13,10 @@ is maintained in the mozetl repository.
 import datetime
 
 from airflow import DAG
+from airflow.datasets import Dataset
 from airflow.operators.subdag import SubDagOperator
 from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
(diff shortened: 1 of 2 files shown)


mikaeld:
I agree with the shortcoming you explained. I'd say it could be worth considering revisiting this feature periodically since it is in active development. We'd likely need to break down some of our DAGs into smaller units to fit closer to a dataset-centric structure (dataset as defined in Airflow). However, I doubt it's worth doing this kind of work at moment.

PR #1748, opened 2023-07-05
Temporarily remove logging from data validation DAG:

chelseatroy:
You can see precedent for this solution in [this issue record](https://github.com/apache/airflow/issues/12136) as well as [this similar change](https://github.com/mozilla/telemetry-airflow/pull/1615) in this repo. 

I made a timeboxed effort (2 hours) at a more comprehensive solution, but couldn't do so without monkeypatching pieces of Airflow in a way that felt super brittle.

### DIFF
--- a/dags/search_term_data_validation.py
+++ b/dags/search_term_data_validation.py
@@ -49,5 +49,6 @@ with DAG(
             "moz-fx-data-shared-prod.search_terms_derived.search_term_data_validation_reports_v1",
         ],
         docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/search-term-data-validation_docker_etl:latest",
+        get_logs=False,
         dag=dag,
     )


quiiver:
If the job fails how can we tell what the failure is without the logs in airflow? Is there another method for pulling logs that isn't through the airflow UI?

whd (approved):
> If the job fails how can we tell what the failure is without the logs in airflow? Is there another method for pulling logs that isn't through the airflow UI?

The other method is to look at GKE logs directly. This can be done similarly to the instructions in https://workflow.telemetry.mozilla.org/dags/probe_scraper/grid / https://github.com/mozilla/telemetry-airflow/blob/main/dags/probe_scraper.py#L27 since that DAG has [similar configuration](https://github.com/mozilla/telemetry-airflow/blob/main/dags/probe_scraper.py#L127).

PR #1757, opened 2023-07-12
Update partybal image

scholtzan:
This fixes the population ratio plotting issue.
I had to update the partybal CI and figured we just push images to the experiments project, like all the other tools are doing

### DIFF
--- a/dags/partybal.py
+++ b/dags/partybal.py
@@ -1,4 +1,6 @@
 """
+DAG to schedule generation of results for partybal.
+
 Partybal is an experimental service to visualize experiment results that have been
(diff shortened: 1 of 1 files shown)


dags/partybal.py
```diff
@@ -18,7 +18,7 @@
 
 default_args = {
     "owner": "ascholtz@mozilla.com",
-    "email": ["ascholtz@mozilla.com", "kignasiak@mozilla.com",],
+    "email": ["ascholtz@mozilla.com", "mwilliams@mozilla.com",],
```
scholtzan, line 21: This belongs to you now ;)

mikewilli (approved):

PR #1768, opened 2023-07-28
Glam glean imports stops emailing telemetry-alerts to reduce noise

edugfilho:
Removing telemetry-alerts from alert recipient since it's causing too much noise and we already have a good amount of folks left in the `"email"` cfg to look at it.


alekhyamoz (commented):
LGTM, probably add linh as well?

alekhyamoz (approved):

edugfilho:
> LGTM, probably add linh as well?

Linh's in the line just below when you click to expand :)

PR #1770, opened 2023-07-28
DENG-1026: Update orphaning to write to GCS instead of S3

edugfilho:
Fixes [DENG-1026](https://mozilla-hub.atlassian.net/jira/software/c/projects/DENG/boards/593?modal=detail&selectedIssue=DENG-1026)

Configured with the test GCS bucket for now. Will update with prod bucket in case it works

[DENG-1026]: https://mozilla-hub.atlassian.net/browse/DENG-1026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jobs/update_orphaning_dashboard_etl.py
+++ b/jobs/update_orphaning_dashboard_etl.py
@@ -6,10 +6,10 @@
 # This is a complex, legacy job. Disabling linter will make browsing history of changes easier.
 
 import argparse
-import boto3
 import datetime as dt
 import json
 import re
+from google.cloud import storage
 from urllib.request import urlopen
 
 from pyspark.sql.functions import udf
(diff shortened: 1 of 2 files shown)


jobs/update_orphaning_dashboard_etl.py
```diff
@@ -1188,16 +1186,15 @@ def has_update_apply_failure_mapper(d):
 # Save the output to be uploaded automatically once the job completes.
 # The file will be stored at:
 # * https://analysis-output.telemetry.mozilla.org/app-update/data/out-of-date/FILENAME
```
akkomar, line 1188: This will need to be updated.

akkomar (changes requested):
Thanks for working on this!
LGTM, but tests are failing.

jobs/update_orphaning_dashboard_etl.py
akkomar: I retract that, reading the document attached in the bug it looks like this url is now backed by both s3 and gcs buckets.

akkomar (approved):

PR #1771, opened 2023-08-01
Create custom timetable to run shredder every 4 weeks

relud:
the shredder dag is currently running every 28 days after it was turned on, and sometimes randomly adding new runs based on the current time.

This ensures that the dag will run at UTC midnight every 28 days after start_date.

created following this guide: https://airflow.apache.org/docs/apache-airflow/2.5.3/howto/timetable.html


edugfilho:
Is it worth and feasible to add a test for this?

Marlene-M-Hirose (approved):

relud:
> Is it worth it and feasible to add a test for this?

shouldn't be hard, will do

PR #1797, opened 2023-09-13
If this works, I will be upset

chelseatroy:


fbertsch (approved):
I will be as well

mikaeld:
This is already the default parameter
https://github.com/mozilla/telemetry-airflow/blob/a3464e49fbe748a07b58ab5d9fad36ab0df3569a/dags/utils/gcp.py#L497

PR #1807, opened 2023-09-18
feat: Move utils, operators and glam_subdags out of dags directory

mikaeld:
# Description
Implements [proposal](https://docs.google.com/document/d/1Q9iEyfX2xTrpSJl5nC_K3n1WTRGlG89bK_6WfwwLHew/edit#heading=h.5x0d5h95i329) with options 2B and 4A. This was chosen because it did not require refactoring the DAGs generation code in bigquery-etl. We were able to achieve the desired gitsync structure with minor refactoring in the way we import from `utils`, `operators`, and `glam_subdags` in this repository.

### Refactoring of `utils`, `operators`, and `glam_subdags`
- `utils` moved out of `telemetry-airflow/dags` directory to project root (`telemetry-airflow/utils`)
  - Allows to keep using imports like `from utils.gcp import ...` in the generated DAGs
- `operators` moved out of `telemetry-airflow/dags/` directory to project root (`telemetry-airflow/operators`)
  - Allows to keep using imports like `from operators.bq_sensor import ...` in all DAGs 
- `glam_subdags` moved out of `telemetry-airflow/dags` directory to `telemetry-airflow/utils/glam_subdags`
  - Required refactoring imports to be `from utils.glam_subdags import ...` in the GLAM DAGs.
 
#### As a result, changes in `utils`, `operators`, and `glam_subdags` will now require a new deployment. Only the `dags` directory is sync'ed with git.

Changes to those modules is less frequent than DAG changes, so the loss in functionality should not be too large. We'll need to clearly communicate this change to DE to prevent issues. We also plan to move to [CD for telemetry-airflow](https://docs.google.com/document/d/1v_O0N0UcWt0E7qVWJYmmWBsKiNrEDYmq1T8_JYnk3P0/edit?usp=sharing) in the near future, which will address this issue.

### CI changes
This introduces two new jobs in the publish workflow
- Update the `dags` branch
  - This branch contains only the dag files and a short README, similar to `generated-sql` in bigquery-etl
  - Branch is updated on commits in `main`
- Synchronize telemetry-airflow-dags
  - This job updates the git submodules for `telemetry-airflow` (pointing at the `dags` branch described above) in the DAGs repo [telemetry-airflow-dags](https://github.com/mozilla/telemetry-airflow-dags) 
  
#### Current DAGs synchronization infra with PV
image

#### New DAGs synchronization infra using sidecars
image

# Related PR
https://github.com/mozilla-it/dataservices-infra/pull/81


.circleci/config.yml
```diff
@@ -297,3 +342,22 @@ workflows:
               only: /.*/
             branches:
               only: main
+
+      - sync-dags-branch:
+          name: 🔃 Update DAGs branch
+          filters:
+            tags:
+              only: /.*/
```
haroldwoo, line 350: do we want to do this for all tags? i thought just for main branch only

.circleci/config.yml
```diff
@@ -297,3 +342,22 @@ workflows:
...
+            branches:
+              only: main
+
+      - sync-dags-repo:
+          name: 🔃 Synchronize telemetry-airflow-dags repository
+          filters:
+            tags:
+              only: /.*/
```
haroldwoo, line 358: same comment as above. My assumption is that on commits to main, these branches will be updated so the sidecars can sync them

haroldwoo (approved):
lgtm. some comment about when to run branch ci stuff.
Also bc of all the linting, should we add a lint to the circleci tests?

.circleci/config.yml
mikaeld: Good point, sync'ing on tags is superfluous

.circleci/config.yml
mikaeld: Fixed as well

.circleci/config.yml
```diff
@@ -180,6 +180,51 @@ jobs:
...
+    executor: docker/machine
+    steps:
+      - checkout
+      - add_ssh_keys:
+          fingerprints:
+            - "e4:30:50:41:53:f0:d6:3a:bb:c9:38:54:2d:ca:56:41"
+      - run:
+          name: 🤖 Update DAGs repository
+          command: |
+            ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
```
scholtzan, line 193: Is this needed here as well? https://github.com/mozilla/bigquery-etl/blob/f0c53d7afebaf1e6d8c10754b7dcdffdab0a9a37/.circleci/config.yml#L429-L438

scholtzan (approved):
It looks like most of these changes are related to re-formatting files.

.circleci/config.yml
mikaeld: Not required because we clone telemetry-airflow which is public. The SSH config override was necessary in BQetl because we clone private-bigquery-etl.

PR #1815, opened 2023-09-21
Add to triage notes for probe_scraper DAG

mikewilli:
Adding context for future triage so that issues like this https://bugzilla.mozilla.org/show_bug.cgi?id=1854405 may be resolved more easily.


curtismorales (approved):
Thanks for adding this!

PR #1818, opened 2023-09-22
Update external task markers in copy_deduplicate

scholtzan:
Adding external task markers that are currently missing


fbertsch (approved):
👏

PR #1820, opened 2023-09-25
Set reattach_on_restart=True in copy_deduplicate

akkomar:
We've been recently seeing Airflow declaring running GKE tasks as failed and retrying them. This results in multiple pods running per task run.

This change should stop Airflow from starting duplicate pod instances by reattaching retry runs to existing pods.

`reattach_on_restart` was originally defaulted to `False` in https://github.com/mozilla/telemetry-airflow/pull/1184. It seems fine to enable this now since `run_id` is one of the labels but out of caution this PR does not affect other DAGs.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -223,6 +223,7 @@ with models.DAG(
     )
 
     event_events = bigquery_etl_query(
+        reattach_on_restart=True,
         task_id="event_events",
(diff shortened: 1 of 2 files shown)


akkomar:
cc @haroldwoo who authored https://github.com/mozilla/telemetry-airflow/pull/1820

scholtzan (approved):

haroldwoo (approved):
hopefully this works

PR #1822, opened 2023-09-26
Deduplicate main_use_counter and _remainder outside of copy_deduplicate_all

akkomar:
Recently we've been seeing copy_deduplicate_all run time increasing. This might be related to introduction of split main ping tables that were deduplicated in this task. They should be moved to copy_deduplicate_main_ping to properly reflect dataset dependencies.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -84,6 +84,8 @@ with models.DAG(
         # copy_deduplicate job elsewhere.
         except_tables=[
             "telemetry_live.main_v4",
+            "telemetry_live.main_use_counter_v4",
+            "telemetry_live.main_remainder_v4",
             "telemetry_live.event_v4",
             "telemetry_live.first_shutdown_v4",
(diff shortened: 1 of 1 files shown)


lucia-vargas-a (approved):

dags/copy_deduplicate.py
```diff
@@ -146,7 +148,11 @@
         task_id="copy_deduplicate_main_ping",
         target_project_id="moz-fx-data-shared-prod",
         billing_projects=("moz-fx-data-shared-prod",),
-        only_tables=["telemetry_live.main_v4"],
+        only_tables=[
```
lucia-vargas-a, line 151: The grouping makes sense... Could `copy_deduplicate_main_ping` be timing out now? Is it worth a separate DAG for these two tasks?

akkomar: It shouldn't timeout, these queries are running in chunks (`slices=20`).

PR #1824, opened 2023-09-26
Fixes #1823 - Add dag definition to schedule webcompat-kb job

### LINKED ISSUE #1823: Add dag definition to schedule webcompat-kb job
The link to the job is https://github.com/mozilla/docker-etl/tree/main/jobs/webcompat-kb

ksy36:
For scheduling https://github.com/mozilla/docker-etl/tree/main/jobs/webcompat-kb

### DIFF
--- a/dags/webcompat_kb.py
+++ b/dags/webcompat_kb.py
@@ -1,0 +1,59 @@
+from datetime import datetime
+
+from airflow import DAG
+from utils.gcp import gke_command
+from utils.tags import Tag
+
+DOCS = """
+### Bugzilla to BigQuery import
+
+#### Description
+
+Runs a Docker image that fetches bugzilla bugs from
+Web Compatibility > Knowledge Base component, as well as their core
+bugs dependencies and breakage reports and stores them in BQ.
+
+The container is defined in
+[docker-etl](https://github.com/mozilla/docker-etl/tree/main/jobs/webcompat-kb)
+
+#### Owner
+
+kberezina@mozilla.com
+"""
+
+default_args = {
+    "owner": "kberezina@mozilla.com",
+    "email": ["kberezina@mozilla.com", "webcompat-internal@mozilla.org"],
+    "depends_on_past": False,
+    "start_date": datetime(2023, 9, 26),
+    "email_on_failure": True,
+}
+
+
+tags = [
+    Tag.ImpactTier.tier_2,
+]
+
+every_fifteen_minutes = "*/15 * * * *"
+
+with DAG(
+    "webcompat_kb",
+    default_args=default_args,
+    doc_md=DOCS,
+    schedule_interval=every_fifteen_minutes,
+    tags=tags,
+    catchup=False,
+) as dag:
+    webcompat_kb_import = gke_command(
+        task_id="webcompat_kb",
+        command=[
+            "python",
+            "webcompat_kb/main.py",
+            "--bq_project_id",
+            "moz-fx-dev-dschubert-wckb",
+            "--bq_dataset_id",
+            "webcompat_knowledge_base",
+        ],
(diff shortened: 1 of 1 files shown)


dags/webcompat-kb.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+tags = [
+    Tag.ImpactTier.tier_2,
+]
+
+every_fifteen_minutes = "*/15 * * * *"
+
+with DAG(
+    "webcompat-kb",
```
scholtzan, line 43:
```suggestion
    "webcompat_kb",
```

Also the file name should be renamed to `webcompat_kb.py` for consistency

dags/webcompat-kb.py
```diff
@@ -0,0 +1,62 @@
...
+    tags=tags,
+    catchup=False,
+) as dag:
+    webcompat_kb_import = gke_command(
+        task_id="webcompat_kb",
+        command=[
+            "python",
+            "webcompat_kb/main.py",
+            "--bq_project_id",
+            "moz-fx-dev-dschubert-wckb",
```
scholtzan, line 56: We'll need to make sure that the Airflow service account has access to this project. If you are the owner then you could add it there manually

dags/webcompat-kb.py
ksy36: good catch, thanks!

dags/webcompat_kb.py
```diff
@@ -0,0 +1,62 @@
...
+
+default_args = {
+    "owner": "kberezina@mozilla.com",
+    "email": ["kberezina@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2023, 9, 26),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
```
scholtzan, line 32: it looks like this is running every 15 minutes, so a 30 minute retry delay might not be ideal here. Is a retry actually needed or would the next run 15 minutes later just be enough to process data that wasn't processed by the previous run?

dags/webcompat-kb.py
ksy36: Yeah, I've added a sandbox service account to test dev Airflow instance locally, will add production service account as well

dags/webcompat_kb.py
ksy36: True, next run 15 minutes later would be enough, will remove the retry then :)

dags/webcompat_kb.py
```diff
@@ -0,0 +1,62 @@
...
+[docker-etl](https://github.com/mozilla/docker-etl/tree/main/jobs/webcompat-kb)
+
+#### Owner
+
+kberezina@mozilla.com
+"""
+
+default_args = {
+    "owner": "kberezina@mozilla.com",
+    "email": ["kberezina@mozilla.com"],
```
denschub, line 26: It might be good to add a distribution list here so that we see the errors even if you're on vacation. I'm not sure if you can add `webcompat-internal@mozilla.org` to that, as Google might block emails from senders outside the org - so maybe we'll need a dedicated distribution list for that (and our other jobs that'll follow)

denschub (approved):
Running every 15 minutes is a good idea, and the definition looks fine as I can tell (but of course, I'm not at all an expert here 😄)

scholtzan (approved):

dags/webcompat_kb.py
ksy36: Thanks! I've added the email to the list.

ksy36:
I think this should be good to go. @scholtzan wonder if you could merge it, please?

ksy36:
Thank you!

PR #1841, opened 2023-10-19
Bqetl backfill 4real

fbertsch:
This works the same as running `bqetl query backfill`, but _not locally_!

### DIFF
--- a/dags/bqetl_backfill.py
+++ b/dags/bqetl_backfill.py
@@ -1,0 +1,105 @@
+import datetime
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+from utils.gcp import gke_command
+from utils.tags import Tag
+
+doc_md = """
+# Bqetl Backfill DAG
+
+#### Use with caution: This will overwrite the dates and tables you choose!
+
+#### Some tips/notes:
+* Date formats are 2020-03-01
+* Try with dryrun=True first
+* Read the associated CLI command in bqetl: https://github.com/mozilla/bigquery-etl/blob/7a36416554193c853bb562b34dda6270462acd66/bigquery_etl/cli/query.py#L679
+
+#### Owner
+frank@mozilla.com
+"""
+
+def is_true(bool_str: str) -> bool:
+    if bool_str == 'True':
+        return True
+    if bool_str == 'False':
+        return False
+    return None
+
+
+@dag(
+    dag_id="bqetl_backfill",
+    schedule_interval=None,
+    doc_md=doc_md,
+    catchup=False,
+    start_date=datetime.datetime(2023, 10, 18),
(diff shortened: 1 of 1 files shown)


dags/bqetl_backfill.py
```diff
@@ -0,0 +1,159 @@
...
+        print("To run the command locally, execute the following:\r" + " ".join(cmd))
+
+        return cmd
+
+    run_backfill = gke_command(
+        reattach_on_restart=True,
+        task_id="bqetl_backfill",
+        command="{{ "
+        + "task_instance.xcom_pull(task_ids='generate_backfill_command')"
+        + " }}",
```
ANich, line 151: nit: weird formatting here?
```suggestion
        command="{{ task_instance.xcom_pull(task_ids='generate_backfill_command') }}",
```

ANich (approved):
Overall lgtm, left one nit. Also caveat that some of these details will change once work on managed backfills gets going again

dags/bqetl_backfill.py
fbertsch: I ran it through ruff and black 🤷

fbertsch: Oh! I see. Yeah this is actually important. We don't want Airflow to parameterize that string until runtime of the task itself. This lets us do that (this DAG doesn't work if we make the change you recommended!)

fbertsch: Adding a comment

dags/bqetl_backfill.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+
+        print("To run the command locally, execute the following:\r" + " ".join(cmd))
+
+        return cmd
+
+    run_backfill = gke_command(
+        reattach_on_restart=True,
+        task_id="bqetl_backfill",
+        command="{{ "
```
fbertsch, line 149:
```suggestion
        # This pattern delays parameterizing the string until the task runs
        # See also https://github.com/mozilla/telemetry-airflow/blob/main/utils/gcp.py#L460
        command="{{ "
```

dags/bqetl_backfill.py
ANich: Ahh in that case, would something like this work?
```suggestion
        command=generate_backfill_command.output
```

fbertsch: It might, I can try. I'll also try using the result of `generate_backfill_command`, which looks like may work directly from the [taskflow docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html)

fbertsch: Oh heck yes, it works to just use `generate_backfill_command()`! The taskflow API is way nicer, and that means we can also remove the explicit dependencies below :)

PR #1845, opened 2023-11-02
DSRE-1456 - Archive main_v5 partitions after 37 months

relud:


whd (commented):
Do we care about main clones?

relud:
> Do we care about main clones?

I don't think we do for now. The archive decision brief doesn't mention them, and they are small enough that they are neither expensive to store nor slow in shredder.

whd (approved):

PR #1846, opened 2023-11-07
feat(): Updating DIM related scheduling to match that of aggregates DAG

kik-kik:
feat(): Updating DIM related scheduling to match that of aggregates DAG

### DIFF
--- a/dags/dim_active_users_aggregates.py
+++ b/dags/dim_active_users_aggregates.py
@@ -61,7 +61,7 @@ for platform, config in CONFIGS.items():
     with DAG(
         dag_id,
         default_args=default_args,
-        schedule_interval="30 4 * * *",
+        schedule_interval="15 4 * * *",
         tags=TAGS,
         start_date=datetime(2023, 3, 20),
         doc_md=__doc__,
@@ -72,10 +72,10 @@ for platform, config in CONFIGS.items():
 
     for app_name in config.apps:
         wait_for_aggregates = ExternalTaskSensor(
-            task_id=f"wait_for_{app_name}",
+            task_id=f"wait_for_bqetl_analytics_aggregations_{app_name}",
             external_dag_id="bqetl_analytics_aggregations",
             external_task_id=f"{app_name}_active_users_aggregates",
-            execution_delta=timedelta(seconds=3600),
+            execution_delta=timedelta(seconds=0),
             check_existence=True,
             mode="reschedule",
             allowed_states=ALLOWED_STATES,
--- a/dags/kpi_forecasting.py
+++ b/dags/kpi_forecasting.py
@@ -77,7 +77,7 @@ with DAG(
                 task_id=f"wait_for_{wait_task}",
                 external_dag_id=config.wait_dag,
                 external_task_id=wait_task,
-                execution_delta=timedelta(minutes=30),
+                execution_delta=timedelta(minutes=45),
                 check_existence=True,
                 mode="reschedule",
                 allowed_states=ALLOWED_STATES,
--- a/dags/kpi_forecasting.py
+++ b/dags/kpi_forecasting.py
@@ -11,6 +11,7 @@ from datetime import datetime, timedelta
 
 from airflow import DAG
 from airflow.sensors.external_task import ExternalTaskSensor
+
 from utils.constants import ALLOWED_STATES, FAILED_STATES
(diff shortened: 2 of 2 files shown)


dags/dim_active_users_aggregates.py
```diff
@@ -61,7 +61,7 @@
     with DAG(
         dag_id,
         default_args=default_args,
-        schedule_interval="30 4 * * *",
+        schedule_interval="15 4 * * *",
```
lucia-vargas-a, line 64: The aggregates are now running at 4:15 so the same schedule wouldn't work, that's what is causing the conflict now. Should this run after or before?

lucia-vargas-a: Ref. https://github.com/mozilla/bigquery-etl/blob/65f7f55848e09abf556a68429f3d298cbd961734/dags.yaml#L816

lucia-vargas-a (changes requested):

dags/dim_active_users_aggregates.py
kik-kik: Ehmm, why would it not work? the execution delta for checking for external task completion has been updated to be `0` so the schedules between the two DAGs needs to match as far as I understand:
https://github.com/mozilla/telemetry-airflow/pull/1846/files#diff-8b74dd31a49290bfa268f73dd62b36b88a3d1e09cae8e232ebcdfa377461ef84R78

lucia-vargas-a: Match meaning that both DAGS run at the same time, correct?
In that case wouldn't the checks fail if they run before the aggregates are fully updated, therefore blocking the KPI dashboard?

lucia-vargas-a: Is 4:30 a better time to run the checks if the kpi_dashboard is moved to 4:45?

lucia-vargas-a (commented):
Recommend 4:30 for the dim_checks so they run after the aggregates are updated.

kik-kik:
I don't see how your schedule recommendation makes a difference here. What is important here is that the schedule of this DAG and the schedule of the downstream DAG + the execution delta for sensor task matches which my current understanding is they do. Can we merge so I can test this works we can always change the schedule later if we deem it necessary.

lucia-vargas-a:
> I don't see how your schedule recommendation makes a difference here. What is important here is that the schedule of this DAG and the schedule of the downstream DAG + the execution delta for sensor task matches which my current understanding is they do. Can we merge so I can test this works we can always change the schedule later if we deem it necessary.

This hopefully explains https://github.com/mozilla/telemetry-airflow/pull/1846#discussion_r1385155089
Could we pleease move the scheduler to 4:30 to make sure the checks run after the aggregates?
We also need to rerun the DAG and kpi_forecast for today.

dags/dim_active_users_aggregates.py
kik-kik: The checks will not run before the upstream task as the sensor checks for the task successful completion and only then the checks are executed.

kik-kik:
> > I don't see how your schedule recommendation makes a difference here. What is important here is that the schedule of this DAG and the schedule of the downstream DAG + the execution delta for sensor task matches which my current understanding is they do. Can we merge so I can test this works we can always change the schedule later if we deem it necessary.
> 
> This hopefully explains [#1846 (comment)](https://github.com/mozilla/telemetry-airflow/pull/1846#discussion_r1385155089) Could we pleease move the scheduler to 4:30 to make sure the checks run after the aggregates? We also need to rerun the DAG and kpi_forecast for today.

@lucia-vargas-a As per https://github.com/mozilla/telemetry-airflow/pull/1846#discussion_r1385186604, we don't rely on timing for correct execution but sensors instead. Dim checks will not run until the upstream task successfully completed, only then will the checks run regardless of the schedule.

gkabbz (approved):
I'm approving this as a test but IMO I'll be surprised if it solves the issue. To me it looks like `wait_for_[app_name]` e.g. wait_for_fenix is what's holding everything up and it is waiting for fenix_active_users_aggregates to run. That task successfully ran so I'd expect this to run even with current timing. But given this mostly affects DIM and we are delaying the kpi_forecasting by 15 minutes I don't see a risk. Let's monitor this job though for the next week or two even if this does prove successful.

PR #1852, opened 2023-11-10
DENG-1946. Update dependency of KPI dashboard to bqetl_analytics_aggr…

lucia-vargas-a:
Implements DENG-1946 to update the dependencies of the KPI dashboard.

It's related to [Bug 1858184](https://bugzilla.mozilla.org/show_bug.cgi?id=1858184) where the KPI dashboard is repeatedly on hold.


relud (approved):

kik-kik:
@lucia-vargas-a just a quick question, since this is used for generating a KPI dashboard should we consider at the very least adding some basic ETL checks to the aggregates tables (like make sure data is there, or there are no duplicates) in which case we might actually want to pause the execution of this DAG?

PR #1855, opened 2023-11-16
Add DAG to backfill 3 days of ga_sessions each day

fbertsch:


dags/ga_session_backfill.py
```diff
@@ -0,0 +1,87 @@
...
+            task_id="checks__fail_" + task_id,
+            source_table="ga_sessions_v1",
+            dataset_id="mozilla_org_derived",
+            project_id="moz-fx-data-shared-prod",
+            is_dq_check_fail=True,
+            owner="frank@mozilla.com",
+            email=["frank@mozilla.com", "telemetry-alerts@mozilla.com"],
+            depends_on_past=False,
+            parameters=[
+                "session_date:DATE:{{ macros.ds_add(ds, " + day_offset + ") }}"
```
ANich, line 57: Curious, why isn't `'%Y-%m-%d', '%Y%m%d')` needed here?

ANich: nit: seems like `"macros.ds_add(ds, " + day_offset + ")"` is used in a couple places, you could extract this into a variable above

ANich (approved):

dags/ga_session_backfill.py
fbertsch: This one is the BQ parameter, which expects the dash-separated strings (e.g. `2023-11-01`).

fbertsch: Much more readable, thank you for the suggestion!

fbertsch: Oh I see why you asked! The other parameter was wrong, this one was right. I changed the other one to include dashes, thank you!

PR #1856, opened 2023-11-17
Add dummy task to test bhr gcs bucket data transfer

alekhyamoz:
The PR adds a dummy task that writes s3 (source) data to GCS test bucket. Upon successful validation, the task will be removed.


dags/bhr_collection.py
```diff
@@ -129,4 +129,40 @@
...
+            python_driver_code="https://raw.githubusercontent.com/mozilla/python_mozetl/main/mozetl/bhr_collection/bhr_collection.py",
+            init_actions_uris=[
+                "gs://dataproc-initialization-actions/python/pip-install.sh"
+            ],
+            additional_metadata={"PIP_PACKAGES": "boto3==1.16.20 click==7.1.2"},
+            additional_properties={
+                "spark:spark.jars": "gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar",
+                "spark-env:AWS_ACCESS_KEY_ID": aws_access_key,
+                "spark-env:AWS_SECRET_ACCESS_KEY": aws_secret_key,
+            },
```
sean-rose, line 151: I didn't realize this was run in Spark, and assumed it was run in GKE like the `gcs_sync` task above. I suspect why it hasn't been working is due to GCS authentication/permissions, and either the Google account that the Spark job runs as will need to be granted write access to the destination GCS bucket, or we'll need to specify some Google auth environment variables similar to what's done for AWS here.  I'll be curious to see what the logs show when this tries to run.

sean-rose (approved):

PR #1859, opened 2023-11-22
Changing the cluster for bhr gcs test job

alekhyamoz:
The PR does the following 
1. Changing the cluster for the test job
2. Add missing pip packages
3. remove AWS related spark env variables.


akkomar (approved):
r+
Context: we want the `bhr_collection_gcs_test` task to run a different cluster than `bhr_collection`.

PR #1862, opened 2023-11-28
Change the bucket to gcp for bhr data and remove sync job

alekhyamoz:
Changing the original task to write the BHR data to GCS bucket. 
In the event of failure the PR will be reverted (There is a potential risk of data loss)

### DIFF
--- a/dags/bhr_collection.py
+++ b/dags/bhr_collection.py
@@ -22,7 +22,6 @@ from airflow.operators.subdag import SubDagOperator
 from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
 from airflow.sensors.external_task import ExternalTaskSensor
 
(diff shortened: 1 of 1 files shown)


sean-rose (approved):
Before this is merged mozilla/python_mozetl#393 should be reverted and the [GCS bucket there](https://github.com/mozilla/python_mozetl/blob/a7e2615509b537b2c91933badd04d442cf506c9d/mozetl/bhr_collection/bhr_collection.py#L1138) should be changed back to `moz-fx-data-static-websit-8565-analysis-output`, so if this succeeds the files end up in the same place as before.

alekhyamoz:
@sean-rose here is the PR for changes related to bhr code - https://github.com/mozilla/python_mozetl/pull/395

PR #1865, opened 2023-11-29
[DISCO-2699] Added xcom pull for diff file access

taddes:
This allows us to access the change diff text comparing most recent domain file to previous


ncloudioj (approved):
LGTM, let's merge this after we merge the one in Merino.

PR #1875, opened 2023-12-05
Update socorro_import.py overview comment

willkg:
This updates the link to the Socorro GCP migration bug. It also restructures that paragraph so that the history of S3 buckets, what the DAG does now and what the DAG will do after the Socorro GCP project is done.


fbertsch (approved):
I find this existing paragraph inconsistent, but I'll leave that to a future revision. Thanks for updating the bug!

PR #1876, opened 2023-12-05
Method for sending Slack notifications

scholtzan:


utils/slack.py
ANich, line 1: Looks like there's a builtin we can use to replace this entire module actually:
`from airflow.providers.slack.notifications.slack import send_slack_notification`. Requires creating a Slack API Connection instead of using the `Variable` directly

ANich (approved):
LGTM, but maybe we use the builtin instead?

utils/slack.py
scholtzan: This might be a good question for @mikaeld: we'd like to send Slack notifications when certain Airflow tasks are failing (certain ETL checks for now). We added a `send_slack_notification` function here that uses the Slack token, but would it be preferable to sending messages via a webhook?
I guess we might need to work with the Slack admins to get a webhook added then...

mikaeld: To be tested, but I'd start with the [recommended setup](https://airflow.apache.org/docs/apache-airflow-providers-slack/stable/notifications/slack_notifier_howto_guide.html) using callbacks like @ANich mentions. We'd need to add a `slack_api_default` [connection](https://airflow.apache.org/docs/apache-airflow-providers-slack/stable/connections/slack.html). This is something you can probably request for Slack admins.

mikaeld:
Closing; feel free to reopen if needed.

PR #1879, opened 2023-12-11
Remove adjust_import and android_client_history_sim dags

fbertsch:


kik-kik (approved):
@fbertsch did you check that there are no downstream consumers of these tables by one of the queries inside bqetl or inside looker? Also, should those tables be also dropped inside BQ?

Seeing that you're an owner of all of these I will approve, but I think there are works around a proper deprecation process which maybe we should try and stick to? cc: @alekhyamoz

kik-kik:
Last time I deleted a DAG it sounded like I should have paused it in the UI and deleted it via the UI prior to deleting the code itself. @mikaeld can you confirm these would be the correct steps?

fbertsch:
🙏 thanks for the review @kik-kik 

> @fbertsch did you check that there are no downstream consumers of these tables by one of the queries inside bqetl or inside looker? Also, should those tables be also dropped inside BQ?

The Adjust tables were dropped awhile ago, so I think we're good there. The simulation ones have a single use case (LTV), so we'll keep those tables but remove this job (which we shouldn't need again).

> Seeing that you're an owner of all of these I will approve, but I think there are works around a proper deprecation process which maybe we should try and stick to? cc: @alekhyamoz

Agreed, great point. Given that the former is already deleted and the latter aren't, I don't think there's anything to follow with deprecation! There is one here: https://mozilla-hub.atlassian.net/wiki/spaces/DATA/pages/6848636/Deprecation+Process

mikaeld:
Pause the DAGs prior to merging. It should prevent the issue we had last time @kik-kik.

PR #1885, opened 2023-12-15
fix():  Removing MySQL repo from our Airflow docker image sources due to incorrect GPG Key error

kik-kik:
# fix():  Removing MySQL repo from our Airflow docker image sources due to incorrect GPG Key error

As per @mikaeld suggestion, we're going to try the following temp fix to unblock docker build failures:
https://github.com/apache/airflow/issues/36231#issuecomment-1857798655

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,16 @@ ENV PYTHONPATH="$PYTHONPATH:$PROJECT_DIR"
 ENV AIRFLOW_HOME=$PROJECT_DIR
 
 USER root
+
+# This is a temporary workaround for MySQL invalid GPG Key errors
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,6 +7,7 @@ ENV PYTHONPATH="$PYTHONPATH:$PROJECT_DIR"
 ENV AIRFLOW_HOME=$PROJECT_DIR
 
 USER root
+RUN rm /etc/apt/sources.list.d/mysql.list
```
mikaeld, line 10: Add comment linking to issue and temp fix

mikaeld (approved):

PR #1904, opened 2024-01-26
chore: Delete create_gke_cluster.

mikaeld:
# Description

Looks like this was added a while ago (https://github.com/mozilla/telemetry-airflow/pull/527) and is no longer used


haroldwoo (approved):
rip prio

PR #1905, opened 2024-01-26
chore: delete burnham and bq_sensor operator.

mikaeld:
# Description

Burnham DAG has been deactivated for a while now. `bq_sensor` is not used anywhere else.


fbertsch (approved):
Let's do this ⛵

PR #1915, opened 2024-02-12
Add `bigquery_table_sensor` and `bigquery_table_partition_sensor` functions

sean-rose:
I intend to use these in new bigquery-etl code to generate such sensors based on ETL metadata.

### DIFF
--- a/utils/gcp.py
+++ b/utils/gcp.py
@@ -486,7 +486,7 @@ def bigquery_xcom_query(
 def bigquery_table_sensor(
     task_id: str,
     table_id: str,
-    gcp_conn_id: str = "google_cloud_airflow_gke",
+    gcp_conn_id: str = "google_cloud_shared_prod",
     deferrable: bool = True,
     poke_interval: timedelta | float = 60,
     timeout: timedelta | float | None = None,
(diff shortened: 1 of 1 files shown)


utils/gcp.py
```diff
@@ -477,6 +484,95 @@ def bigquery_xcom_query(
     )
 
 
+def bigquery_table_sensor(
+    task_id: str,
+    table_id: str,
+    gcp_conn_id: str = "google_cloud_airflow_gke",
```
sean-rose, line 490: I'm not sure if using this `google_cloud_airflow_gke` connection is appropriate, since the sensors will run directly in Airflow and not in GKE pods.

Is there a different connection that would be better to use? (we'd want it to have access to all the same BigQuery projects/datasets/tables as the `google_cloud_airflow_gke` connection)

mikaeld: I believe `google_cloud_shared_prod` should be used for this. You are right that `google_cloud_airflow_gke` will not work as it is used only to access the GKE cluster.

mikaeld:
I must strongly advocate against creating abstractions (functions) that simply instantiate a class. This has been a pattern in this repo, and we are now left with the task of cleaning up tens of unused abstractions all over the place.

PR #1917, opened 2024-02-13
DSRE-1526 - Move probe_scraper secrets to GSM

akkomar:


dags/probe_scraper.py
```diff
@@ -309,7 +353,9 @@ def choose_branch(self, context):
         name="schema-generator-1",
         image="mozilla/mozilla-schema-generator:latest",
         env_vars={
-            "MPS_SSH_KEY_BASE64": "{{ var.value.mozilla_pipeline_schemas_secret_git_sshkey_b64 }}",
+            "MPS_SSH_KEY_BASE64": Variable.get(
```
akkomar, line 356: Temporary change to get the value via Airflow UI.

PR #1919, opened 2024-02-14
Remove stray && in bqetl_artifact_deployment

scholtzan:


dags/bqetl_artifact_deployment.py
```diff
@@ -76,7 +76,7 @@
             "script/bqetl query schema update '*' --use-cloud-function=false --ignore-dryrun-skip --project-id=moz-fx-data-marketing-prod && "
             "script/bqetl query schema deploy '*' --use-cloud-function=false --force --ignore-dryrun-skip --project-id=moz-fx-data-marketing-prod"
```
ANich, line 77: Is `marketing-prod` missing an `&&` at the end?

ANich (approved):

dags/bqetl_artifact_deployment.py
scholtzan: ah, good catch

PR #1933, opened 2024-03-01
Bug 1821773 - Update debugging notes for probe scraper and bqetl deploy

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1821773

Fix the regex in the probe scraper notes and update the bqetl_artifact_deployment notes to point to logs explorer because gke has limited retention


kwindau (approved):

kik-kik (approved):
This is great, we should consider coming up with some sort of standard for DAG docs to ensure consistency and that all the needed information / context is available so that everyone can go on holiday without any worries.

PR #1936, opened 2024-03-07
Reduce shredder deletion request interval to 56 days

BenWu:
To see if this has any effect on performance.  I'll start a new task instance when this gets deployed.


akkomar:
I ran two test queries to see if it has any effect:
* `moz-fx-data-bq-batch-prod:US.bquxjob_12082436_18e1a69bd08` - current setting (4 month window)
* `moz-fx-data-bq-batch-prod:US.bquxjob_681d58a1_18e1a694f12` - 2 month window
There's not much difference, in fact the first one consumed 2% less slot time but I wouldn't consider it significant.

@relud do you remember if increasing Shredder's time window to 4 intervals did have a visible effect on its run time?

relud:
shredder's time window never had a significant impact one way or the other in my experience

BenWu:
Alright, I'll close this then

PR #1940, opened 2024-03-08
Revert "Bring back SQL generation in artifact deployments"

scholtzan:
Reverts mozilla/telemetry-airflow#1931
Depends on https://github.com/mozilla/bigquery-etl/pull/5188

Removes SQL generation before artifact deploys


sean-rose (approved):
It'd be helpful to add some explanation to the PR description about why this is being done/possible now versus before when we had to revert the original attempt at this.

PR #1941, opened 2024-03-08
feat: Create shredder backfill dag

BenWu:
Created for the tables for https://mozilla-hub.atlassian.net/browse/DENG-2854 but I went with handling the general case instead of hard-coding the list of tables so we can change it as needed and use it again in the future.

This allows serving of deletion requests outside of the predefined interval in the normal shredder run for a provided list of tables.  This only allows shredding of tables that are already being shredded based on the shredder config.

These are the accepted parameters:
image

Tested using the `moz-fx-data-gke-sandbox` project to verify arguments are correctly passed into the shredder command.

### DIFF
--- a/dags/shredder_backfill.py
+++ b/dags/shredder_backfill.py
@@ -12,14 +12,14 @@ docs = """
 
 #### Description
 
-Manually triggered DAG that handles deletion requests from a specified time period 
+Manually triggered DAG that handles deletion requests from a specified time period
 for a list of given tables.
 
 `target_tables` is a list of tables formatted as `dataset.table_name` with one table per line.
-The moz-fx-data-shared-prod project is assumed because shredder currently only runs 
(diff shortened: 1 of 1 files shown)


dags/shredder_backfill.py
```diff
@@ -0,0 +1,130 @@
...
+            "--end-date={{ params.request_end_date }}",
+            "--start-date={{ params.request_start_date }}",
+            "--no-use-dml",
+            # low parallelism to reduce slot contention with scheduled task
+            "--parallelism=1",
+            "--billing-project=moz-fx-data-bq-batch-prod",
+            "--only",
+        ],
+        # target_tables will be rendered as a python list
+        arguments="{{ params.target_tables }}",
```
BenWu, line 100: I checked that this doesn't allow for command injection by passing in something like `; echo 123`, although this would work for the `cmds` parameter.

edugfilho (approved):

PR #1946, opened 2024-03-11
feat: Add DAG for running opmon backfills

scholtzan:
## Description
Fixes https://github.com/mozilla/opmon/issues/132
This adds a DAG to trigger opmon backfills.


dags/operational_monitoring_backfill.py
```diff
@@ -0,0 +1,80 @@
...
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.tags import Tag
+
+docs = """
+### operational_monitoring_backfill
+
```
lelilia, line 11:
```suggestion
Build from telemetry-airflow repo, [dags/operational_monitoring_backfill.py](https://github.com/mozilla/telemetry-airflow/blob/main/dags/operational_monitoring_backfill.py)
```

In bigquery-etl this step is automated but putting it in manually is quicker for now :)

lelilia (approved):
lgtm

dags/operational_monitoring_backfill.py
```diff
@@ -0,0 +1,80 @@
...
+Triggers backfills for specifc operational monitoring projects.
+
+#### Owner
+
+ascholtz@mozilla.com
+lschiestl@mozilla.com
+"""
+
+tags = [Tag.ImpactTier.tier_3, Tag.Triage.no_triage]
```
lelilia, line 20:
```suggestion
tags = [Tag.ImpactTier.tier_3, Tag.Triage.no_triage, Tag.Repo.airflow]
```
This Tag only exists after #1947 would me merged
Alternatively you could also add "repo/telemetry-airflow" 
Again this is automatically added in bigquery-etl, and here it's only manually but I'd like to have the tags consistent

ncalexan:
This is really cool, I'm excited to try it.  Since this is for human use, can we add a "get out of jail" extra parameters option as well?  Locally, I have been hacking on patches that allow to filter statistics, and some additional ideas that allow to compute window functions over existing statistics.  It would be nice to not require updating this Airflow detail in order to be able to run a "backfill" with extra arguments that is actually _augmenting_ (but not replacing) the existing data.

PR #1953, opened 2024-03-15
fix: Remove dev requirements from Docker image and improve `Makefile`

sean-rose:
## Description

This PR:
- Removes the pytest, Black, and Ruff packages (and some of their dependencies) from the production Docker image, since they're only needed for local development and CI.
- Adds `pip-tools` as a dev requirement and no longer runs `pip-compile` when running `make pip-install-local` so that running `make pip-install-local` in a fresh environment works properly.
- ~Adds `pip-install-local` as a prerequisite for `make` commands that rely on having Python packages installed locally.~
- Updates the `Makefile` to properly configure and document all available commands, and normalize whitespace in strings.

It supersedes #1950.

N/A


Makefile
```diff
@@ -31,8 +34,9 @@ clean: stop
 	rm -rf logs/*
 	if [ -f airflow-worker.pid ]; then rm airflow-worker.pid; fi
 
-pip-install-local: pip-compile
-	pip install -r requirements.txt
+# Install Python packages in the local environment, but ignore pip's "Requirement already satisfied" messages.
+pip-install-local:
+	set -o pipefail; pip install -r requirements.txt -r requirements-dev.txt | { grep -v 'already satisfied' || test $$? = 1; }
```
sean-rose, line 39: It's unfortunate that this kludge is necessary to ignore pip's "_Requirement already satisfied_" messages (see pypa/pip#5900), but IMO it's worth it so packages can be auto-installed when needed without being barraged by hundreds of "_Requirement already satisfied_" messages every time we run commands like `make fixes` or `make test`.

BenWu: It's not pretty but I think it's ok with the comment explaining it.

Is the `test $? = 1` so it still errors out if grep returns >1?  Seems overkill to handle a grep failure that's fine

BenWu (approved):
Quite the rabbit hole resulting from just needing to install pip-tools

Makefile
sean-rose: Correct.  grep will exit with status `1` if it didn't select any lines (I wish there was an option to override that behavior), or `>1` if an actual error occurred.  So we have to do something to prevent the result status of the overall command being non-zero when no packages needed to be installed (i.e. `pip` reported "_Requirement already satisfied_" for everything), because `make` would balk at that if the target was being run as a prerequisite.

A slightly simpler alternative would be `... | { grep -v 'already satisfied' || true; }`, which would ignore any potential grep errors, but since I'm not familiar with what errors those might be I figured I should play it safe.

Makefile
```diff
@@ -1,27 +1,30 @@
...
+	@echo "  gke                Create a sandbox gke cluster for testing"
+	@echo "  clean-gke          Delete the sandbox gke cluster"
+	@echo "  stop               Stops the docker containers"
 
 build:
 	docker-compose build
 
-pip-compile:
-	pip-compile --strip-extras --no-annotate
+pip-compile: pip-install-local
```
mikaeld, line 23: This is creating a bit of chicken and egg situation where you need to install pip-tools via the output requirements, compile requirements.in and then reinstall updated dependencies via the new output requirements. IMO, pip-tools is a project prerequisite and it should be removed from the dev requirements.

A common pattern assumes the required package manager has to be installed outside of the application tooling. E.g. libraries and applications managed with poetry may have poetry listed in the README as a prerequisite, but poetry is never installed via pip within the repository. This should be covered by newcomers reading the README before contributing.

Makefile
sean-rose: > This results in `/bin/sh: 1: set: Illegal option -o pipefail` on certain Linux distributions such as Ubuntu.

Good point. We could avoid that by [configuring](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) `make` to use `bash` instead of `sh`. However...

> I'm personally in favour of not having `pip-install-local` as requirement for `pip-compile`, `fixes`, and `test`. IMO having to install a project dependencies before working on the project is ubiquitous enough that we don't need to provide crutches in the `makefile` to enforce this order of operations. It also adds a few seconds of dependency verification before each test run which is annoying when iterating on tests. It also helps keeping a simpler make command.

Those are fair arguments.  Also, since `pip-install-local` is running `pip install` commands directly in the local system without guardrails like bigquery-etl uses to enforce a virtual environment it's actually a bit dangerous, as people could easily run those other commands without realizing they'll also run `pip install` and thus accidentally modify their global Python installation.  So I'll revert the change to have `pip-install-local` as a prerequisite for other commands.

Longer term, given the mostly containerized nature of the dev workflow here, IMO it would be great to fully containerize the dev workflow so the only thing that would need to be installed locally is Docker.

Makefile
sean-rose: I disagree:
- While Poetry is explicit that [it should be installed separately from the project](https://python-poetry.org/docs/#installation), `pip-tools` is equally explicit that [it should be installed in the same environment as the project](https://pip-tools.readthedocs.io/en/stable/#installation).
- I think it's important that we pin (or at least version constrain) all necessary tooling for consistency/reproducibility (I consider the fact that Poetry doesn't support version constraints on itself a deficiency). For example, CI currently always [installs the latest version of `pip-tools`](https://github.com/mozilla/telemetry-airflow/blob/1c463acbd56354a513f9a3073d421bfb71dc3f42/.circleci/config.yml#L107), and that could start failing at any time if a backwards incompatible version of `pip-tools` is released.
- Given the previous two points, IMO specifying `pip-tools` as a dev requirement pinned to a specific version is reasonable and prudent.

mikaeld: `pip-tools` documentation mentions it needs to be installed in the project's virtual environment, but it does not indicate it needs to be managed within the project requirements. They cite `pip` as an analogous example; pip version is never managed as a project requirement. 

A quick google search lead me to this short [discussion](https://github.com/jazzband/pip-tools/issues/1679) where the accepted answer seems to be "both are fine". I agree that both options are acceptable and will work but I generally prefer the pattern of having package managers handled outside of the project dependencies. However that remains strictly a personal preference influenced by my past experience using poetry extensively.

> that could start failing at any time if a backwards incompatible version of pip-tools is released.

As far as I can recall, this happened once since I've been working on this repo and the fix was upgrading my local pip-tools version to regenerate requirements. The actual issue was in the requirements.txt file header changing due to a new `pip-tools` version. This should not happen anymore now that we use `--strip-extras --no-annotate` to generate requirements.

With all that said, I'm not opposed to the proposed changes.

sean-rose: True, the `pip-tools` docs don't recommend managing it with the project requirements, but I think anything that materially impacts the project's virtual environment should be version controlled if possible (in both the source control and package management senses).  An example of why this is a concern is that installing the latest `pip-tools` would also install the `build`, `click`, `packaging`, `pyproject-hooks`, and `wheel` packages while ignoring the project's own constraints on any of those packages, which could have adverse effects.

And `pip` is a bit of a special case since it's distributed with Python itself.

Looking for other examples of how `pip-tools` is managed here at Mozilla, I see `pip-tools` is included in the requirements of a number of other projects like `bigquery-etl`, `gcp-ingestion-edge`, `pontoon`, `treeherder`, etc. (though IMO in some of those cases it should be moved to a separate dev requirements rather be included in the main requirements).

mikaeld (approved):

scholtzan:
I think this might be breaking the DAG validation in bigquery-etl: https://app.circleci.com/pipelines/github/mozilla/bigquery-etl/31518/workflows/016c0e56-8658-4052-b21c-e53214fa6dc0/jobs/342609

scholtzan:
bigquery-etl should be fixed via https://github.com/mozilla/bigquery-etl/pull/5238

sean-rose:
Good catch @scholtzan.  I've submitted https://github.com/mozilla/private-bigquery-etl/pull/365 following your lead on https://github.com/mozilla/bigquery-etl/pull/5238.

PR #1955, opened 2024-03-18
fix: 1952 Acoustic Suppression List DENG 3082 date format

lelilia:
## Description

This PR should really really fix the broken date format ... 
The attepted fix in #1952 added the quotation marks ending up with `"2023-07-10" 06:33 PM` datetimes 🙃 

This PR also hard codes the starting date for the suppression list DAG since we don't need to backfill from last year but will merge this with a one time download for historic data

* DENG-3082

### DIFF
--- a/dags/fivetran_acoustic.py
+++ b/dags/fivetran_acoustic.py
@@ -169,7 +169,7 @@ REPORTS_CONFIG = {
     "suppression_export": {
         "request_template": """
     <!-- https://developer.goacoustic.com/acoustic-campaign/reference/export-from-a-database -->
-    <!-- date_format: 07/25/2011 12:12:11 (time is optional) -->
+    <!-- date_format: 2011-12-20 12:12 PM (time is optional) -->
     <Envelope>
     <Body>
         <ExportList>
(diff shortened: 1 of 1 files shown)


dags/fivetran_acoustic.py
```diff
@@ -180,10 +180,10 @@ def _generate_acoustic_report(
         <EXPORT_COLUMNS>
             {columns}
         </EXPORT_COLUMNS>
-        <DATE_START>{date_start}</DATE_START>
+        <DATE_START>datetime(2024, 3, 13)</DATE_START>
```
kik-kik, line 183: This would mean that each report generation would be expanding on each run and contain data starting March 13th 2024. The main issue with this is that over time this will mean that the report generated is larger and larger on each run and eventually we might start potentially experiencing issues due to the file size and Fivetran will continuously reprocess the same rows (incurring unnecessary additional cost).

If we're happy to continue processing one day worth of changes captured by acoustic then we should just update the start_date of the DAG itself.

kik-kik: https://github.com/mozilla/telemetry-airflow/blob/b8a686ba0963385fb4a653d3a0bee793226562d5/dags/fivetran_acoustic.py#L210

lelilia: good catch thank you
I now change the start_date for all DAGs ... if that is not a good idea I will just move the new DAG out of this dag generation thing that is so annoying me right now

kik-kik: If you prefer, you can also just create a new DAG definition file with just the suppression export.

lelilia: I prefer the least amount of effort right now :)

kik-kik (approved):

PR #1976, opened 2024-04-29
Add DAP Collector dag for PPA Dev

bbirdsong2:
## Description

This PR adds a new PPA Dev DAP Collector DAG to run the new docker-etl job here: https://github.com/mozilla/docker-etl/pull/189. This DAG will be used to test out the ads use case in the DAP infrastructure.

* [AE-109](https://mozilla-hub.atlassian.net/browse/AE-109)


dags/dap_collector_ppa_dev.py
```diff
@@ -0,0 +1,78 @@
...
+    Tag.Triage.no_triage,
+]
+
+
+with DAG(
+    "dap_collector_ppa_dev",
+    default_args=default_args,
+    doc_md=DOCS,
+    schedule_interval="*/15 * * * *",
+    tags=tags,
```
BenWu, line 57: You might want to explicitly set `catchup=False` so it doesn't trigger all runs since the start_date/last run if it's not necessary.  False might be the default value but I can't remember
```suggestion
    tags=tags,
    catchup=False,
```

dags/dap_collector_ppa_dev.py
```diff
@@ -0,0 +1,78 @@
...
+
+default_args = {
+    "owner": "bbirdsong@mozilla.com",
+    "email": ["ads-eng@mozilla.com", "bbirdsong@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2024, 4, 19),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=15),
```
BenWu, line 39: Retry delay is the same as schedule interval so you'll have two running concurrently if it fails.  Would it make sense to wait for the next run instead of retrying?  Or is the continuity important?

dags/dap_collector_ppa_dev.py
bbirdsong2: Yeah, I can add that.

dags/dap_collector_ppa_dev.py
bbirdsong2: That makes sense. I copied this from another job and did not evaluate the retry logic.

If a job fails, I do not actually think a retry or subsequent run would work, because the job depends on the {{ ts }} variable of the dag run, and only goes back every 15 minutes.

I think what I can do is remove the retry and run the script manually if a job fails.

dmueller (approved):

gleonard-m (approved):

quiiver (approved):

PR #1979, opened 2024-05-02
feat: DENG-3096 Create 3 new cloudflare DAGs

kwindau:
## Description
This PR adds 3 new DAGs that call the open-source Cloudflare API to pull data on device usage, browser usage, and OS usage in different locations and by different cuts.

* DENG-3096

### DIFF
--- a/dags/cloudflare_browser_usage.py
+++ b/dags/cloudflare_browser_usage.py
@@ -19,6 +19,8 @@ from utils.tags import Tag
 
 # Load auth token
 auth_token = Variable.get("cloudflare_auth_token")
+if auth_token is None:
+    auth_token = "AUTH_TOKEN"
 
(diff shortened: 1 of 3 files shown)


Marlene-M-Hirose (approved):

sean-rose:
Is there a reason this is being done using DAG code rather than in `query.py` ETLs?

kwindau:
> Is there a reason this is being done using DAG code rather than in `query.py` ETLs?

@sean-rose Good question - To be honest, I didn't know the query.py option even existed until this morning when Kik & Leli mentioned it to me, and I already had majority of this code written.  For future things I will definitely use that instead.  For now, I don't expect any other dependencies off of these DAGs so I think it's OK to leave it here for now, but can always move it later to bqetl if needed

PR #1987, opened 2024-05-16
feat(bqetl_backfill): adding new scheduling overrides json object param to bqetl_backfill

braunreyes:
## Description
This PR fully completes the intention of implementing https://github.com/mozilla/bigquery-etl/issues/5364

The cli command the bqetl_backfill executes can now override all necessary scheduling params.
We need to update the dag to pass along the new argument

## Related Tickets & Documents
* [AD-243](https://mozilla-hub.atlassian.net/browse/AD-243)

[AD-243]: https://mozilla-hub.atlassian.net/browse/AD-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1989, opened 2024-05-21
feat(bqetl_backfill): applying shelex quote removal for scheduling overrides handling that was missing from previous PR

braunreyes:
## Description

This is a a follow up PR for https://github.com/mozilla/telemetry-airflow/pull/1987

There was a change to remove shelex quote that did not get pushed, which was used locally to create the log file sample.


chelseybeck (approved):
🙌

PR #1992, opened 2024-05-21
DENG-3851 Invited managed backfills watchers to slack channel

wwyc:
## Description
This PR adds the step in managed backfills to invite watchers to slack channel prior sending notifications.  It was discovered that users tagged in notifications were not alerted if they were not in the channel.

* DENG-3851

### DIFF
--- a/dags/bqetl_backfill_initiate.py
+++ b/dags/bqetl_backfill_initiate.py
@@ -9,6 +9,9 @@ from airflow.providers.slack.operators.slack import SlackAPIPostOperator
 from operators.gcp_container_operator import GKEPodOperator
 from utils.tags import Tag
 
+import os
+from slack_sdk import WebClient
+
 AUTOMATION_SLACK_CHANNEL = "#dataops-alerts"
 SLACK_CONNECTION_ID = "overwatch_slack"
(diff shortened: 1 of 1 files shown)


dags/bqetl_backfill_initiate.py
```diff
@@ -44,8 +47,22 @@
...
     def initiate_backfill(backfill):
+        def invite_watchers_to_slack_channel(entry):
+            token = os.environ["SLACK_BOT_TOKEN"]
+            slack_client = WebClient(token=token)
+
+            for watcher in entry["watchers"]:
+                user_id = slack_client.users_lookupByEmail(email=watcher)
+
+                slack_client.channels_invite(
+                    channel="testing_channel_id", # is it safe to share channel_id publicly?
```
ANich, line 58: Some of the APIs accept a channel name as well I think

dags/bqetl_backfill_initiate.py
```diff
@@ -44,8 +47,22 @@
 
     @task_group
     def initiate_backfill(backfill):
+        def invite_watchers_to_slack_channel(entry):
```
ANich, line 50: If it doesn't work as is, I think it might end up having to be a task 🤔 We could also encapsulate this in a subclass of `SlackAPIOperator`

PR #1995, opened 2024-05-24
fix: Increase bhr_collection spark memory

BenWu:
## Description

Continuation of:
https://bugzilla.mozilla.org/show_bug.cgi?id=1892258
https://bugzilla.mozilla.org/show_bug.cgi?id=1897484

Switching vm types to n2 which have slightly more memory per core, and increasing the spark memory allocation


fbertsch (approved):
Amazing that Spark logs w.r.t. memory still haven't improved. You could consider switching to `n4` machines.

PR #1999, opened 2024-05-27
feat:Add POC dags for triggering data observability checks

scholtzan:
Adds two POC DAGs for triggering checks in BigEye and Acryl. Depends on https://github.com/mozilla/bigquery-etl/pull/5663


constraints.txt
```diff
@@ -413,7 +413,7 @@ jsonschema==4.21.1
 junit-xml==1.9
 jupyter_client==8.6.0
 jupyter_core==5.7.1
-keyring==24.3.0
+keyring==24.3.1
```
scholtzan, line 416: @mikaeld `bigeye-airflow` requires `keyring>=24.3.1`. Is it ok to bump the version here?

mikaeld: It's preferable to not modify the constraints file. However, you can add dependency overrides in a separate `requirements-override.txt` (or similarly named file) and install them after the existing pip install step in the docker image and in the local environment. [I have done something similar in the past](https://github.com/mozilla/telemetry-airflow/pull/1792/files), except I specified the requirements to override directly in the Dockerfile because it was a very temporary fix for an upstream issue.

scholtzan: I added a `requirements-override.txt`. Let me know if the setup looks reasonable

requirements-override.txt
```diff
@@ -0,0 +1,3 @@
+bigeye-airflow==0.1.34
+bigeye-sdk==0.4.81
+keyring==24.3.1
```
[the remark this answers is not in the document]
scholtzan: Right, it find the right `keyring` version.

```suggestion

```

mikaeld (approved):

dags/bigeye_poc.py
```diff
@@ -0,0 +1,79 @@
...
+    latest_versions = bigquery_etl_query(
+        reattach_on_restart=True,
+        task_id="latest_versions",
+        destination_table="latest_versions_v1",
+        dataset_id="telemetry_derived",
+        sql_file_path="sql/data-observability-dev/telemetry/latest_versions/query.sql",
+        project_id="data-observability-dev",
+        date_partition_parameter=None,
+        arguments=("--replace",),
+        dag=dag,
```
kik-kik, line 46:
```suggestion
```

not needed since we're using the DAG object as a context manager.

kik-kik (approved):

PR #2019, opened 2024-06-12
chore(glam): don't export to GCS since aggregates are read from BQ

edugfilho:
## Description

This PR removes the export to gcs  step from the `glam` dag, since GLAM reads aggregates from BQ and no longer needs that data on GCS. 

See also https://github.com/mozilla/bigquery-etl/pull/5780 which removes export code from `bqetl`


dags/glam.py
```diff
@@ -369,19 +370,7 @@
...
-            default_args,
-            dag.schedule_interval,
-            table_project_id,
-            billing_project_id,
-            fully_qualified_dataset,
-            dataset_id,
-        ),
-        task_id="extracts",
-    )
+    extracts_per_channel = EmptyOperator(task_id="extracts")
```
BenWu, line 373: What's the purpose of the empty operator instead of updating the task dependencies?

BenWu (approved):

PR #2026, opened 2024-06-18
chore(bqetl_artifact_deployment): unschedule DAG

whd:
## Description

Requires https://github.com/mozilla-services/cloudops-infra/pull/5736. We should then be running bqetl artifact deploys after every merge to main and every new schemas deploy. Note that on days where there are no schemas updates/bqetl merges, this DAG won't run, so if there is some reason to keep a fallback scheduled this shouldn't land as-is.

* DENG-2963


scholtzan (approved):

whd:
This apparently won't be particularly useful unless we add a flag for doing sql generation again: https://mozilla-hub.atlassian.net/browse/DENG-2963?focusedCommentId=897578

PR #2031, opened 2024-06-21
feat: Separate DAG for generating Looker artifacts

scholtzan:
## Description

This moves the lookml generation out of the `probe_scraper` DAG into a new `looker` DAG. Having a separate DAG will enable us to trigger lookml generation from the CI of the lookml-generator repository.

### DIFF
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -1,0 +1,156 @@
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.models import Variable
+from airflow.providers.cncf.kubernetes.secret import Secret
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.tags import Tag
+
+DOCS = """\
+# Looker
+
+*Triage notes*
+
+As long as the most recent DAG run is successful this job can be considered healthy.
+In such case, past DAG failures can be ignored.
+
+- Failure of the `lookml_generator` task may be due to a new Glean app or changes to
+`custom-namespaces.yaml`. In these cases, the task will have created a PR in
+[looker-spoke-default](https://github.com/mozilla/looker-spoke-default)
+with the title "Auto-push from LookML Generator". These PRs will need to be merged
+and the task re-run.
+"""
+
+DEFAULT_LOOKML_GENERATOR_IMAGE_VERSION = "v1.17.0"
+
+
+default_args = {
+    "owner": "ascholtz@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2024, 5, 21),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 2,
+    "retry_delay": timedelta(minutes=30),
+}
+
+tags = [Tag.ImpactTier.tier_1]
+
+looker_repos_secret_git_ssh_key_b64 = Secret(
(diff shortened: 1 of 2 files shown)


dags/looker.py
```diff
@@ -0,0 +1,155 @@
...
+    secret="airflow-gke-secrets",
+    key="probe_scraper_secret__dataops_looker_github_secret_access_token",
+)
+
+
+with DAG(
+    "looker",
+    doc_md=DOCS,
+    default_args=default_args,
+    schedule_interval=None,
```
scholtzan, line 82: This DAG doesn't need to be scheduled. It's either triggered from the `probe_scraper` DAG after schema deploys or via CI

dags/probe_scraper.py
```diff
@@ -394,71 +352,11 @@ def choose_branch(self, context):
...
-        secrets=[
-            looker_repos_secret_git_ssh_key_b64,
-            looker_api_client_id_staging,
-            looker_api_client_secret_staging,
-            dataops_looker_github_secret_access_token,
-        ],
-        **airflow_gke_prod_kwargs,
-    )
-
-    delay_python_task >> lookml_generator_staging >> lookml_generator_prod
```
BenWu, line 460: Should this dependency be maintained?  Looks like `trigger_looker` just runs from the start

dags/probe_scraper.py
```diff
@@ -394,71 +352,11 @@ def choose_branch(self, context):
...
-        secrets=[
-            looker_repos_secret_git_ssh_key_b64,
-            looker_api_client_id_prod,
-            looker_api_client_secret_prod,
-            dataops_looker_github_secret_access_token,
-        ],
-        **airflow_gke_prod_kwargs,
+    # trigger lookml generation
+    trigger_looker = TriggerDagRunOperator(
+        task_id="lookml_generator_staging", trigger_dag_id="looker"
```
BenWu, line 357: I think the task id should be updated as well
```suggestion
        task_id="trigger_lookml_generator", trigger_dag_id="looker"
```

dags/probe_scraper.py
```diff
@@ -478,4 +376,4 @@ def choose_branch(self, context):
         dag=dag,
     )
 
-    lookml_generator_prod >> glean_dictionary_netlify_build
+    trigger_looker >> glean_dictionary_netlify_build
```
BenWu, line 379: Does `TriggerDagRunOperator` succeed immediately after triggering or does it wait for the run status?

kwindau (approved):

dags/probe_scraper.py
scholtzan: There is a `wait_for_completion` parameter which would wait for the DAG run. I'll add it.
Although Glean dictionary doesn't really depend on anything that is in Looker

BenWu: Found it https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/operators/trigger_dagrun/index.html

There's a `wait_for_completion` param that defaults to false.  Seems like we want it to be true?  Assuming glean dictionary is actually dependent on lookml, I'm not sure

scholtzan: I added it to keep it consistent with the current state

PR #2032, opened 2024-06-24
fix: microsoft store DAG update when DAG runs to day before actual date

Marlene-M-Hirose:
* DENG-2631

### DIFF
--- a/dags/microsoft_store.py
+++ b/dags/microsoft_store.py
@@ -80,7 +80,7 @@ with DAG(
             arguments=[
                 "python",
                 f"sql/moz-fx-data-shared-prod/microsoft_derived/{table}_v1/query.py",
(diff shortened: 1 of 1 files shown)


dags/microsoft_store.py
```diff
@@ -80,7 +80,7 @@
             arguments=[
                 "python",
                 f"sql/moz-fx-data-shared-prod/microsoft_derived/{table}_v1/query.py",
-                "--date={{ ds }}",
+                "--date='{{ macros.ds_add(ds, -1) }}",
```
BenWu, line 83: missing a quote
```suggestion
                "--date='{{ macros.ds_add(ds, -1) }}'",
```

Marlene-M-Hirose: Thank you~!

PR #2034, opened 2024-06-26
feat(dags): Add DAP collector for PPA Prod

dmueller:
### Description

https://mozilla-hub.atlassian.net/browse/AE-364

related PRs:
* create tables & give airflow permission https://github.com/mozilla-it/webservices-infra/pull/2247
* create prod job that this will run https://github.com/mozilla/docker-etl/pull/217

this airflow job will run the new prod etl.

### Testing

https://workflow.telemetry.mozilla.org/dags/dap_collector_ppa_dev/grid is now working (had to fix a permissions issue, the airflow & docker etl were both OK)

### DIFF
--- a/dags/dap_collector_ppa_prod.py
+++ b/dags/dap_collector_ppa_prod.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 from airflow import DAG
+from airflow.providers.cncf.kubernetes.secret import Secret
 
 from operators.gcp_container_operator import GKEPodOperator
 from utils.tags import Tag
@@ -47,6 +48,19 @@ tags = [
     Tag.Triage.no_triage,
 ]
 
+hpke_private_key = Secret(
+    deploy_type="env",
+    deploy_target="HPKE_PRIVATE_KEY",
+    secret="airflow-gke-secrets",
+    key="DAP_PPA_PROD_HPKE_PRIVATE_KEY",
+)
+
+auth_token = Secret(
+    deploy_type="env",
+    deploy_target="AUTH_TOKEN",
+    secret="airflow-gke-secrets",
+    key="DAP_PPA_PROD_AUTH_TOKEN",
+)
 
 with DAG(
     "dap_collector_ppa_prod",
@@ -62,8 +76,6 @@ with DAG(
             "python",
             "dap_collector_ppa_prod/main.py",
             "--date={{ ts }}",
-            "--auth-token={{ var.value.dap_ppa_prod_auth_token }}",
-            "--hpke-private-key={{ var.value.dap_ppa_prod_hpke_private_key }}",
             "--task-config-url={{ var.value.dap_ppa_prod_task_config_url }}",
             "--ad-config-url={{ var.value.dap_ppa_prod_ad_config_url }}",
             "--project",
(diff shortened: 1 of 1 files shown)


dags/dap_collector_ppa_prod.py
```diff
@@ -0,0 +1,78 @@
...
+    Tag.ImpactTier.tier_3,
+    Tag.Triage.no_triage,
+]
+
+
+with DAG(
+    "dap_collector_ppa_prod",
+    default_args=default_args,
+    doc_md=DOCS,
+    schedule_interval="*/15 * * * *",
```
dmueller, line 55: can we adjust this schedule interval freely? we're still figuring out the aggregation window on the dap side so may want to run this less frequently.

braunreyes (commented):
LGTM

dags/dap_collector_ppa_prod.py
```diff
@@ -0,0 +1,78 @@
...
+    tags=tags,
+    catchup=False,
+) as dag:
+    dap_collector = GKEPodOperator(
+        task_id="dap_collector_ppa_prod",
+        arguments=[
+            "python",
+            "dap_collector_ppa_prod/main.py",
+            "--date={{ ts }}",
+            "--auth-token={{ var.value.dap_ppa_prod_auth_token }}",
```
quiiver, line 65: are these secret values? if so, we should use secret storage for them otherwise they will be exposed in the dag config

dags/dap_collector_ppa_prod.py
```diff
@@ -0,0 +1,78 @@
...
+            "--hpke-private-key={{ var.value.dap_ppa_prod_hpke_private_key }}",
```
quiiver, line 66: same question about secrets as the token

dmueller: yeah the private key and auth token are credentials for the dap collector to be able to read data from the leader.

Reading through https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27922811/WTMO+Developer+Guide#WTMODeveloperGuide-SecretManagement it looks like it should be a simple adjustment in this file to read from secrets. I'll need help setting the values in the secrets though

dags/dap_collector_ppa_prod.py
```diff
@@ -0,0 +1,94 @@
...
+    secret="airflow-gke-secrets",
+    key="DAP_PPA_PROD_HPKE_PRIVATE_KEY",
+)
+
+auth_token = Secret(
+    deploy_type="env",
+    deploy_target="AUTH_TOKEN",
+    secret="airflow-gke-secrets",
+    key="DAP_PPA_PROD_AUTH_TOKEN",
+)
```
dmueller, line 63: Switched the secrets, but may need to tweak the `key` params here once the values are configured in GSM.

dags/dap_collector_ppa_prod.py
dmueller: I updated the secrets to be read from GSM, will coordinate with you to ensure the names of the values match. Also updated the docker-etl job to get these from environment variables instead of arguments https://github.com/mozilla/docker-etl/pull/219

quiiver (approved):

PR #2035, opened 2024-06-26
fix(bhr): Update bhr_collection_child worker count and sampling

BenWu:
## Description

bhr_collection_child is failing because of occasionally higher volume of bhr pings so slightly decreasing the sampling ratio and increase workers/mem pool

* https://bugzilla.mozilla.org/show_bug.cgi?id=1904695

### DIFF
--- a/dags/bhr_collection.py
+++ b/dags/bhr_collection.py
@@ -90,7 +90,8 @@ with DAG(
             "spark:spark.executor.memory": "20g",
         },
         "idle_delete_ttl": 14400,
-        "num_workers": 6,
+        # dataproc does not support all GCE instance types, e.g. newer ones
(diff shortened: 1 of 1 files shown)


dags/bhr_collection.py
```diff
@@ -90,7 +90,8 @@
             "spark:spark.executor.memory": "20g",
         },
         "idle_delete_ttl": 14400,
-        "num_workers": 6,
+        # dataproc does not support all GCE instance types, e.g. newer ones
+        # https://cloud.google.com/compute/docs/general-purpose-machines
```
BenWu, line 94: Adding comment as a reminder since I almost changed these to n4

edugfilho (approved):
This makes sense to me, although I lack proper context to speak to the actual values that were chosen.

PR #2039, opened 2024-06-28
feat: create two new dags for the Firefox-CI ETL

ahal:
* [Bug 1904928](https://bugzilla.mozilla.org/show_bug.cgi?id=1904928) (tracking)
* [DENG-4195](https://mozilla-hub.atlassian.net/browse/DENG-4195) (secrets)

[DENG-4195]: https://mozilla-hub.atlassian.net/browse/DENG-4195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/fxci_metric_export.py
+++ b/dags/fxci_metric_export.py
@@ -1,0 +1,72 @@
+"""
+Exports Firefox-CI worker data from the Google Cloud Monitoring to BigQuery.
+
+The container is defined in [fxci-etl](https://github.com/mozilla-releng/fxci-etl).
+"""
+
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.providers.cncf.kubernetes.secret import Secret
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.tags import Tag
+
+default_args = {
+    "owner": "ahalberstadt@mozilla.com",
+    "depends_on_past": False,
+    "start_date": datetime(2024, 7, 8),
+    "email_on_failure": True,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
+}
+
+tags = [Tag.ImpactTier.tier_3]
+
+env_vars = {
+    "FXCI_ETL_BIGQUERY_PROJECT": "moz-fx-data-shared-prod",
+    "FXCI_ETL_BIGQUERY_DATASET": "fxci",
+    "FXCI_ETL_STORAGE_PROJECT": "moz-fx-dev-releng",
+    "FXCI_ETL_STORAGE_BUCKET": "fxci-etl",
+}
+
+secrets = [
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_STORAGE_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci_etl_secret__gcp-credentials",
+    ),
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_MONITORING_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci_etl_secret__gcp-credentials",
+    ),
+]
+
+with DAG(
+    "fxci_metric_export",
+    default_args=default_args,
+    doc_md=__doc__,
+    schedule_interval="@daily",
+    tags=tags,
+) as dag:
+    fxci_metric_export = GKEPodOperator(
+        task_id="fxci_metric_export",
+        arguments=[
+            "fxci-etl",
+            "metric",
+            "export",
+            "-vv",
+        ],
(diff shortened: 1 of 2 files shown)


ahal:
This blocked both on https://mozilla-hub.atlassian.net/browse/DENG-4195 and https://github.com/mozilla/docker-etl/pull/218, but I think it should be pretty close to the final form.

dags/fxci_metric_export.py
```diff
@@ -0,0 +1,80 @@
...
+    "FXCI_ETL_STORAGE_BUCKET": "fxci-etl",
+}
+
+secrets = [
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_BIGQUERY_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci-etl-gcp-credentials",
+    ),
```
ahal, line 42: Because I have this set to the `moz-fx-shared-prod.fxci` dataset, I'm a bit unsure if I still need a service account with ability to write data to this dataset, or if the Airflow account will implicitly have permission..

akkomar: Airflow jobs can write to `shared-prod` by default, no additional configuration is needed.

akkomar (commented):
Can we combine these two tasks in a single DAG?

akkomar:
> > Can we combine these two tasks in a single DAG?
> 
> Can I do that but still run them on separate schedules? The pulse queue one might need to run more frequently than the monitoring one.. but I'm not 100% sure yet. It depends how big the queue is allowed to grow before it becomes a problem for our RabbitMQ instance.
> 
> Or did you just mean you want both DAGs defined in the same file?

I meant a single DAG because both tasks are scheduled daily and there's some code duplication. If they need to run on different schedules then it's indeed better to have them in two separate DAGs.

ahal:
Thanks, I moved one of the DAGs to run every 6 hours (as I worry that once a day may not consume pulse events fast enough).

Though if it turns out not to be a problem, I may end up putting these on the same schedule and merging into a single DAG.

ahal:
Actually before this lands, I need to get the tables created.. Not sure if it's best to:

1. Do this manually
2. Do it in the code (assuming the Airflow SA has permission to create tables)
3. Update some terraform / configuration somewhere

dags/fxci_metric_export.py
```diff
@@ -0,0 +1,74 @@
...
+    "FXCI_ETL_STORAGE_PROJECT": "moz-fx-dev-releng",
+    "FXCI_ETL_STORAGE_BUCKET": "fxci-etl",
+}
+
+secrets = [
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_STORAGE_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci-etl-gcp-credentials",
```
akkomar, line 41: For consistency with other DAGs:
```suggestion
        key="fxci_etl_secret__gcp-credentials",
```

dags/fxci_metric_export.py
```diff
@@ -0,0 +1,74 @@
...
+    ),
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_MONITORING_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci-etl-gcp-credentials",
```
akkomar, line 47:
```suggestion
        key="fxci_etl_secret__gcp-credentials",
```

dags/fxci_pulse_export.py
```diff
@@ -0,0 +1,79 @@
...
+    "FXCI_ETL_STORAGE_BUCKET": "fxci-etl",
+    "FXCI_ETL_PULSE_USER": "fxci-etl",
+}
+
+secrets = [
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_STORAGE_CREDENTIALS",
+        secret="airflow-gke-secrets",
+        key="fxci-etl-gcp-credentials",
```
akkomar, line 45:
```suggestion
        key="fxci_etl_secret__gcp-credentials",
```

dags/fxci_pulse_export.py
```diff
@@ -0,0 +1,79 @@
...
+    ),
+    Secret(
+        deploy_type="env",
+        deploy_target="FXCI_ETL_PULSE_PASSWORD",
+        secret="airflow-gke-secrets",
+        key="fxci-etl-pulse-password",
```
akkomar, line 51:
```suggestion
        key="fxci_etl_secret__pulse-password",
```

akkomar (approved):
r+, we decided to create tables as part of the job logic.

PR #2050, opened 2024-07-23
refactor(GLAM): Remove percentile calculation

edugfilho:
## Description

This PR removes steps that calculate percentiles from the GLAM dags.
Percentiles will be calculated ad hoc from now on for cost savings
See also: https://github.com/mozilla/bigquery-etl/pull/5957

### DIFF
--- a/dags/glam.py
+++ b/dags/glam.py
@@ -138,30 +138,6 @@ clients_scalar_aggregates = bigquery_etl_query(
     dag=dag,
 )
 
-scalar_percentiles = GKEPodOperator(
-    reattach_on_restart=True,
-    task_id="scalar_percentiles",
-    arguments=[
-        "python3",
-        "script/glam/run_scalar_agg_clustered_query.py",
-        "--submission-date",
-        "{{ds}}",
-        "--dst-table",
-        "scalar_percentiles_v1",
-        "--project",
-        project_id,
-        "--billing-project",
-        billing_project_id,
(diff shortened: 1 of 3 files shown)


kwindau (approved):

dags/glam.py
```diff
@@ -401,17 +362,12 @@
...
 
 clients_scalar_aggregates >> glam_user_counts
 glam_user_counts >> extract_counts
 
-
 extract_counts >> extracts_per_channel
 client_scalar_probe_counts >> extracts_per_channel
-scalar_percentiles >> extracts_per_channel
-histogram_percentiles >> extracts_per_channel
+clients_histogram_probe_counts >> extracts_per_channel
```
BenWu, line 372: Is this needed as well?
```suggestion
clients_scalar_aggregates >> extracts_per_channel
clients_histogram_probe_counts >> extracts_per_channel
```

BenWu (commented):
Looks like CI failure is an existing issue.  If they're known and not going to be fixed, does adding something like `# ignore: RUF017` to those lines work

dags/glam.py
edugfilho: Thanks, but not really. The dep chain here is:
`clients_scalar_aggregates >> client_scalar_probe_counts`  (L352)
and
`client_scalar_probe_counts >> extracts_per_channel` (L372)

PR #2057, opened 2024-08-01
fix(dbt_daily): Set dbt account ID

scholtzan:
The dbt account ID needs to be specified as part of the DAG config. This ID is not secret so it's fine to have it in an environment variable. It's more convenient to store it in a variable for other DAGs triggering dbt to reference it.

### DIFF
--- a/dags/dbt_daily.py
+++ b/dags/dbt_daily.py
@@ -29,7 +29,7 @@ default_args = {
     "retries": 2,
     "retry_delay": timedelta(minutes=30),
     "dbt_cloud_conn_id": "dbt_cloud",
-    "account_id": Variable.get("dbt_account_id")
+    "account_id": "{{ var.value.dbt_account_id }}"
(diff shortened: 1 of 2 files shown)


kwindau (approved):

dags/dbt_daily.py
```diff
@@ -27,6 +28,8 @@
     "email_on_retry": True,
     "retries": 2,
     "retry_delay": timedelta(minutes=30),
+    "dbt_cloud_conn_id": "dbt_cloud",
+    "account_id": Variable.get("dbt_account_id")
```
BenWu, line 32: I think best practice is to retrieve variables at runtime so it doesn't do a database query on each dag parse https://www.astronomer.io/docs/learn/airflow-variables#using-environment-variables
```suggestion
    "account_id": "{{ var.value.dbt_account_id }}"
```

PR #2062, opened 2024-08-14
fix(fxci): change fxci metric export schedule to run at 00:30

ahal:
There's a check in the code that prevents the export from running too close to midnight. This is because there's a slight lag in the Google Cloud Monitoring data becoming available.

Running 30 minutes past the hour should be plenty of buffer time.


ahal:
Note this isn't currently causing any issues because the DAG is set to retry in 30 minutes, so it passes on the retry. But it will prevent us from re-trying each time.

akkomar (approved):

PR #2068, opened 2024-08-21
feat: Add Looker content validation for KPI related dashboards and explores

scholtzan:
## Description

This adds a new task to the `looker` DAG which runs content validation on KPI dashboards and related explores via `spectacles`

### DIFF
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -170,13 +170,13 @@ with DAG(
         cmds=["bash", "-x", "-c"],
         arguments=[
             "pip install spectacles==2.4.10 &&" # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
-            r"spectacles content --verbose \ "
-            r" --base-url ${BASE_URL} \ "
-            r" --client_id ${LOOKER_API_CLIENT_ID} \ "
-            r" --client_secret ${LOOKER_API_CLIENT_SECRET} \ "
-            r" --project spoke-default \ "
-            r" --branch ${BRANCH} \ "
-            r" --pin-imports looker-hub:main \ "
+            "spectacles content --verbose"
+            " --base-url ${BASE_URL}"
+            " --client_id ${LOOKER_API_CLIENT_ID}"
+            " --client_secret ${LOOKER_API_CLIENT_SECRET}"
+            " --project spoke-default"
+            " --branch ${BRANCH}"
+            " --pin-imports looker-hub:main"
             "  --folders {' --folders '.join(looker_folders_to_validate)}"
         ],
         env_vars={
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -177,7 +177,7 @@ with DAG(
             " --project spoke-default"
             " --branch ${BRANCH}"
             " --pin-imports looker-hub:main"
-            f" --folders {' --folders '.join(looker_folders_to_validate)}"
+            f" --folders {' '.join(looker_folders_to_validate)}"
         ],
         env_vars={
             "BRANCH": "main-validation", # this branch is a mirror of main, but Looker cannot open production branches (like main) for validation
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -154,7 +154,7 @@ with DAG(
     )
 
     looker_folders_to_validate = [
-        706,  # KPI metrics
+        "706",  # KPI metrics
     ]
 
     lookml_validate_spectacles = GKEPodOperator(
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -169,7 +169,7 @@ with DAG(
         dag=dag,
         cmds=["bash", "-x", "-c"],
         arguments=[
-            "pip install spectacles==2.4.10 &&" # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
+            "pip install spectacles==2.4.10 && " # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
             "spectacles content --verbose"
             " --base-url ${BASE_URL}"
             " --client_id ${LOOKER_API_CLIENT_ID}"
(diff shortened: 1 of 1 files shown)


dags/looker.py
```diff
@@ -153,4 +153,33 @@
...
+            "ascholtz@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+        ],
+        task_id="lookml_validate_spectacles",
+        name="lookml-validate-spectacles",
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/lookml-generator:latest",
+        dag=dag,
+        cmds=["bash", "-x", "-c"],
+        arguments=[
+            "pip install spectacles==2.4.10 &&" # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
```
scholtzan, line 172: `lookml-generator` does have `spectacles` installed, but it needed to be pinned to version <=2.3.16 due to incompatible dependencies (pydantic) with mozilla-nimbus-schemas. These older versions however encounter some issue when connecting to the Looker API. Manually installing a more recent version here seems to work though for now.
I already pinged Mike to see if pydantic can be updated for mozilla-nimbus-schemas, but it looks like a more difficult thing to do.

dags/looker.py
```diff
@@ -153,4 +153,33 @@
         **airflow_gke_prod_kwargs,
     )
 
-    lookml_generator_staging >> lookml_generator_prod
+    looker_folders_to_validate = [
+        706,  # KPI metrics
```
scholtzan, line 157: This seems like the folder with the most important dashboards for now. Validation took about 3 minutes when I tested locally. We can add more folders as we test this out more

dags/looker.py
```diff
@@ -153,4 +153,38 @@
...
+            "ascholtz@mozilla.com",
+            "telemetry-alerts@mozilla.com",
+        ],
+        task_id="lookml_validate_spectacles",
+        name="lookml-validate-spectacles",
+        image="gcr.io/moz-fx-data-airflow-prod-88e0/lookml-generator:latest",
+        dag=dag,
+        cmds=["bash", "-x", "-c"],
+        arguments=[
+            "pip install spectacles==2.4.10 &&" # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
```
sean-rose, line 172:
```suggestion
            "pip install spectacles==2.4.10 && " # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
```

dags/looker.py
```diff
@@ -153,4 +153,38 @@
...
+            r"spectacles content --verbose \ "
+            r" --base-url ${BASE_URL} \ "
+            r" --client_id ${LOOKER_API_CLIENT_ID} \ "
+            r" --client_secret ${LOOKER_API_CLIENT_SECRET} \ "
+            r" --project spoke-default \ "
+            r" --branch ${BRANCH} \ "
+            r" --pin-imports looker-hub:main \ "
```
sean-rose, line 179: Pretty sure these don't need the backslash escapes since this doesn't end up as a mult-line command (and thus the strings also don't need to be raw strings).
```suggestion
            "spectacles content --verbose"
            " --base-url ${BASE_URL}"
            " --client_id ${LOOKER_API_CLIENT_ID}"
            " --client_secret ${LOOKER_API_CLIENT_SECRET}"
            " --project spoke-default"
            " --branch ${BRANCH}"
            " --pin-imports looker-hub:main"
```

dags/looker.py
```diff
@@ -153,4 +153,38 @@
...
+            "  --folders {' --folders '.join(looker_folders_to_validate)}"
```
sean-rose, line 180: This needs to be an f-string for the embedded expression to get evaluated.
```suggestion
            f" --folders {' --folders '.join(looker_folders_to_validate)}"
```

dags/looker.py
```diff
@@ -153,4 +153,38 @@
...
+        arguments=[
+            "pip install spectacles==2.4.10 && " # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
+            "spectacles content --verbose"
+            " --base-url ${BASE_URL}"
+            " --client_id ${LOOKER_API_CLIENT_ID}"
+            " --client_secret ${LOOKER_API_CLIENT_SECRET}"
+            " --project spoke-default"
+            " --branch ${BRANCH}"
+            " --pin-imports looker-hub:main"
+            f" --folders {' --folders '.join(looker_folders_to_validate)}"
```
sean-rose, line 180: The [Spectacles CLI docs](https://docs.spectacles.dev/cli/reference/content-validator/#folders) indicate `--folders` takes multiple values itself rather than being specified multiple times.
```suggestion
            f" --folders {' '.join(looker_folders_to_validate)}"
```

sean-rose: Also, it looks like calling `str.join()` on `looker_folders_to_validate` fails because the values are integers and it wants strings.  You could either make the list values literal strings upfront, or do something like `str(f) for f in looker_folders_to_validate`.

sean-rose (approved):
r+wc

dags/looker.py
```diff
@@ -153,4 +153,38 @@
         **airflow_gke_prod_kwargs,
     )
 
-    lookml_generator_staging >> lookml_generator_prod
+    looker_folders_to_validate = [
+        706,  # KPI metrics
```
scholtzan, line 157:
```suggestion
        "706",  # KPI metrics
```

scholtzan:
One slight annoyance I'm noticing is that the CLIENT_ID and CLIENT_SECRET appear in clear text in the Airflow logs. These credentials aren't super confidential, so this is fine for now. In the long term we might need to wrap the validation logic into some separate script that accesses the environment variables

sean-rose:
> One slight annoyance I'm noticing is that the CLIENT_ID and CLIENT_SECRET appear in clear text in the Airflow logs. These credentials aren't super confidential, so this is fine for now. In the long term we might need to wrap the validation logic into some separate script that accesses the environment variables

This has been resolved via #2071 / #2076.

PR #2071, opened 2024-08-22
fix: Using Airflow variables for LookML validation

scholtzan:
This PR switches from referencing secrets to referencing existing Airflow variables to pass the client secret for LookML validation. Previously the referenced secrets appeared in clear text in the logs. This change will ensure that they are masked

### DIFF
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -172,8 +172,8 @@ with DAG(
             "pip install spectacles==2.4.10 && " # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
             "spectacles content --verbose"
             " --base-url ${BASE_URL}"
-            " --client-id ${LOOKER_API_CLIENT_ID}"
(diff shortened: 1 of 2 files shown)


kwindau (approved):

dags/looker.py
```diff
@@ -172,8 +172,8 @@
             "pip install spectacles==2.4.10 && " # todo: remove this once mozilla-nimbus-schemas supports newer pydantic version
             "spectacles content --verbose"
             " --base-url ${BASE_URL}"
-            " --client-id ${LOOKER_API_CLIENT_ID}"
-            " --client-secret ${LOOKER_API_CLIENT_SECRET}"
```
sean-rose, line 176: I'd think you'd no longer want to pass these secrets below as `secrets=[looker_api_client_id_prod, looker_api_client_secret_prod]` below given this change, right?

PR #2073, opened 2024-08-26
feat: DENG-2963 Add optional sql generation to artifact deployment

BenWu:
## Description

Adds a `generate_sql` dag param that will run sql generation at the start of the table, view, and metadata publishing.  Plan is to use this when a run is triggered after schema deploy through https://github.com/mozilla-it/dataservices-infra/pull/345 and https://github.com/mozilla-services/cloudops-infra/pull/5736.

Verified this locally by looking at the rendered templates:

image

image

* DENG-2963

### DIFF
--- a/dags/bqetl_artifact_deployment.py
+++ b/dags/bqetl_artifact_deployment.py
@@ -73,8 +73,12 @@ generate_sql_cmd_template = (
 )
 
 
-def check_for_queued_runs(dag_id: str, generate_sql: bool) -> bool:
-    """Return true if there are no other queued dag runs or if the generate_sql param is true."""
+def should_run_deployment(dag_id: str, generate_sql: bool) -> bool:
+    """
+    Run deploys if there are no other queued dag runs or if the generate_sql param is true.
+
+    When used with ShortCircuitOperator, true means run downstream tasks and false means skip.
+    """
     queued_runs = DagRun.find(dag_id=dag_id, state=DagRunState.QUEUED)
     print(f"Found {len(queued_runs)} queued dag runs for {dag_id}")
     return len(queued_runs) == 0 or generate_sql == "True"
@@ -94,7 +98,7 @@ with DAG(
     skip_if_queued_runs_exist = ShortCircuitOperator(
         task_id="skip_if_queued_runs_exist",
         ignore_downstream_trigger_rules=True,
-        python_callable=check_for_queued_runs,
+        python_callable=should_run_deployment,
         op_kwargs={"dag_id": dag.dag_id, "generate_sql": "{{ params.generate_sql }}"},
     )
 
--- a/dags/bqetl_artifact_deployment.py
+++ b/dags/bqetl_artifact_deployment.py
@@ -1,5 +1,7 @@
 """
-Nightly deploy of bigquery etl views.
+Deploy of bigquery etl tables and views, typically triggered by bigquery-etl merges to main.
+
+SQL generation can optionally run during the tasks using the `generate_sql` DAG param.
 
 *Triage notes*
 
@@ -32,7 +34,7 @@ To find logs for specific datasets/views, add a string to search for to the end 
 from datetime import datetime, timedelta
 
 from airflow import DAG
-from airflow.models import DagRun
+from airflow.models import DagRun, Param
 from airflow.operators.python import ShortCircuitOperator
 from airflow.utils.state import DagRunState
 from airflow.utils.trigger_rule import TriggerRule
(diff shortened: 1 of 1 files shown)


dags/bqetl_artifact_deployment.py
```diff
@@ -56,11 +58,26 @@
...
+# renders generate sql command if params.generate_sql is true, else empty string
+generate_sql_cmd_template = (
+    "{{ 'script/bqetl generate all --use-cloud-function=false && ' "
+    "if params.generate_sql else '' }}"
+)
+
 
-def check_for_queued_runs(dag_id: str) -> bool:
+def check_for_queued_runs(dag_id: str, generate_sql: bool) -> bool:
+    """Return true if there are no other queued dag runs or if the generate_sql param is true."""
```
ANich, line 77: nit: since there's another condition here I'd maybe change this name and docstring:
```suggestion
def should_skip_run(dag_id: str, generate_sql: bool) -> bool:
    """Skip artifact deployment run if there are no other queued dag runs or if the generate_sql param is true."""
```

dags/bqetl_artifact_deployment.py
```diff
@@ -70,14 +87,15 @@ def check_for_queued_runs(dag_id: str) -> bool:
     schedule_interval="@daily",
     doc_md=__doc__,
     tags=tags,
+    params=params,
 ) as dag:
     docker_image = "gcr.io/moz-fx-data-airflow-prod-88e0/bigquery-etl:latest"
```
ANich, line 92: Isn't the generated SQL already included in this image?

dags/bqetl_artifact_deployment.py
```diff
@@ -119,7 +138,8 @@ def check_for_queued_runs(dag_id: str) -> bool:
         task_id="publish_views",
         cmds=["bash", "-x", "-c"],
         arguments=[
-            "script/bqetl view publish --add-managed-label --skip-authorized --project-id=moz-fx-data-shared-prod && "
+            generate_sql_cmd_template
```
sean-rose, line 145: Could we avoid running SQL generation three separate times?  Maybe we could have a conditional task that does the SQL generation and pushes it to the `private-generated-sql` branch if the param is set, then the publish tasks could conditionally pull down the latest files from that branch if the param is set?

dags/bqetl_artifact_deployment.py
BenWu: Yes but if we trigger this from jenkins after a schema deployment, the image won't be rebuilt so generation is run in the tasks.  This is to avoid issues like with the fenix views being outdated https://bugzilla.mozilla.org/show_bug.cgi?id=1904009 although we already have a workaround for this specific issue

dags/bqetl_artifact_deployment.py
```diff
@@ -56,11 +58,26 @@
...
+)
+
 
-def check_for_queued_runs(dag_id: str) -> bool:
+def check_for_queued_runs(dag_id: str, generate_sql: bool) -> bool:
+    """Return true if there are no other queued dag runs or if the generate_sql param is true."""
     queued_runs = DagRun.find(dag_id=dag_id, state=DagRunState.QUEUED)
     print(f"Found {len(queued_runs)} queued dag runs for {dag_id}")
-    return len(queued_runs) == 0
+    return len(queued_runs) == 0 or generate_sql == "True"
```
ANich, line 84: Is this a typo or is there some Airflow reason the `generate_sql` comparison is a `str` here but `bool` in the type hint? 
```suggestion
    return len(queued_runs) == 0 or generate_sql
```

dags/bqetl_artifact_deployment.py
```diff
@@ -56,11 +58,26 @@
...
+
+# renders generate sql command if params.generate_sql is true, else empty string
+generate_sql_cmd_template = (
+    "{{ 'script/bqetl generate all --use-cloud-function=false && ' "
+    "if params.generate_sql else '' }}"
+)
+
 
-def check_for_queued_runs(dag_id: str) -> bool:
+def check_for_queued_runs(dag_id: str, generate_sql: bool) -> bool:
```
sean-rose: The param will get passed as a string unless you specify `render_template_as_native_obj`.
```suggestion
def check_for_queued_runs(dag_id: str, generate_sql: str) -> bool:
```

ANich (approved):
LGTM, left some thoughts

On the linked ticket/bug: my first intuition probably would have been to trigger CircleCI instead of artifact deployment. It seems weird to have to re-generate sql even though the image _should_ have the most up to date SQL

sean-rose:
> On the linked ticket/bug: my first intuition probably would have been to trigger CircleCI instead of artifact deployment.

I like this a lot more than my idea of having  this push to the `private-generated-sql` branch.

Triggering a CircleCI build on `main` would regenerate the SQL, and that would then trigger a run of this DAG with the updated SQL already in the image.

BenWu:
> trigger CircleCI instead of artifact deployment

I considered that in this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1883727#c4 but went with this approach because we already have all the permissions setup.  @whd would it make sense to trigger circleci instead of the cloud function in https://github.com/mozilla-services/cloudops-infra/pull/5736?

whd:
> @whd would it make sense to trigger circleci instead of the cloud function in https://github.com/mozilla-services/cloudops-infra/pull/5736?

While it might, this is something somewhat notoriously finicky in CircleCI and part of the motivation for the ongoing OIDC work: https://mozilla-hub.atlassian.net/browse/DSRE-1261?focusedCommentId=811787. You can also read a bit about this on https://circleci.com/docs/triggers-overview/#run-a-pipeline-using-the-api as well.

Basically I don't expect it to be easy to trigger CircleCI externally and going down that path is something we'd only be able to explore after all the OIDC changes land. For now, I think this PR as designed is our best bet for getting something better than what we have now.

sean-rose:
> You can also read a bit about this on https://circleci.com/docs/triggers-overview/#run-a-pipeline-using-the-api as well.

That says "_for projects using GitHub App, the [custom webhooks](https://circleci.com/docs/triggers-overview/#trigger-a-pipeline-from-a-custom-webhook) functionality provides similar capabilities_" (though that custom webhooks feature is in preview at the moment, so I'm not sure if we'd be OK to use it given our stance on preview features).

whd (approved):

PR #2074, opened 2024-08-26
fix: update fxci_metric_export dag to pass in --date

ahal:
The underlying ETL is switching to explicitly passing in the date, so adjust accordingly.

* Bug 1912067


ahal:
This can't land until after https://github.com/mozilla/docker-etl/pull/274

akkomar (approved):

PR #2076, opened 2024-08-27
fix: Use Airflow templated variables for LookML validation

sean-rose:
## Description

So the variable values aren't shown in the task instance details.

* Fixup for #2071.


mikaeld (approved):
:rocket:

PR #2078, opened 2024-08-27
refactor: Change `lookml_validate_spectacles` task to use environment variables that Spectacles expects

sean-rose:
## Description

So the Looker base URL, client ID, and client secret don't need to be passed as command line arguments.

* Follow-up to #2076.

### DIFF
--- a/dags/looker.py
+++ b/dags/looker.py
@@ -67,6 +67,18 @@ looker_api_client_secret_staging = Secret(
     secret="airflow-gke-secrets",
     key="probe_scraper_secret__looker_api_client_secret_staging",
 )
+looker_client_id_prod = Secret(
+    deploy_type="env",
+    deploy_target="LOOKER_CLIENT_ID",
+    secret="airflow-gke-secrets",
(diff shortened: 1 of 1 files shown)


dags/looker.py
```diff
@@ -67,6 +67,18 @@
...
+    deploy_target="LOOKER_CLIENT_ID",
+    secret="airflow-gke-secrets",
+    key="probe_scraper_secret__looker_api_client_id_prod",
+)
+looker_client_secret_prod = Secret(
+    deploy_type="env",
+    deploy_target="LOOKER_CLIENT_SECRET",
+    secret="airflow-gke-secrets",
+    key="probe_scraper_secret__looker_api_client_secret_prod",
+)
```
sean-rose, line 81: In the future it could be nice to change `lookml-generator` to expect these same environment variables rather than `LOOKER_API_CLIENT_ID` and `LOOKER_API_CLIENT_SECRET`.

mikaeld (approved):

PR #2080, opened 2024-09-03
chore(GLAM): Turn FoG histogram_bucket_counts into TaskGroup

edugfilho:
## Description
This PR runs GLAM FoG ETL's `histogram_bucket_counts` in sample ranges, because `release_histogram_bucket_counts` frequently goes over the 6h BQ time limit and has to retry, which spends more money.

Complements https://github.com/mozilla/bigquery-etl/pull/6150 to fix https://github.com/mozilla/glam/issues/2936. 

* https://github.com/mozilla/glam/issues/2936

### DIFF
--- a/utils/glam_subdags/generate_query.py
+++ b/utils/glam_subdags/generate_query.py
@@ -140,6 +140,7 @@ def generate_and_run_glean_task(
         env_vars = {}
 
     query_name = task_name.split("_sampled_")[0]
+    # If a range smaller than 100% of the samples is being used then sample_id is needed.
     use_sample_id = min_sample_id > 0 or max_sample_id < 99
 
     env_vars = {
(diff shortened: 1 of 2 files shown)


utils/glam_subdags/generate_query.py
```diff
@@ -136,13 +139,17 @@ def generate_and_run_glean_task(
     if env_vars is None:
         env_vars = {}
 
+    query_name = task_name.split("_sampled_")[0]
+    use_sample_id = min_sample_id > 0 or max_sample_id < 99
```
scholtzan, line 143: what is this part supposed to do?

should it be something like this to check the sample ID range?

```suggestion
    use_sample_id = min_sample_id >= 0 and max_sample_id <= 99
```

edugfilho: This wants to find whenever a range smaller than 100% of the samples is being used. Otherwise we don't need to  `use_sample_id`. Does that make sense?

scholtzan: I see. Might be worth adding a comment

scholtzan (approved):

PR #2081, opened 2024-09-04
fix(GLAM): Fix glam gen query

edugfilho:
## Description

This PR fixes the currently broken GLAM FoG and Fenix ETLs by passing a string parameters to the k8s pod instead of a boolean.

Error in [logs](https://workflow.telemetry.mozilla.org/dags/glam_fenix/grid?dag_run_id=scheduled__2024-09-03T02%3A00%3A00%2B00%3A00&task_id=view_org_mozilla_fenix_glam_release__view_clients_daily_histogram_aggregates_v1&tab=logs):
```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod in version \"v1\" cannot be handled as a Pod: json: cannot unmarshal bool into Go struct field EnvVar.spec.containers.env.value of type string","reason":"BadRequest","code":400}```

### DIFF
--- a/utils/glam_subdags/generate_query.py
+++ b/utils/glam_subdags/generate_query.py
@@ -140,6 +140,7 @@ def generate_and_run_glean_task(
         env_vars = {}
 
     query_name = task_name.split("_sampled_")[0]
+    # If a range smaller than 100% of the samples is being used then sample_id is needed.
     use_sample_id = min_sample_id > 0 or max_sample_id < 99
 
     env_vars = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: utils/glam_subdags/generate_query.py [hunk 1]
```diff
@@ -150,7 +150,7 @@ def generate_and_run_glean_task(
         "DATASET": destination_dataset_id,
         "SUBMISSION_DATE": "{{ ds }}",
         "IMPORT": "true",
-        "USE_SAMPLE_ID": use_sample_id,
+        "USE_SAMPLE_ID": str(use_sample_id),
```
fbertsch, line 153: Does it matter that this will be `True` or `False`, while the above `IMPORT` env var is lowercase `true`?

edugfilho: Not really. [On the other side](https://github.com/mozilla/bigquery-etl/pull/6155/files#) the code is ready to receive `True` or `False`

edugfilho: While I ack that the difference in casing is not the best, though. Not sure if that's what you're referring to.

PR #2082, opened 2024-09-04
fix(GLAM): Fix TaskGroup upstream deps

edugfilho:
## Description

It looks like Airflow always uses the last task of a TaskGroup for dependency purposes. This PR fixes the dependency of a TaskGroup by making the firs task (instead of the last) depend on the upstream task.

### DIFF
--- a/dags/glam_fog.py
+++ b/dags/glam_fog.py
@@ -143,7 +143,7 @@ with DAG(
         with TaskGroup(
             group_id=f"{product}__histogram_bucket_counts_v1", dag=dag, default_args=default_args
         ) as histogram_bucket_counts:
-            prev_task = None
+            prev_task = clients_histogram_aggregate
             # Windows + Release data is in [0-9] so we're further splitting that range.
             for sample_range in ([0, 2], [3, 5], [6, 9], [10, 49], [50, 99]):
                 histogram_bucket_counts = query(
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

dags/glam_fog.py
```diff
@@ -143,7 +143,7 @@
         with TaskGroup(
             group_id=f"{product}__histogram_bucket_counts_v1", dag=dag, default_args=default_args
         ) as histogram_bucket_counts:
-            prev_task = None
+            prev_task = clients_histogram_aggregate
             # Windows + Release data is in [0-9] so we're further splitting that range.
             for sample_range in ([0, 2], [3, 5], [6, 9], [10, 49], [50, 99]):
                 histogram_bucket_counts = query(
```
sean-rose, line 149: I think the underlying problem here is that the `histogram_bucket_counts` variable for the task group is getting overridden here to be a specific task, which means that later `clients_histogram_aggregate >> histogram_bucket_counts` ends up adding a dependency on a specific task rather than the task group as a whole.  I'd recommend using different variable names for each case (e.g. `histogram_bucket_counts_sampled` here) and then set the dependency on the task group as was attempted before.

edugfilho: 🤦  you're right.
r? https://github.com/mozilla/telemetry-airflow/pull/2083

PR #2084, opened 2024-09-04
fix(socorro_import): try manually built image

whd:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1915996

### DIFF
--- a/dags/socorro_import.py
+++ b/dags/socorro_import.py
@@ -93,7 +93,10 @@ with DAG(
     bq_dataset = "telemetry_derived"
     bq_table_name = f"{dataset}_{dataset_version}"
 
-    docker_image = "docker.io/mozilla/parquet2bigquery:20190722"
+    # This image was manually built from
+    # https://github.com/mozilla/parquet2bigquery/commit/6bf1f86076de8939ba2c4d008080d6c159a0a093
+    # using python:3.7.4-slim-buster
+    docker_image = "gcr.io/moz-fx-data-airflow-prod-88e0/parquet2bigquery:20190722"
 
     gke_args = [
         "--dataset",


dags/socorro_import.py
```diff
@@ -93,7 +93,10 @@
     bq_dataset = "telemetry_derived"
     bq_table_name = f"{dataset}_{dataset_version}"
 
-    docker_image = "docker.io/mozilla/parquet2bigquery:20190722"
+    # This image was manually built from
+    # https://github.com/mozilla/parquet2bigquery
+    # using python:3.7.4-slim-buster
+    docker_image = "gcr.io/moz-fx-data-airflow-prod-88e0/parquet2bigquery:20190722"
```
sean-rose, line 99: Did you build this image from [parquet2bigquery commit 6bf1f86](https://github.com/mozilla/parquet2bigquery/commit/6bf1f86076de8939ba2c4d008080d6c159a0a093) on 2019-07-22, excluding subsequent changes?

Because if you built it from the latest parquet2bigquery code I'd recommend tagging it instead as `20191017` (date of the latest change).

dags/socorro_import.py
```diff
@@ -93,7 +93,10 @@
     bq_dataset = "telemetry_derived"
     bq_table_name = f"{dataset}_{dataset_version}"
 
-    docker_image = "docker.io/mozilla/parquet2bigquery:20190722"
+    # This image was manually built from
+    # https://github.com/mozilla/parquet2bigquery
+    # using python:3.8.19-slim-bullseye
```
whd, line 98: Bumped per https://mozilla.slack.com/archives/C019WG3TTM2/p1725490574279479?thread_ts=1725488070.789599&cid=C019WG3TTM2, seems to have built correctly

sean-rose (approved):
r+wc

dags/socorro_import.py
whd: I've reverted usage of python 3.8 and updated the comment to make this maximally similar to the version that we were using in prod before.

PR #2099, opened 2024-09-19
chore: update ownership for anicholson DAGs

ANich:
## Description

This PR updates ownership for anicholson DAGs

...

### DIFF
--- a/dags/bqetl_backfill_complete.py
+++ b/dags/bqetl_backfill_complete.py
@@ -17,6 +17,7 @@ tags = [Tag.ImpactTier.tier_3]
 
 default_args = {
     "email": [
+        "ascholtz@mozilla.com",
         "wichan@mozilla.com",
     ]
 }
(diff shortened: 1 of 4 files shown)


dags/bqetl_backfill_complete.py
```diff
@@ -17,7 +17,6 @@
 
 default_args = {
     "email": [
-        "anicholson@mozilla.com",
```
scholtzan, line 20:
```suggestion
        "ascholtz@mozilla.com",
```

dags/bqetl_backfill_initiate.py
```diff
@@ -17,7 +17,6 @@
 
 default_args = {
     "email": [
-        "anicholson@mozilla.com",
```
scholtzan, line 20:
```suggestion
        "ascholtz@mozilla.com",
```

dags/bqetl_backfill_complete.py
ANich: GH won't let me commit suggestions on a delete interestingly

scholtzan: huh, interesting

sean-rose (approved):

PR #2107, opened 2024-10-02
feat: Install Bigeye SDK

scholtzan:
## Description

Install Bigeye SDK in Airflow to trigger Bigeye metric runs. This unblocks https://github.com/mozilla/bigquery-etl/pull/6285


requirements.txt
```diff
@@ -43,8 +43,11 @@ babel==2.14.0
 backoff==2.2.1
 bcrypt==4.1.2
 beautifulsoup4==4.12.3
+betterproto==1.2.5
 bigeye-airflow==0.1.12
```
scholtzan, line 47: oddly enough, bigeye-sdk doesn't come with bigeye-airflow

BenWu: This version of bigeye-airflow is a couple years old so things might have changed https://pypi.org/project/bigeye-airflow/#history

Is it using this version because of compatibility reasons?

scholtzan: Right, the new `bigeye-airflow` version is incompatible with the required `keyring` version: https://github.com/mozilla/telemetry-airflow/blob/323e75dfe6a57a584d77b89a5b1ae7abf1942efa/constraints.txt#L416

We actually do override the install in production though to install the most recent bigeye-sdk and bigeye-airflow: https://github.com/mozilla/telemetry-airflow/blob/main/requirements-override.txt

So this is really just needed to unblock bigquery-etl which does not install the overrides (though maybe it should?)

BenWu (approved):
Worth keeping in mind that it's using such an old version in case the behaviour might not match prod but we if it works it works

scholtzan:
Ok, this isn't necessary after all now that we install the override requirements in bigquery-etl CI

PR #2108, opened 2024-10-07
fix(providers): Upgrade to kubernetes provider 8.3.4.

mikaeld:
## Description
Patch k8s provider to version 8.3.4 to fix the issue of executor open slots reaching negative number
image

## Validation
image
_new provider installed_

image
_running clean_gke_pods_

* https://github.com/apache/airflow/pull/39551
* https://github.com/apache/airflow/issues/36998
* https://github.com/apache/airflow/issues/40106

### DIFF
--- a/requirements-override.txt
+++ b/requirements-override.txt
@@ -1,2 +1,3 @@
 bigeye-airflow==0.1.34
 bigeye-sdk==0.4.81
+apache-airflow-providers-cncf-kubernetes==8.3.4


akkomar (approved):

sean-rose (approved):

requirements-override.txt
```diff
@@ -1,2 +1,3 @@
 bigeye-airflow==0.1.34
 bigeye-sdk==0.4.81
+apache-airflow-providers-cncf-kubernetes==8.3.4
```
sean-rose, line 3: Tested this locally and it doesn't look like it will affect any other packages/dependencies. 👍

sean-rose: Also, looking forward to when we upgrade Airflow itself, this override will still be necessary for Airflow 2.9.2 since it defaults to `apache-airflow-providers-cncf-kubernetes` version 8.3.1, but it can be removed once we upgrade to Airflow 2.10.2 since that defaults to `apache-airflow-providers-cncf-kubernetes` version 8.4.1.

PR #2112, opened 2024-10-17
fix(glam): add separate import revisions step

edugfilho:
## Description

Fixes https://github.com/mozilla/glam/issues/2989 by adding a separate step to import build revisions into GLAM's database.

* https://github.com/mozilla/glam/issues/2989

### DIFF
--- a/dags/glam_glean_imports.py
+++ b/dags/glam_glean_imports.py
@@ -131,16 +131,16 @@ for env in ["Dev", "Prod"]:
             [wait_for_fenix, wait_for_fog] >> glean_env_task_group
 
 
-default_glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2023.07.1-43"
+default_glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2024.10.0-58"
 
 base_docker_args = ["/venv/bin/python", "manage.py"]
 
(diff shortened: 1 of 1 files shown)


dags/glam_glean_imports.py
```diff
@@ -131,16 +131,16 @@
...
 base_docker_args = ["/venv/bin/python", "manage.py"]
 
 for env in ["Dev", "Prod"]:
     glam_import_image = default_glam_import_image
     if env == "Dev":  # noqa 114
-        glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2023.07.1-43"
+        glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2024.10.0-58"
     elif env == "Prod":
-        glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2023.07.1-43"
+        glam_import_image = "gcr.io/moz-fx-dataops-images-global/gcp-pipelines/glam/glam-production/glam:2024.10.0-58"
```
BenWu, line 143: Why is this needed when all the images are the same?

BenWu (approved):

dags/glam_glean_imports.py
edugfilho: GLAM has different psql dbs for dev and prod, so this would allow for testing of more complicated things. That said, I don't think I've ever seen it being used.

PR #2113, opened 2024-10-22
feat(bigeye): Add operator to trigger Bigeye runs from within bqetl

scholtzan:
## Description

This adds a new function to trigger Bigeye check runs via the `bqetl monitoring run` command. Using this over the provided Bigeye Airflow operators allows us to handle failures and warning differently, and to trigger custom SQL rules.

Depends on https://github.com/mozilla/bigquery-etl/pull/6383

https://mozilla-hub.atlassian.net/browse/DENG-5588

### DIFF
--- a/utils/gcp.py
+++ b/utils/gcp.py
@@ -406,6 +406,51 @@ def bigquery_dq_check(
         **kwargs,
     )
 
+def bigquery_bigeye_check(
+    task_id,
+    table_id,
+    warehouse_id,
+    project_id="moz-fx-data-shared-prod",
+    gcp_conn_id="google_cloud_airflow_gke",
+    gke_project_id=GCP_PROJECT_ID,
+    gke_location="us-west1",
+    gke_cluster_name="workloads-prod-v1",
+    gke_namespace="default",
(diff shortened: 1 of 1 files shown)


utils/gcp.py
```diff
@@ -406,6 +406,51 @@ def bigquery_dq_check(
...
+    task_id,
+    table_id,
+    warehouse_id,
+    project_id="moz-fx-data-shared-prod",
+    gcp_conn_id="google_cloud_airflow_gke",
+    gke_project_id=GCP_PROJECT_ID,
+    gke_location="us-west1",
+    gke_cluster_name="workloads-prod-v1",
+    gke_namespace="default",
+    docker_image="gcr.io/moz-fx-data-airflow-prod-88e0/bigquery-etl:latest",
```
kik-kik, line 419: I assume this image is now referenced in multiple places and should always be the same. Maybe it would be worth making this a constant value on top of the module and referencing that everywhere instead to avoid accidental use of different bqetl images?

kik-kik (approved):

PR #2121, opened 2024-11-05
fix: Change webcompat_kb DAG to run a module instead of a script

ksy36:
## Description
Since webcompat_kb was converted to a module in https://github.com/mozilla/docker-etl/pull/298#event-15123335703, the job is failing as it still trying to run a script, this PR fixes it.

### DIFF
--- a/dags/webcompat_kb.py
+++ b/dags/webcompat_kb.py
@@ -62,7 +62,7 @@ with DAG(
         task_id="webcompat_kb",
         arguments=[
             "python",
-            "-m"
+            "-m",
             "webcompat_kb.main",
             "--bq_project_id",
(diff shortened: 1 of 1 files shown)


dags/webcompat_kb.py
```diff
@@ -62,7 +62,8 @@
         task_id="webcompat_kb",
         arguments=[
             "python",
-            "webcompat_kb/main.py",
+            "-m"
```
scholtzan, line 65:
```suggestion
            "-m",
```

scholtzan (approved):

ksy36:
Oh I missed the comma, thanks for fixing it!

PR #2122, opened 2024-11-05
fix: Change webcompat_kb DAG arguments to use dashes instead of underscores

ksy36:
## Description
This PR changes arguments to match the new format that webcompat ETL job expects (dashes instead of underscores)


ksy36:
r? @scholtzan 

Sorry for too many PRs. We kept support for legacy arguments in the ETL job, but realized it's actually writing to the test table instead of production because of an additional check in https://github.com/mozilla/docker-etl/blob/main/jobs/webcompat-kb/webcompat_kb/main.py#L59

scholtzan (approved):
No problem :)

ksy36:
thanks!

PR #2135, opened 2024-12-10
feat(glam): Extract glam_fog_release

edugfilho:
## Description

The release aggregates table is being recalculated nightly, but release channel builds happen infrequently enough that we can reduce this cadence to weekly without meaningful loss in usability.

This PR extracts the second half of  `glam_fog` `release` to `glam_fog_release` and schedules the latter every Sat at 10am

* DENG-6752

### DIFF
--- a/dags/glam_fog.py
+++ b/dags/glam_fog.py
@@ -76,6 +76,10 @@ with DAG(
         task_id="pre_import",
     )
 
+    daily_release_done = EmptyOperator(
+        task_id="daily_release_done",
+    )
+
     with TaskGroup("glam_fog_external") as glam_fog_external:
         ExternalTaskMarker(
             task_id="glam_glean_imports__wait_for_fog",
@@ -102,6 +106,7 @@ with DAG(
         reduce(operator.iadd, LOGICAL_MAPPING.values(), [])
     )
     for product in final_products:
+        is_release = "release" in product
         func = partial(
             generate_and_run_glean_task,
             product=product,
@@ -168,20 +173,6 @@ with DAG(
 
         sample_counts = view(task_name=f"{product}__view_sample_counts_v1")
 
-        export = GKEPodOperator(
-            task_id=f"export_{product}",
-            image="gcr.io/moz-fx-data-airflow-prod-88e0/bigquery-etl:latest",
-            arguments=["script/glam/export_csv"],
-            env_vars={
-                "SRC_PROJECT": PROJECT,
-                "DATASET": "glam_etl",
-                "PRODUCT": product,
-                "BUCKET": BUCKET,
-            },
-            is_delete_operator_pod=False,
-            cmds=["bash"],
-        )
-
         # set all of the dependencies for all of the tasks
 
         # get the dependencies for the logical mapping, or just pass through the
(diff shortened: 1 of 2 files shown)


dags/glam_fog_release.py
```diff
@@ -0,0 +1,124 @@
...
+    default_args=default_args,
+    max_active_runs=1,
+    schedule_interval="0 10 * * 6",  # 10am on Saturday
+    tags=tags,
+) as dag:
+    wait_for_glam_fog = ExternalTaskSensor(
+        task_id="wait_for_daily_glam_fog_release",
+        external_dag_id="glam_fog",
+        external_task_id="daily_release_done",
+        execution_delta=timedelta(hours=1),
```
BenWu, line 48: `glam_fog` runs at 2:00 so time delta should be 8 hours but also since this is weekly I think this needs an additional 1 week offset since the execution_date would be at the start of the interval, otherwise  it just waits for last week's task.  This is done in `firefox_public_data_report`:
https://github.com/mozilla/telemetry-airflow/blob/b8c132b9be38c636df23e66a11cae935fbfecaf5/dags/firefox_public_data_report.py#L45-L51

I think this is what it should be
```suggestion
        execution_delta=timedelta(days=-5, hours=-16),
```
or `timedelta(days=-6, hours=8)`.  I'm not sure though so it might be worth testing locally.

dags/glam_fog.py
```diff
@@ -168,20 +173,6 @@
 
         sample_counts = view(task_name=f"{product}__view_sample_counts_v1")
 
-        export = GKEPodOperator(
```
BenWu, line 171: Was this meant to be moved to the other dag?  I don't see a replacement

dags/glam_fog.py
```diff
@@ -206,24 +197,27 @@
...
-        probe_counts >> sample_counts >> extract_probe_counts >> export >> pre_import
-        (
-            clients_scalar_aggregate
-            >> user_counts
-            >> extract_user_counts
-            >> export
-            >> pre_import
-        )
-        clients_histogram_aggregate >> export >> pre_import
+        if is_release:
```
BenWu, line 200: The `counts` tasks above still get defined and run if release is true.  Should those be moved into the else block?

edugfilho: That is the intended behaviour. `daily` and `clients` tasks need to run daily for release too, otherwise the single weekly run would have to process the whole week in a day.

edugfilho: I'll add a comment there clarifying

dags/glam_fog_release.py
BenWu: Since this is weekly, the logical date of the dag would be a week before when it actually runs since it would be the start of the data interval.  So if it runs on 2024-12-14 with `timedelta=8 hours`, the logical date would be `2024-12-07 10:00` and it would wait for the task in `glam_fog` that has the date of `2024-12-07 02:00` which is the one that ran on `2024-12-08`.  -6 days would make it wait for the task that ran on 2024-12-14 since `glam_fog` is daily which means the logical date would be 1 day before when it runs.

BenWu: The [firefox_public_data_report](https://workflow.telemetry.mozilla.org/dags/firefox_public_data_report/grid?search=firefox_public_data_report&dag_run_id=scheduled__2024-12-02T01%3A00%3A00%2B00%3A00&tab=details&task_id=wait_for_main_ping) dag is an example of this.  It's weekly but waits for copy_deduplicate

edugfilho: I see it now! The `logical_date` for weekly runs is at the start of the data interval. Thanks for taking the time to explain it twice 😬

Your suggestion below is correct, then
`execution_delta=timedelta(days=-5, hours=-16),`

PR #2141, opened 2024-12-16
chore(glam): Remove useless export/import user_counts tasks from glam_fog and glam_fenix

edugfilho:
## Description

This PR removes tasks that import `user_counts`. These imports have been irrelevant since GLAM stopped using separate counts [here](https://github.com/mozilla/glam/pull/1876/files#r1887675292)

PR #2149, opened 2025-01-16
chore(glam): remove unused constants

edugfilho:
## Description

This PR removes the `BUCKET` constant from `glam` dags. The constant is no longer used after https://github.com/mozilla/telemetry-airflow/pull/2141


corbanCloud (approved):
This is the Glam V1 project value so good to remove if not used as it could cause confusion

PR #2154, opened 2025-01-22
chore(glam): Extract glam_fenix_release to a different DAG

edugfilho:
fixes https://github.com/mozilla/bigquery-etl/issues/6849
## Description

Following https://github.com/mozilla/telemetry-airflow/pull/2135, this PR extracts the second half of glam_fenix release to glam_fenix_release and schedules the latter every Sat at 10am.

The reason why I didn't make `glam_fog_release` and `glam_fenix_release` one single DAG is there may be a need to schedule them at different times, since `glam_fog_release` is already suffering from lack of resources. I might join them in the future

* DENG-7623

### DIFF
--- a/dags/glam_fenix_release.py
+++ b/dags/glam_fenix_release.py
@@ -1,0 +1,119 @@
+from datetime import datetime, timedelta
+from functools import partial, reduce
+
+from airflow import DAG
+from airflow.operators.empty import EmptyOperator
+from airflow.sensors.external_task import ExternalTaskSensor
+from airflow.utils.task_group import TaskGroup
+
+from utils.constants import ALLOWED_STATES, FAILED_STATES
+from utils.glam_subdags.generate_query import (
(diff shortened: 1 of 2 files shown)


dags/glam_fenix_release.py
```diff
@@ -0,0 +1,119 @@
...
+            product=product,
+            destination_project_id=PROJECT,
+            env_vars={"STAGE": "incremental"},
+        )
+        view, init, query = (
+            partial(func, task_type=task_type)
+            for task_type in ["view", "init", "query"]
+        )
+
+        # stage 2 - downstream for export
```
BenWu, line 67: I think these task definitions could be put in a function and shared between the dags, which would make it easier to maintain and ensure they're identical. It's fine as-is though

edugfilho: It's something I definitely intend to do

PR #2168, opened 2025-02-18
fix(glam_fog_release): redistribute bucket_counts sample processing

edugfilho:
## Description

fixes [Bug 1944388](https://bugzilla.mozilla.org/show_bug.cgi?id=1944388)
This PR splits that 0 to 9 sample range for `histogram_bucket_counts` in an attempt to avoid running out of resources. This range has more data than the rest due to Windows + Release data being sampled at 10%.

* [Bug 1944388](https://bugzilla.mozilla.org/show_bug.cgi?id=1944388)

### DIFF
--- a/dags/glam_fog_release.py
+++ b/dags/glam_fog_release.py
@@ -78,7 +78,7 @@ with DAG(
         ) as histogram_bucket_counts:
             prev_task = None
             # Windows + Release data is in [0-9] so we're further splitting that range.
-            for sample_range in ([0, 0], [1, 2], [3, 5], [6, 9], [10, 49], [50, 99]):
+            for sample_range in ([0, 0], [1, 1], [2, 3], [4, 6], [7, 9], [10, 49], [50, 99]):
                 histogram_bucket_counts_sampled = query(
                     task_name=f"{product}__histogram_bucket_counts_v1_sampled_{sample_range[0]}_{sample_range[1]}",
                     min_sample_id=sample_range[0],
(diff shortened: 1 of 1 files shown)


dags/glam_fog_release.py
```diff
@@ -78,7 +78,7 @@
         ) as histogram_bucket_counts:
             prev_task = None
             # Windows + Release data is in [0-9] so we're further splitting that range.
-            for sample_range in ([0, 0], [1, 2], [3, 5], [6, 9], [10, 49], [50, 99]):
+            for sample_range in ([0, 0], [1, 1], [2, 3], [4, 6], [7, 9], [10, 49], [50, 99]):
```
BenWu, line 81: Do sample ids 0 and 1 have more samples than 2-9?  Just wondering why they're 1% and the others are 2-3%

BenWu (approved):
It might be insightful to dry run the queries in the bq console to see if the bytes estimate accounts for sampling.  That would indicate that the sample sizes are small enough to take advantage of the metadata optimizations discussed in DENG-6782

dags/glam_fog_release.py
edugfilho: The size is the same for each sample from 0-9 inclusive (1.17TB) but I don't know why the first jobs take longer...

edugfilho:
Thanks for the suggestion. I checked and noticed the estimate accounts for sampling and the size is the same for each sample from 0-9 inclusive (1.17TB each sample). I'm thinking of further splitting this to process 1 sample at a time in the critical range then 10 samples at a time after that (each sample in the range of 10-99 inclusive carry about 106GB of data)

PR #2169, opened 2025-02-20
chore(taar): DENG-7245 Remove TAAR jobs

akkomar:
This removes two TAAR DAGs that have been paused for some time and are now ready for decommissioning.
See https://mozilla-hub.atlassian.net/browse/DENG-7245.


scholtzan (approved):
🎉

PR #2171, opened 2025-02-27
fix(fxci): use 'fxci_derived' dataset

ahal:
I have a couple questions I'm hoping someone from the data org can answer before this lands:

1. My understanding is that these should go in the `fxci_derived` dataset instead of `fxci`. Then in bigquery-etl, there should be `view.sql` files that create the user facing views. Just want to verify, is this understanding correct?

2. Can we move the existing tables from the `fxci` dataset to `fxci_derived` right before this is deployed? What's the best way to do that? If this is tricky, I'd honestly be ok with losing that data. This ETL is just on the cusp of exiting the prototype phase, so there's nothing relying on this data quite yet.

### DIFF
--- a/dags/fxci_metric_export.py
+++ b/dags/fxci_metric_export.py
@@ -26,7 +26,7 @@ tags = [Tag.ImpactTier.tier_3]
 
 env_vars = {
     "FXCI_ETL_BIGQUERY_PROJECT": "moz-fx-data-shared-prod",
-    "FXCI_ETL_BIGQUERY_DATASET": "fxci",
+    "FXCI_ETL_BIGQUERY_DATASET": "fxci_derived",
     "FXCI_ETL_STORAGE_PROJECT": "moz-fx-dev-releng",
     "FXCI_ETL_STORAGE_BUCKET": "fxci-etl",
 }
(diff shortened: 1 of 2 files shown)


ahal:
Actually, I guess the tables would need to be duplicated because there are views in `lookml-generator` that reference them and will need to be updated to point to the new tables first.

ahal:
So on second thought, I think I want to just land this first and worry about potentially backfilling the data from the old table later. Though starting over is also an attractive option :)

scholtzan:
> 1. My understanding is that these should go in the fxci_derived dataset instead of fxci. Then in bigquery-etl, there should be view.sql files that create the user facing views. Just want to verify, is this understanding correct?

Yes, your understanding is correct. The `_derived` datasets contain versioned tables which ETL processes write to directly. The datasets without any suffix are considered "user-facing" and generally have views that point to the most recent version of the corresponding table in the `_derived` dataset.

> 2. Can we move the existing tables from the fxci dataset to fxci_derived right before this is deployed? What's the best way to do that? If this is tricky, I'd honestly be ok with losing that data. This ETL is just on the cusp of exiting the prototype phase, so there's nothing relying on this data quite yet.

I moved the existing tables into `fxci_derived`

scholtzan (approved):
I opened https://github.com/mozilla/lookml-generator/pull/1167 to update the table references for Looker

PR #2174, opened 2025-03-13
chore(glean_dictionary): building  glean dictionary whether upstream fails or succeeds

edugfilho:
## Description

Discussed in the Glean Platform Working Group meeting.

Sometimes the `trigger_looker` step that is the immediate upstream of `glean_dictionary_netlify_build` fails and we don't want it to block the Glean Dictionary daily build, because that prevents users from seeing newly added metrics.

* DENG-8049

### DIFF
--- a/dags/probe_scraper.py
+++ b/dags/probe_scraper.py
@@ -396,6 +396,7 @@ with DAG(
             "telemetry-alerts@mozilla.com",
         ],
         task_id="glean_dictionary_build",
+        # Glean Dictionary utilizes data from generated LookML namespaces. If Looker DAG fails we want to run the Dictionary build anyway to load updated generated schemas
         trigger_rule="all_done",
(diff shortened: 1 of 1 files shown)


dags/probe_scraper.py
```diff
@@ -396,6 +396,7 @@ def choose_branch(self, context):
             "telemetry-alerts@mozilla.com",
         ],
         task_id="glean_dictionary_build",
+        trigger_rule="all_done",
```
akkomar, line 399:
```suggestion
        # Glean Dictionary utilizes data from generated LookML namespaces. If Looker DAG fails we want to run the Dictionary build anyway to load updated generated schemas
        trigger_rule="all_done",
```

akkomar (approved):
LGTM, I'd only add a comment explaining the trigger_rule line.

PR #2176, opened 2025-03-18
Add catalyst DAG to run on a daily schedule

dpalmeiro:
## Description

This PR adds https://mozilla-hub.atlassian.net/browse/DENG-7722 which runs the https://github.com/mozilla/catalyst performance analysis framework on recently completed Nimbus experiments.


dpalmeiro:
@mikewilli fyi

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
+"""
+DAG to schedule generation of performance reports for recently completed nimbus experiments.
+
+See the [jetstream repository](https://github.com/mozilla/catalyst).
```
mikewilli, line 4:
```suggestion
See the [catalyst repository](https://github.com/mozilla/catalyst).
```

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "depends_on_past": False,
+    "start_date": datetime(2025, 3, 18),
+    "email_on_failure": True,
+    "email_on_retry": True,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=30),
+}
+
+tags = [Tag.ImpactTier.tier_1]
```
mikewilli, line 28: I think I'd call this tier 2 or 3, but let me see if I can find some sort of documented guidance.

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
...
+
+with DAG(
+    "catalyst",
+    default_args=default_args,
+    schedule_interval="0 4 * * *",
+    doc_md=__doc__,
+    tags=tags,
+) as dag:
+    # Built from repo https://github.com/mozilla/catalyst
+    jetstream_image = "gcr.io/moz-fx-data-experiments/catalyst:latest"
```
mikewilli, line 38:
```suggestion
    catalyst_image = "gcr.io/moz-fx-data-experiments/catalyst:latest"
```

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
...
+
+    jetstream_run = GKEPodOperator(
```
mikewilli, line 40:
```suggestion
    catalyst_run = GKEPodOperator(
```

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
...
+        task_id="catalyst_run",
+        name="catalyst_run",
+        image=jetstream_image,
```
mikewilli, line 43:
```suggestion
        image=catalyst_image,
```

dags/catalyst.py
```diff
@@ -0,0 +1,46 @@
+"""
+DAG to schedule generation of performance reports for recently completed nimbus experiments.
+
+See the [jetstream repository](https://github.com/mozilla/catalyst).
+
```
mikewilli, line 5: It may be helpful to add a section to these docs for Airflow Triage called `*Triage Notes*`. Airflow Triage is a rotating responsibility for DEs to ensure that DAG failures are noticed and documented in a bug. The triage person may also attempt to diagnose and resolve issues when possible.

If you're not sure of any patterns or specific advice now, you can always update this later if any patterns emerge.

dags/catalyst.py
mikewilli: Ok, here are the docs: https://github.com/mozilla/bigquery-etl/blob/main/docs/reference/airflow_tags.md

Based on this, I'd suggest Tier 2, but open to your opinion obviously.

mikewilli, line 28: Related to the above comment on Airflow Triage, just wanted to make sure you're aware of the `no_triage` and `record_only` tags that will tell the triage person to skip the DAG entirely, or only record a bug without attempting to investigate. I don't think you need either of these, but FYI in case you think they're applicable.

mikewilli (changes requested):
Looks like a good start, left you some thoughts and suggestions.

The other thing I'd suggest that I didn't leave a comment about is whether you want this DAG to depend on anything upstream (i.e., wait on the necessary data to be available). You could either depend directly on the relevant DAGs, or on a proxy DAG that is already waiting on the things you need (i.e., you could wait on Jetstream or something similar).

dags/catalyst.py
dpalmeiro: Thanks, not sure how I missed all of these!

dags/catalyst.py
dpalmeiro: Tier 2 sounds good here.

dags/catalyst.py
```diff
@@ -0,0 +1,59 @@
...
+*Triage notes*
+
+This app will perform some bigquery queries, and generate statistical reports based on that data which are
+then published to https://protosaur.dev/perf-reports/index.html.
+
+Occasionally, the app can fail if it cannot understand the nimbus experiment configuration properly, but most
+of these cases should have coverage by now.  Other issues that may occur is if the sample size is too low which
+may cause some problems during the statistical analysis.  I'm still looking for more examples of these situations
+to fix.
```
mikewilli, line 14: Since you're saying that app execution failures will be fixed by the performance team, you probably don't need to include this part in the triage notes (unless I'm misinterpreting).

dags/catalyst.py
```diff
@@ -0,0 +1,59 @@
...
+    "catalyst",
+    default_args=default_args,
+    schedule_interval="0 4 * * *",
+    doc_md=__doc__,
+    tags=tags,
+) as dag:
+    # Built from repo https://github.com/mozilla/catalyst
+    catalyst_image = "gcr.io/moz-fx-data-experiments/catalyst:latest"
+
+    catalyst_run = GKEPodOperator(
```
mikewilli, line 53: What do you think about having this dependent on an upstream DAG or task? This would ensure that new data is available before this task runs. I'm guessing you could do the same upstream tasks as Jetstream since I expect that it needs the same (or similar) data.

mikewilli: See https://github.com/mozilla/telemetry-airflow/blob/main/dags/jetstream.py#L86-L145 for how the jetstream DAG configures this.

mikewilli (changes requested):
I think you want to have an upstream dependency to ensure data is available before execution, but otherwise looks good!

dags/catalyst.py
dpalmeiro: Hmm.  What would the value of doing this be?  The project currently checks the nimbus list for completed experiments and then starts working on any recently completed ones that fit the requirements.  Do you propose separating that step out and doing it independently as a separate DAG?

dpalmeiro: I see.  If there is a slightly delay for telemetry data to come in, then that is a valid concern.  For now, I can adjust the experiment search so that it only kicks off an analysis for experiments that have finished a day or two prior.  I do agree it might be useful to add a separate upstream DAG to do this, but I'm a bit eager to mostly get some kind of performance coverage at the moment since we don't have any right now.  What do you think?

mikewilli: I think it would be pretty safe to wait on the same telemetry that Jetstream waits on, which isn't necessarily an exact overlap but includes general events and client info. If you'd prefer to delay by a day that's alright too, but I think this should be a pretty easy thing to put in here now without having to create additional DAGs or anything. You can just copy those lines I linked above and swap out the jetstream variable name in the last part (`jetstream_run.set_upstream(` --> `catalyst_run.set_upstream(`), so I think this would be even quicker than updating the code logic.

dags/catalyst.py
```diff
@@ -0,0 +1,117 @@
...
+from utils.constants import ALLOWED_STATES, FAILED_STATES
+from utils.tags import Tag
+
+default_args = {
+    "owner": "dpalmeiro@mozilla.com",
+    "email": [
+        "dpalmeiro@mozilla.com",
+    ],
+    "depends_on_past": False,
+    "start_date": datetime(2025, 3, 18),
```
mikewilli, line 31: I believe Airflow will try to catch up by running all the intervening dates, so to be safe let's go with today.
```suggestion
    "start_date": datetime(2025, 4, 30),
```

mikewilli (approved):
Ok PR looks good now, thanks for sticking with me on the dependencies!

I'll approve now but please update the start_date before merging.

PR #2181, opened 2025-04-02
chore: Upgrade `acryl-datahub-airflow-plugin` to v0.15.0.5

sean-rose:
## Description
Part of the instigation for doing this is that after upgrading WTMO to Airflow 2.10 (DENG-5881) there's been an issue with some task tries not showing in the main UI, and I believe I've identified that the `acryl-datahub-airflow-plugin`'s custom retry handler code is somehow the culprit (disabling the DataHub plugin in the local Docker Compose setup makes the problem go away).  I'm hoping upgrading to a newer version of `acryl-datahub-airflow-plugin` might solve that issue.

* DENG-5881: Upgrade WTMO to Airflow 2.10

### DIFF
--- a/requirements-override.txt
+++ b/requirements-override.txt
@@ -1,3 +1,7 @@
+# acryl-datahub-airflow-plugin 0.15 requires openlineage-integration-common 1.25.0,
+# while the Airflow 2.10.5 constraints specify openlineage-integration-common 1.27.0.
+acryl-datahub-airflow-plugin==0.15.0.5
+
 # There's a bug in apache-airflow-providers-google 12.0.0 where Dataproc operators fail to import
 # without OpenLineage installed, which was fixed in 14.0.0 (https://github.com/apache/airflow/pull/46561).
 apache-airflow-providers-google==14.0.0
--- a/requirements.in
+++ b/requirements.in
@@ -16,7 +16,7 @@ apache-airflow-providers-slack
 airflow-provider-fivetran-async==2.0.2
 
 # Acryl DataHub integration
-acryl-datahub-airflow-plugin
+acryl-datahub-airflow-plugin==0.14.1.12
 gql
 
 # dbt integration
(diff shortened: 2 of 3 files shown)


requirements-override.txt
```diff
@@ -1,3 +1,7 @@
+# acryl-datahub-airflow-plugin 0.15 requires openlineage-integration-common 1.25.0,
+# while the Airflow 2.10.5 constraints specify openlineage-integration-common 1.27.0.
+acryl-datahub-airflow-plugin==0.15.0.5
```
sean-rose, line 3: Here are the additional package changes from installing this newer version of `acryl-datahub-airflow-plugin`:
```diff
-acryl-datahub==0.14.1.12
+acryl-datahub==0.15.0.5
+openlineage-airflow==1.25.0
+openlineage-integration-common==1.25.0
+openlineage-python==1.25.0
+openlineage_sql==1.25.0
+patchy==2.8.0
+sqlglot==25.32.1
+sqlglotrs==0.3.0
```

sean-rose: Immediately after this change was [deployed](https://github.com/mozilla-sre-deploy/deploy-telemetry-airflow/actions/runs/14232731267) at around 2025-04-03 01:12 I checked Airflow and it seemed to be running fine.  However, a bit later at 2025-04-03 01:48 @chelseybeck [reported](https://mozilla.slack.com/archives/C01E8GDG80N/p1743644881431479) that the Airflow webservers were offline.  I didn't see any specific errors to indicate this was the cause, but it seemed likely, so I reverted this in #2183 and Airflow has since recovered.

If this was indeed the cause, my guess is that it was due to the discrepancy between the OpenLineage 1.27 packages Airflow 2.10.5 was expecting and the OpenLineage 1.25 packages that this installed for `acryl-datahub-airflow-plugin` (i.e. that's a non-fungible version requirement).  Although we didn't install the `apache-airflow-providers-openlineage` package to formally enable Airflow's OpenLineage integration, maybe other Airflow code detected the core OpenLineage packages being installed and interacted with them poorly?

PR #2184, opened 2025-04-03
feat: Tag production images with `main-{commit SHA1}` for use in ArgoCD deployments (DSRE-1692)

sean-rose:
## Description
ArgoCD deployments are triggered by new images getting published, and ideally we'd like the image tags used for deployments to be unique per deployment (rather than using `latest`).  This PR tags production images with `main-{commit SHA1}` so ArgoCD can look for those tags to trigger deployments.

* DSRE-1692: Migrate Telemetery-airflow to GHA + ArgoCD
* https://github.com/mozilla-it/global-platform-admin/pull/2771


.circleci/config.yml
```diff
@@ -309,6 +312,7 @@ workflows:
       - publish-registry:
           name: Publish latest image to GAR
           registry: us-docker.pkg.dev/moz-fx-telemetry-airflow-prod/telemetry-airflow-prod
+          additional_tags: main-$CIRCLE_SHA1, latest
```
sean-rose, line 315: Note that I opted to use the full 40-character commit SHA1 string (rather than the 7-character commit SHA1 prefix being used elsewhere in this CI setup) to eliminate the possibility of tag collisions.

mikaeld: nit: this is a bit overkill as 7-char hash is 2^28 (over 250M) possible hashes. Tag collision probability is low enough. 7-char is far less overwhelming to read than 40-char. Eventually, we could use 8, 9 or even 10-char hash if we are truly worried about collisions. IMO 40-char is not very useful.

sean-rose: According to [these calculations](https://github.com/source-foundry/font-v/issues/2#issuecomment-327189455), with the current 2.3k commits currently in this repo there's already a 1% chance of 7-character commit prefix collisions, and that chance will grow non-linearly as commits are added over time (e.g. at 5.2k commits the chance goes up to 5%).

I'm fine continuing to use the 7-character commit prefix in places like log messages (where readability matters more), but personally I'd prefer to avoid any chance of collisions in situations like this where such collisions would actually matter (similar to how Git displays commit hash prefixes in most output to users but uses the full hashes for actually identifying commits and other objects).

kwindau (approved):

.circleci/config.yml
BenWu: What would happen if there's a collision?  Wouldn't it just untag the existing image and still deploy correctly?  `additional_tags` are based on the default 7 char tag anyway so they would still collide

BenWu (approved):
LGTM pending deciding on the tag length but it doesn't matter that much to me

.circleci/config.yml
sean-rose: > What would happen if there's a collision? Wouldn't it just untag the existing image and still deploy correctly?

I believe that's correct.  So the minor downsides are the previous image would end up without any tags, and there'd be some ambiguity about which commit is being referred to.  But I'd still rather just use the full commit hash in the tag and not worry about collisions, rather than optimizing for making the tags a bit more readable.

> `additional_tags` are based on the default 7 char tag anyway so they would still collide

If we stick with using the full commit hash here then my plan is to also change the default tags to use the full commit hash after we switch production deployments to ArgoCD (changing it now would break the existing GHA+Helm production deployments which look for the `sha-{commit SHA1 prefix}` tags).

BenWu: Ok I would lean towards keeping shorter but I don't think it's worth bikeshedding over.  I think this is fine

sean-rose: I wrote this per my own preferences, but I'm fine with following whatever the consensus is.  At this point that's two for shorter tags vs one for longer tags, so shorter tags wins.  I'll make that change and merge.

sean-rose: That sounds reasonable to me.  So how about after production is migrated to ArgoCD I proceed with changing the default `sha-*` tags to use the full commit hash (leaving the `main-*` tags using the commit hash prefix)?

PR #2189, opened 2025-04-14
fix: Fix GKE pod extra link for GKEPodOperator

BenWu:
## Description

This makes the gke pod link in the task instance details work.  This is a workaround for https://github.com/apache/airflow/issues/46658 pending that being fixed and us upgrading the provider version.  I tested this with a sandbox gke cluster with both deferrable and non-deferrable operators.

I'm open to opinions on whether this is worth doing


BenWu:
This is the link
image

scholtzan (approved):

PR #2204, opened 2025-05-28
fix: adjust fxci_pulse_export dag to run every 4 hours

ahal:
The RabbitMQ queue this pulls from builds up a little too high when run every 6 hours, and it ends up triggering some alerting.

Every 4 hours should be enough to keep the queue below the alert threshold.


scholtzan (approved):

dags/fxci_pulse_export.py
```diff
@@ -54,7 +54,7 @@
     "fxci_pulse_export",
     default_args=default_args,
     doc_md=__doc__,
-    schedule_interval="30 */6 * * *",
+    schedule_interval="30 */4 * * *",
```
sean-rose, line 57: Please be aware that with this schedule change being made without resetting the start date, if the `catchup` setting is ever enabled for this DAG then Airflow will schedule historical DAG runs for 04:30, 08:30, 16:30, and 20:30 for every date between 2024-07-08 and 2025-05-28.

PR #2206, opened 2025-05-29
fix: Stop using deprecated `is_delete_operator_pod` parameter

sean-rose:
## Description
Using the `is_delete_operator_pod` parameter is preventing the [fix](https://github.com/mozilla/telemetry-airflow/pull/2205) for DENG-8670 from working.

* DENG-8670: Clearing Airflow GKE pod operator tasks with `reattach_on_restart` enabled won't re-run the task's logic if a pod from a previous successful try exists
* https://github.com/mozilla/telemetry-airflow/pull/2205

### DIFF
--- a/dags/shredder.py
+++ b/dags/shredder.py
@@ -4,7 +4,7 @@ from airflow import DAG
 from kubernetes.client import models as k8s
 from timetable import MultiWeekTimetable
 
-from operators.gcp_container_operator import GKEPodOperator
+from operators.gcp_container_operator import GKEPodOperator, OnFinishAction
 from utils.tags import Tag
 
 docs = """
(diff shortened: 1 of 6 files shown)


dags/shredder.py
```diff
@@ -82,7 +82,7 @@
 ]
 common_task_args = {
     "image": docker_image,
-    "is_delete_operator_pod": True,
+    "on_finish_action": OnFinishAction.DELETE_POD.value,
```
BenWu, line 85: Shouldn't this work without the `.value`?

BenWu: Code is
```
elif on_finish_action is not None:
    kwargs["on_finish_action"] = OnFinishAction(on_finish_action)
```
I think `OnFinishAction(on_finish_action)` resolves correctly if you give it the enum object

sean-rose: While passing an `OnFinishAction` enum would indeed currently work (and as you noticed `GKEStartPodOperator` is itself passing an `OnFinishAction` enum to the `KubernetesPodOperator` constructor), the `GKEStartPodOperator` and `KubernetesPodOperator` constructors type hint their `on_finish_action` arguments as `str | None` and `str`, respectively, so I figured it was safer to abide by that and not assume passing an `OnFinishAction` enum would continue to work.

PR #2217, opened 2025-06-11
chore(GLAM): aggregates_new for scalar and histogram

edugfilho:
## Description

This PR complements https://github.com/mozilla/bigquery-etl/pull/7612

It creates clients_scalar_aggregates_new_v1 and clients_histogram_aggregates_new_v1 for the glam_fog and glam_fenix DAGs. The two tasks are extracted from clients_scalar_aggregates_v1 and clients_histogram_aggregates_v1 respectively. The intention behind this is to reduce resource contention on both DAGs. so if this refactor isn't enough I will add processing by samples on top of it.

* DENG-8798

### DIFF
--- a/dags/glam_fenix.py
+++ b/dags/glam_fenix.py
@@ -131,7 +131,7 @@ with DAG(
         scalar_wait_for_yesterdays_aggregates = ExternalTaskSensor(
             task_id=f"{product}_wait_for_yesterdays_scalar_aggregates",
             external_dag_id="glam_fenix",
-            external_task_id=f"{product}__clients_scalar_aggregates_v1",
+            external_task_id=f"query_{product}__clients_scalar_aggregates_v1",
             execution_delta=timedelta(days=1),
             mode="reschedule",
         )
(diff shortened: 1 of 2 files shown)


dags/glam_fenix.py
```diff
@@ -128,13 +128,38 @@
...
             task_name=f"{product}__clients_scalar_aggregates_v1"
         )
         clients_scalar_aggregates = query(
             task_name=f"{product}__clients_scalar_aggregates_v1"
         )
-
+        histogram_wait_for_yesterdays_aggregates = ExternalTaskSensor(
+            task_id=f"{product}_wait_for_yesterdays_histogram_aggregates",
+            external_dag_id="glam_fenix",
+            external_task_id=f"{product}__clients_histogram_aggregates_v1",
```
BenWu, line 153: Doesn't the task id start with init or query?

dags/glam_fenix.py
```diff
@@ -128,13 +128,38 @@
         )
         latest_versions = query(task_name=f"{product}__latest_versions_v1")
 
+        scalar_wait_for_yesterdays_aggregates = ExternalTaskSensor(
```
BenWu, line 131: What's the reason for doing it this way instead of `depends_on_past=True` with `max_active_runs=1`?

PR #2219, opened 2025-06-13
feat(backfills): [DENG-8647] Send slack alerts for backfill failures

BenWu:
## Description

Adds slack alerts when managed backfills fail.

I made a new slack workspace to do some janky testing:

image

* DENG-8647

### DIFF
--- a/dags/bqetl_backfill_complete.py
+++ b/dags/bqetl_backfill_complete.py
@@ -4,6 +4,7 @@ from datetime import datetime
 
 from airflow import DAG
 from airflow.decorators import task, task_group
+from airflow.providers.slack.notifications.slack import send_slack_notification
 from airflow.providers.slack.operators.slack import SlackAPIPostOperator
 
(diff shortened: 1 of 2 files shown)


dags/bqetl_backfill_complete.py
```diff
@@ -4,21 +4,23 @@
...
+DATA_PLATFORM_WG_CHANNEL_ID = "C01E8GDG80N"
 DOCKER_IMAGE = "gcr.io/moz-fx-data-airflow-prod-88e0/bigquery-etl:latest"
 
 tags = [Tag.ImpactTier.tier_3]
 
 default_args = {
     "email": [
         "ascholtz@mozilla.com",
-        "bewu@mozilla.com",
+        "benwu@mozilla.com",
```
BenWu, line 23: Changing my email alias because sendgrid seems to have bewu on some block list even after whd removed it

PR #2225, opened 2025-07-25
feat(jetstream_rerun): Add new DAG to trigger jetstream reruns

scholtzan:
## Description

During a Nimbus meeting it came up that some people are trying to trigger reruns by doing a noop on a experiment config (e.g. after an outcome changes, which doesn't trigger a rerun). It would be handy to have a DAG for these kinds of reruns, which would also allow for reruns of specific analysis periods

### DIFF
--- a/dags/jetstream_rerun.py
+++ b/dags/jetstream_rerun.py
@@ -1,0 +1,69 @@
+import datetime
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.tags import Tag
+
+docs = """
+### jetstream_rerun
+Build from telemetry-airflow repo, [dags/jetstream_rerun.py]
+Triggers rerun for Jetstream experiments.
+
+#### Owner
+ascholtz@mozilla.com
+mwilliams@mozilla.com
+"""
+
+tags = [Tag.ImpactTier.tier_3, Tag.Triage.no_triage]
+
+@dag(
+    dag_id="jetstream_rerun",
+    start_date=datetime.datetime(2025, 1, 1, 0, 0),
+    schedule_interval=None,
+    catchup=False,
+    doc_md=docs,
+    dagrun_timeout=datetime.timedelta(days=4),
+    tags=tags,
+    render_template_as_native_obj=True,
+    params={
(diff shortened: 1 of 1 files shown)


dags/jetstream_rerun.py
```diff
@@ -0,0 +1,69 @@
...
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.tags import Tag
+
+docs = """
+### jetstream_rerun
+Build from telemetry-airflow repo, [dags/jetstream_rerun.py]
+Triggers rerun for Jetstream experiments.
```
mikewilli, line 12: If this is meant to be run by non-DE users, it's probably a good idea to include some info here on how to run it.

dags/jetstream_rerun.py
```diff
@@ -0,0 +1,69 @@
...
+        "experiment_slug": Param("slug", title="Experiment Slug", type="string", description="Experiment slug to rerun"),
+        "recreate_enrollments": Param(False, title="Recreate Enrollments", type="boolean", description="Recreate enrollments option"),
+        "statistics_only": Param(False, title="Statistics Only", type="boolean", description="Statistics only option"),
+        "analysis_overall": Param(False, title="Analysis Overall", type="boolean", description="Include overall analysis period"),
+        "analysis_week": Param(False, title="Analysis Week", type="boolean", description="Include week analysis period"),
+        "analysis_day": Param(False, title="Analysis Day", type="boolean", description="Include day analysis period"),
+    },
+)
+def jetstream_rerun_dag():
+    @task
```
mikewilli, line 40: I think we need a second task in this DAG to export the results to GCS if the first task succeeds.

scholtzan: This DAG triggers Argo, which will export the statistics to GCS per https://github.com/mozilla/jetstream/blob/666142a349d0e2dfd322fa6aabf43b282ee6a731/jetstream/workflows/run.yaml#L120

mikewilli: Ah oops somehow missed that sorry!

mikewilli (approved):
Looks great, thanks!

PR #2234, opened 2025-08-07
fix(glam_fenix): fix external_task_id

edugfilho:
## Description

Fixes the `external_task_id` for a TaskSensor

### DIFF
--- a/dags/glam_fenix.py
+++ b/dags/glam_fenix.py
@@ -169,7 +169,7 @@ with DAG(
         if is_release:
             histogram_wait_for_yesterdays_aggregates = (
                 histogram_wait_for_yesterdays_aggregates_partial(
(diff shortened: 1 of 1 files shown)


dags/glam_fenix.py
```diff
@@ -169,7 +169,7 @@
         if is_release:
             histogram_wait_for_yesterdays_aggregates = (
                 histogram_wait_for_yesterdays_aggregates_partial(
-                    external_task_id="daily_release_done",
+                    external_task_id="org_mozilla_fenix_glam_release_done}",
```
kwindau, line 172: should it have that bracket at the end?

edugfilho, line 172:
```suggestion
                    external_task_id="org_mozilla_fenix_glam_release_done",
```

PR #2366, opened 2026-04-06
fix(artifact_deployment): continue view deploy if tables fail

BenWu:
## Description

This separates materialized view, table, and view deploys with `;` so the following commands run if any of them fail. This matches the previous behaviour of the publish_views task which had `trigger_rule=TriggerRule.ALL_DONE`

mozdata views currently don't get deployed if there are table deploy errors.

[verified in dev](https://dev.telemetry-airflow.nonprod.dataservices.mozgcp.net/log?dag_id=bqetl_artifact_deployment&task_id=publish_tables_and_views&execution_date=2026-04-06T17%3A46%3A38.869514%2B00%3A00) that the bash is valid

### DIFF
--- a/dags/bqetl_artifact_deployment.py
+++ b/dags/bqetl_artifact_deployment.py
@@ -33,6 +33,7 @@ found at the start of the logs in Airflow, e.g. `INFO - Found matching pod publi
 To find logs for specific datasets/views, add a string to search for to the end of the query.
 """
 
+import textwrap
 from datetime import datetime, timedelta
 
 from airflow import DAG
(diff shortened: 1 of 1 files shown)


dags/bqetl_artifact_deployment.py
```diff
@@ -162,18 +162,23 @@ def should_run_deployment(dag_id: str, generate_sql: bool) -> bool:
...
-            "script/bqetl view clean --skip-authorized --target-project=moz-fx-data-marketing-prod --project-id=moz-fx-data-marketing-prod && "
-            "script/bqetl view clean --skip-authorized --target-project=moz-fx-glam-prod --project-id=moz-fx-glam-prod && "
-            "script/bqetl view clean --skip-authorized --target-project=mozdata --user-facing-only && "
-            "script/bqetl view clean --skip-authorized --target-project=moz-fx-data-bq-people --project-id=moz-fx-data-bq-people && "
-            "script/publish_public_data_views --target-project=moz-fx-data-shared-prod && "
-            "script/publish_public_data_views --target-project=mozdata"
+            + """
+script/bqetl generate derived_view_schemas --use-cloud-function=false &&
+set +e;
+failed=0;
```
sean-rose: **issue (blocking):** It seems like this won't properly consider the job failed if either of the SQL generation commands fail.

I'd suggest putting the `set +e` and `failed=0` at the very start, then having the `generate_sql_cmd_template` (which ends with `&&`) combined with `script/bqetl generate derived_view_schemas --use-cloud-function=false || failed=1;`.

And given that this has gotten so complex, it could also make sense to make the non-SQL-generation-related artifact deployment logic a proper script in the `bigquery-etl` repo (e.g. `script/deploy_artifacts`).

dags/bqetl_artifact_deployment.py
```diff
@@ -162,18 +162,23 @@ def should_run_deployment(dag_id: str, generate_sql: bool) -> bool:
...
+script/bqetl view clean --skip-authorized --target-project=moz-fx-data-shared-prod &&
+script/bqetl view clean --skip-authorized --target-project=moz-fx-data-experiments --project-id=moz-fx-data-experiments &&
+script/bqetl view clean --skip-authorized --target-project=moz-fx-data-marketing-prod --project-id=moz-fx-data-marketing-prod &&
+script/bqetl view clean --skip-authorized --target-project=moz-fx-glam-prod --project-id=moz-fx-glam-prod &&
+script/bqetl view clean --skip-authorized --target-project=mozdata --user-facing-only &&
+script/bqetl view clean --skip-authorized --target-project=moz-fx-data-bq-people --project-id=moz-fx-data-bq-people &&
+script/publish_public_data_views --target-project=moz-fx-data-shared-prod &&
+script/publish_public_data_views --target-project=mozdata || failed=1;
+exit $failed
+            """
```
sean-rose: **suggestion (non-blocking):** I don't love this outdented string syntax, so I'd suggest using [`textwrap.dedent()`](https://docs.python.org/3.11/library/textwrap.html#textwrap.dedent) to avoid that.

dags/bqetl_artifact_deployment.py
sean-rose: > combined with `script/bqetl generate derived_view_schemas --use-cloud-function=false || failed=1;`.

Actually, now that I think of it the whole job should probably fail if SQL generation fails, otherwise `bqetl view clean` could remove views that don't get generated.

sean-rose (approved):
I'm changing my review to `r+wc` so you can proceed with this fix according to your best judgement to get artifact deployment unblocked.

dags/bqetl_artifact_deployment.py
BenWu: I ordered it this way so that it fails early without attempting deploys if the generate commands fail

PR #2367, opened 2026-04-07
chore: switch to cloudservices-aws-prod creds for crash emails

whd:
## Description

See the ticket for details. This needs to be landed in tandem with https://github.com/mozilla/python_mozetl/pull/408

* DENG-8707

### DIFF
--- a/dags/crash_symbolication.py
+++ b/dags/crash_symbolication.py
@@ -5,6 +5,7 @@ Generates correlations data for top crashers.
 
 Uses crash report data imported from Socorro.
 """
+
 import datetime
 
 from airflow import DAG
@@ -47,8 +48,10 @@ with DAG(
     tags=tags,
     doc_md=__doc__,
 ) as dag:
-    # modules_with_missing_symbols sends results as email
-    ses_aws_conn_id = "aws_data_iam_ses"
+    # modules_with_missing_symbols sends results as email via SES
+    # these credentials are shared with probe scraper but should be replaced by
+    # sendgrid credentials eventually
+    ses_aws_conn_id = "aws_prod_probe_scraper"
     ses_access_key, ses_secret_key, _ = AwsBaseHook(
         aws_conn_id=ses_aws_conn_id, client_type="s3"
     ).get_credentials()
--- a/resources/dev_connections.json
+++ b/resources/dev_connections.json
@@ -9,7 +9,7 @@
     "schema": "",
     "extra": "{\"known_hosts\": \"myhost.com AAAABBBBB\"}"
   },
-  "aws_data_iam_ses": {
+  "aws_prod_probe_scraper": {
     "conn_type": "s3",
     "description": null,
     "login": null,


dags/crash_symbolication.py
```diff
@@ -84,8 +77,9 @@
             additional_metadata={"PIP_PACKAGES": " ".join(PIP_PACKAGES)},
             additional_properties={
                 "spark:spark.jars": "gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar",
-                "spark-env:AWS_ACCESS_KEY_ID": ses_access_key,
-                "spark-env:AWS_SECRET_ACCESS_KEY": ses_secret_key,
+                "spark-env:SMTP_HOST": "smtp.sendgrid.net",
+                "spark-env:SMTP_USER": os.environ["AIRFLOW__SMTP__SMTP_USER"],
+                "spark-env:SMTP_PASSWORD": os.environ["AIRFLOW__SMTP__SMTP_PASSWORD"],
```
mikaeld: This is the [Secret Management](https://mozilla-hub.atlassian.net/wiki/spaces/IP/pages/27922811/WTMO+Developer+Guide#Secret-Management) guide. Per the flow chart, you end up in "Ask for help in #data-platform-infra-wg". Ultimately, this is what I would advise.

**Slight overview of why things are complicated here**: 

Those are [Airflow configuration values](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#smtp) and they cannot be referenced in DAGs for obvious security reasons. So `os.environ` will definitely never work.

Ideally, you would need to add either Variable or Connection objects and [insert them into the DAG via Jinja templating](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/connections.html). However, this 💩 DAG uses a convoluted wrapper of wrapper of wrapper of Dataproc Operators. Digging into this unnecessarily convoluted wrapping nonsense, we learn that the `additional_properties` [is not templated](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_modules/airflow/providers/google/cloud/operators/dataproc.html#DataprocCreateClusterOperator.template_fields).

That leaves us with manually inserting Variables/Connections in the DAG. This is far from ideal because it's quite easy to leak secrets as those will be rendered in clear text in the task instance details. As a last line of defense, Airflow introduced secrets redacting for anything with `secret`, `password` and such in its name. As you can see [here](https://workflow.telemetry.mozilla.org/task?dag_id=crash_symbolication.modules_with_missing_symbols&task_id=create_dataproc_cluster&execution_date=2026-04-06T05%3A00%3A00%2B00%3A00&map_index=-1), the AWS credentials are redacted `***` in the task instance details.

So you probably need to implement this via something like

```python
from airflow.models import Variable

def my_dag():
    ...
        "spark-env:SMTP_PASSWORD":  Variable.get("SMTP_PASSWORD"),
```

And hope that Airflow secrets redacting works as expected. The better alternatives would likely require refactoring the DAG quite a bit.

Maybe @BenWu or @sean-rose would have other suggestions, but that is my take on the subject.

TL;DR: RIP ☠️

BenWu: Is it a concern that anything in additional_properties is going to be visible in the created cluster's config? i.e. any viewer of the project can already see the secret values (don't git blame this dag).  If not, using an airflow variable and verifying with a test value that it redacts it correctly sounds reasonable to me.

Updating the job code to use GSM is probably feasible and not that much higher effort but I'd rather do the minimum for this dag

whd: Ok thanks for the reality check here. Since everything is terrible (we still also use SES for probe-scraper, but at least in an IAM that is not being decomm'd imminently) there's actually another option: reuse probe-scraper's SES via an AWS connection ID:

1. Create an AWS connection to replace `aws_data_iam_ses` that copies the probe scraper SES secrets from GSM
2. This PR becomes a one-liner to update `aws_data_iam_ses`  to the new conn id
3. https://github.com/mozilla/python_mozetl/pull/408/changes becomes a one-liner to change the source to be `telemetry-alerts@mozilla.com`

This way we don't solve the underlying problem of still using SES but collapse it into the probe scraper problem. Then I will file a separate ticket for DE to consolidate on sendgrid `airflow-alerts@telemetry.mozilla.org` and probably clean up the mechanism for sending emails.

EDIT: https://mozilla-hub.atlassian.net/browse/DENG-10931

whd: And indeed this `aws_prod_probe_scraper` already exists with the same key so (1) can reuse that key. Cursed but at least consolidated.

whd: Filed https://mozilla-hub.atlassian.net/browse/DENG-10931 for the backlog

mikaeld: > any viewer of the project can already see the secret values

@BenWu any viewer on the Airflow instance can view DAG instance details and may be able to extract secrets if they aren't properly redacted. Anyone with a mozilla.com account can create an Airflow account with viewer access. 

A few years ago, security had a (red? purple? whatever color) security team investigating this and [we had to fix a bunch of DAGs exposing secrets](https://mozilla-hub.atlassian.net/browse/DSRE-1518). A lot of those weren't caught by Airflow's auto-redacting of sensitive values. So this last line of defense has its limits.

I don't think secrets being visible in the additional properties in the Dataproc project is really a concern though.

mikaeld (approved):

PR #2370, opened 2026-04-14
feat: Add live table partition sensors in `copy_deduplicate` DAG

sean-rose:
## Description
As a safeguard against possible data ingestion delays ([example](https://mozilla-hub.atlassian.net/browse/SVCSE-4143?focusedCommentId=1367789)).

* [SVCSE-4143](https://mozilla-hub.atlassian.net/browse/SVCSE-4143): Migrate ingestion-sink to GCPv2
* DENG-10849: Define SLOs for DE-managed telemetry ingestion services

[SVCSE-4143]: https://mozilla-hub.atlassian.net/browse/SVCSE-4143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -3,6 +3,9 @@ import re
 
 from airflow import models
 from airflow.operators.empty import EmptyOperator
+from airflow.providers.google.cloud.sensors.bigquery import (
+    BigQueryTablePartitionExistenceSensor,
+)
 from airflow.sensors.external_task import ExternalTaskMarker
 from airflow.utils.task_group import TaskGroup
 from kubernetes.client import models as k8s
@@ -67,6 +70,36 @@ with models.DAG(
     default_args=default_args,
     tags=tags,
 ) as dag:
+    # Normally the live tables will contain complete data for the previous day when this DAG runs at 01:00 UTC.
+    # However, as a safeguard against possible data ingestion delays, we wait until certain live tables contain
+    # data for the current day as an indication that the live data for the previous day is actually complete.
+    live_tables_to_wait_for = {
+        # Legacy
+        "telemetry": ["event_v4", "first_shutdown_v5", "main_v5"],
+        # Glean
+        "firefox_desktop": ["events_v1", "metrics_v1"],
+        "org_mozilla_firefox": ["events_v1", "metrics_v1"],
+        "org_mozilla_ios_firefox": ["events_v1", "metrics_v1"],
+        # Glean server
+        "accounts_backend": ["events_v1"],
+    }
+    with TaskGroup("wait_for_complete_live_data"):
+        live_table_sensors = {
+            f"{app_name}_live.{live_table}": BigQueryTablePartitionExistenceSensor(
+                task_id=f"wait_for_complete_{app_name}_live_{live_table}",
+                project_id="moz-fx-data-shared-prod",
+                dataset_id=f"{app_name}_live",
+                table_id=live_table,
+                partition_id="{{ data_interval_end | ds_nodash }}",
+                gcp_conn_id="google_cloud_shared_prod",
+                deferrable=True,
+                poke_interval=datetime.timedelta(minutes=5),
+                timeout=datetime.timedelta(hours=4),
+            )
+            for app_name, live_tables in live_tables_to_wait_for.items()
+            for live_table in live_tables
+        }
+
     # This single task is responsible for sequentially running copy queries
     # over all the tables in _live datasets into _stable datasets except those
     # that are specifically used in another DAG.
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
```diff
@@ -67,6 +70,36 @@
...
+            f"{app_name}_live.{live_table}": BigQueryTablePartitionExistenceSensor(
+                task_id=f"wait_for_complete_{app_name}_live_{live_table}",
+                project_id="moz-fx-data-shared-prod",
+                dataset_id=f"{app_name}_live",
+                table_id=live_table,
+                partition_id="{{ data_interval_end | ds_nodash }}",
+                gcp_conn_id="google_cloud_shared_prod",
+                deferrable=True,
+                poke_interval=datetime.timedelta(minutes=5),
+                timeout=datetime.timedelta(hours=4),
```
sean-rose, line 97: I picked 4 hours as the timeout somewhat arbitrarily, so if there's some reason you feel the timeout should be different I'm happy to discuss.

akkomar: That would time out with notification at 5/9/13 UTC, LGTM.

akkomar (approved):

dags/copy_deduplicate.py
sean-rose: @akkomar actually sensors don't retry, so it would just time out and fail at 05:00 UTC.

I was initially planning on having it be 8 hours (timing out at 09:00 UTC) but then reconsidered, thinking maybe that was too long.

@scholtzan / @BenWu, do either of you want to weigh in on this as well?

scholtzan (approved):
This looks reasonable to me

dags/copy_deduplicate.py
```diff
@@ -67,6 +70,36 @@
     default_args=default_args,
     tags=tags,
 ) as dag:
+    # Normally the live tables will contain complete data for the previous day when this DAG runs at 01:00 UTC.
+    # However, as a safeguard against possible data ingestion delays, we wait until certain live tables contain
+    # data for the current day as an indication that the live data for the previous day is actually complete.
```
BenWu, line 75: I don't think this is true because there are no ordering guarantees so if there's a backlog, data for the current day can (probably will) be written before the previous day is complete.  The partition not existing is an indication that we shouldn't run copy_dedupe, but its existence doesn't mean that it's safe to run.  So with this in place, I still wouldn't trust that it prevents copy-deduping incomplete data unless there's some additional delay after it succeeds.

Taking into account the actual pipeline latency (dataflow watermark age, pubsub oldest unacked message) would be more robust.

sean-rose: @whd recently mentioned some old tickets that concur with your suggested Pub/Sub metrics approach:
- https://github.com/mozilla/bigquery-etl/issues/493
- [Bug 1761102](https://bugzilla.mozilla.org/show_bug.cgi?id=1761102): Programmatically delay copy-deduplicate when pipeline ingestion latency is detected

However, this approach is certainly simpler to implement for now using ready-made sensors, and might even still be worth doing in addition to the Pub/Sub metrics approach to cover the scenario if there's an Ingestion Edge or Cloud Logging issue preventing data from being submitted to Pub/Sub in the first place (at least I didn't see any [Pub/Sub metrics](https://docs.cloud.google.com/monitoring/api/metrics_gcp_p_z#gcp-pubsub) that would tell us when the latest message was received).

@akkomar / @scholtzan / @BenWu we should come to some consensus on how to proceed:
1. Proceed with this PR as is.
2. Proceed with using partition sensors but with an additional delay added after the sensors succeed (in which case we'd either want to reschedule `copy_deduplicate` to start that much earlier, or move the sensors to a separate DAG that starts that much earlier, so the main `copy_deduplicate` tasks would normally still start at 01:00).
3. Don't bother proceeding with this approach, just use the Pub/Sub metrics approach.

Votes/thoughts?

BenWu: I think you're right that there isn't an easy way to get the most recent pubsub message, but even if we could, that doesn't cover every edge case.  To not overcomplicate things, I'm ok with merging this as-is to cover the extreme case of nothing making it to pubsub near the start of the pipeline, then later adding a sensor based on latency metrics.

I think we just need to be clear on what this doesn't do so we don't look at it and assume we're safe from ingestion latency.  I'm not sure that this partition sensor will be that useful as-is but I don't see any downsides apart from potential minor confusion

scholtzan: I agree in going with option 1. Even though this won't guarantee that there are no issues it's still an improvement to the current state. We can consider replacing it if we implement a better approach in the future.
Might be worth adding a comment as Ben mentioned, what this approach covers and what it does not.

PR #2371, opened 2026-04-16
fix(adm-export): updating schedule to match upstream with 0 execution delta

chelseyklein:
## Description
This PR fixes a schedule mismatch from changing the upstream:
- https://github.com/mozilla/bigquery-etl/pull/9204

* [AD-1342](https://mozilla-hub.atlassian.net/browse/AD-1342)

[AD-1342]: https://mozilla-hub.atlassian.net/browse/AD-1342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/adm_dma_export.py
+++ b/dags/adm_dma_export.py
@@ -45,7 +45,7 @@ adm_sftp_secret = Secret(
 
 with DAG(
     dag_name,
-    schedule_interval="0 5 * * *",
+    schedule_interval="0 8 * * *",
     doc_md=DOCS,
     default_args=default_args,
     tags=tags,
@@ -83,7 +83,7 @@ with DAG(
         task_id="wait_for_adm_daily_dma_aggregates",
         external_dag_id="bqetl_search_terms_daily",
         external_task_id="search_terms_derived__adm_daily_dma_aggregates__v1",
-        execution_delta=datetime.timedelta(hours=2),
+        execution_delta=datetime.timedelta(hours=0),
         mode="reschedule",
         allowed_states=ALLOWED_STATES,
         failed_states=FAILED_STATES,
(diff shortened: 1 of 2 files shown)


dags/adm_dma_export.py
```diff
@@ -83,7 +83,7 @@
         task_id="wait_for_adm_daily_dma_aggregates",
         external_dag_id="bqetl_search_terms_daily",
         external_task_id="search_terms_derived__adm_daily_dma_aggregates__v1",
-        execution_delta=datetime.timedelta(hours=2),
+        execution_delta=datetime.timedelta(hours=0),
```
sean-rose: **note:** You could also simply omit `execution_delta` when the DAG schedules are the same.

dags/adm_dma_export.py
sean-rose: **suggestion (non-blocking):** This `wait_for_clients_daily_export` variable name doesn't correspond with the task that's actually being awaited (guessing it was copy-pasted from an unrelated DAG).  It'd be helpful to have a more accurate variable name.

dags/adm_export.py
sean-rose, line 1: The same comments for `adm_dma_export.py` also apply to this file.

sean-rose (approved):
r+wc

dags/adm_dma_export.py
chelseyklein: ah, yes, makes sense ✅

dags/adm_dma_export.py
chelseyklein: ah, yeah, i saw that too and neglected to fix... ✅

dags/adm_export.py
chelseyklein: ✅

PR #2372, opened 2026-04-17
fix: Limit `copy_deduplicate` to one active run and enable catchup

sean-rose:
## Description
To recover from a data ingestion incident ([IIM-153](https://mozilla-hub.atlassian.net/browse/IIM-153)) we're going to need to rerun `copy_deduplicate` for multiple days, and we don't want more than one run active at a time because they're resource intensive.

* [IIM-153](https://mozilla-hub.atlassian.net/browse/IIM-153): Missing Equative Impressions (2026-04-17)

[IIM-153]: https://mozilla-hub.atlassian.net/browse/IIM-153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IIM-153]: https://mozilla-hub.atlassian.net/browse/IIM-153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -63,6 +63,8 @@ tags = [Tag.ImpactTier.tier_1]
 with models.DAG(
     dag_name,
     schedule_interval="0 1 * * *",
+    max_active_runs=1,
+    catchup=True,
     doc_md=DOCS,
     default_args=default_args,
     tags=tags,


dags/copy_deduplicate.py
```diff
@@ -63,6 +63,8 @@
 with models.DAG(
     dag_name,
     schedule_interval="0 1 * * *",
+    max_active_runs=1,
+    catchup=True,
```
scholtzan, line 67: **question**: What's the reason for enabling catchup here? Could we just clear the affected days and let them run?

scholtzan (approved):

dags/copy_deduplicate.py
sean-rose: Because if we limit it to one active run and there's an active (backfill) run at 01:00 when it's normally scheduled to start then Airflow won't create/queue the DAG run with catchup disabled, it'd skip the DAG run.

PR #2375, opened 2026-04-20
fix(adm_exports): Add check_existence to ExternalTaskSensors

akkomar:
Adds `check_existence=True` to the `ExternalTaskSensors` in `adm_dma_export` and `adm_export`.
Without this flag, if the upstream DAG run doesn't exist for a given logical date, the sensor polls forever instead of failing fast.


edugfilho (approved):

sean-rose:
> Adds `check_existence=True` to the `ExternalTaskSensors` in `adm_dma_export` and `adm_export`. Without this flag, if the upstream DAG run doesn't exist for a given logical date, the sensor polls forever instead of failing fast.

`check_existence` is worth specifying, but it doesn't work as you described.  It only verifies that the DAG and task being awaited exist, not that the specific DAG _run_ or task _instance_ being awaited exist.

PR #2380, opened 2026-04-23
feat: [DENG-10973] Use bqetl generated copy_dedupe task markers

BenWu:
## Description

Replaces `ExternalTaskMarker`s for bqetl generated tasks with the task groups in https://github.com/mozilla/bigquery-etl/pull/9242.

~~Draft for now and leaving the existing task markers in because I want to verify the output after https://github.com/mozilla/bigquery-etl/pull/9242 is merged.~~

* DENG-10973

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -231,8 +231,8 @@ with models.DAG(
                 external_dag_id=downstream_dependency[0],
                 external_task_id="wait_for_copy_deduplicate_first_shutdown_ping",
                 execution_date="{{ execution_date.replace("
-                               + downstream_dependency[1]
-                               + ").isoformat() }}",
+                + downstream_dependency[1]
+                + ").isoformat() }}",
             )
 
         ExternalTaskMarker(
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -92,7 +92,10 @@ with models.DAG(
             "telemetry_live.first_shutdown_v5",
             "firefox_desktop_live.metrics_v1",
             *column_removal_backfill_tables_live,
-            *[re.sub(r"_v1$", "_v2", table) for table in column_removal_backfill_tables_live],
+            *[
+                re.sub(r"_v1$", "_v2", table)
+                for table in column_removal_backfill_tables_live
+            ],
         ],
         container_resources=resources,
     )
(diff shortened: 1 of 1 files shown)


dags/copy_deduplicate.py
sean-rose: **suggestion (non-blocking):** Such hard-coded external task markers for `bqetl_*` and `private_bqetl_*` DAGs can now be removed.

dags/copy_deduplicate.py
```diff
@@ -208,6 +208,14 @@
 
         copy_deduplicate_all >> copy_deduplicate_all_external
 
+    copy_deduplicate_all_task_markers = ExternalTaskMarker(
+        task_id="copy_deduplicate_task_markers__wait_for_copy_deduplicate_all",
+        external_dag_id="copy_deduplicate_task_markers",
+        external_task_id="wait_for_copy_deduplicate_all",
+        execution_date="{{ execution_date.isoformat() }}",
+    )
+    copy_deduplicate_all >> copy_deduplicate_all_task_markers
```
sean-rose: **suggestion (non-blocking):** IMO it would make sense for such external task markers to be defined within the `*_external` task groups (and then no additional dependencies would need to be defined).

dags/copy_deduplicate.py
```diff
@@ -276,7 +222,7 @@
 
     with TaskGroup("first_shutdown_ping_external") as first_shutdown_ping_external:
         downstream_dependencies = {
-            ("bqetl_analytics_tables", "hour=2, minute=0"),
+            # existing tasks are covered by copy_deduplicate_task_markers
```
BenWu, line 225: Leaving this in but empty so it's easy to manually add things (although that seems unlikely given this is legacy telemetry)

BenWu:
Verified the generated dag in https://github.com/mozilla/bigquery-etl/blob/a58ab1eba66659a315d00a0c5c39b618915f5617/dags/copy_deduplicate_task_markers.py.  It also covers private_bqetl which I didn't test.  I removed all the bqetl dags and also uncommented the glam markers per the comment.

dags/copy_deduplicate.py
sean-rose: **note:** This technically ends up as an empty dictionary now, rather than an empty set, but that isn't actually consequential with the current logic.

dags/copy_deduplicate.py
```diff
@@ -405,8 +355,7 @@
         "core_clients_first_seen_external"
     ) as core_clients_first_seen_external:
         downstream_dependencies = {
-            ("bqetl_core", "hour=2, minute=0"),
-            ("bqetl_glean_usage", "hour=2, minute=0"),
+            # existing tasks are covered by copy_deduplicate_task_markers
         }
```
sean-rose, line 359: ditto

sean-rose (approved):
r+wc

PR #2394, opened 2026-06-25
fix: Add hackbot secrets to webcompat kb DAG

ksy36:


ksy36:
@scholtzan would you be able to merge this add the secrets to airflow, please? I'll share them with you

scholtzan (approved):
Sure, you can share the with my via 1Password

scholtzan:
I added the secrets

PR #2409, opened 2026-08-06
chore(adm_export): [DENG-11459] Shift adm export DAGs to 10:00 UTC to stay aligned with bqetl_search_terms_daily

chelseyklein:
## Summary
Shifts the `adm_export` and `adm_dma_export` DAGs from `0 8 * * *` to `0 10 * * *` to stay aligned with the `bqetl_search_terms_daily` schedule change (moving 08:00 → 10:00 UTC) in bigquery-etl.

Both DAGs' `ExternalTaskSensor`s wait on `bqetl_search_terms_daily` tasks using the default `execution_delta=0`, which requires all three DAGs to share the same execution_date. Keeping them on the same 10:00 schedule preserves that alignment; the sensors keep working exactly as before, exports just start 2h later.

## Changes
- `dags/adm_export.py`: `0 8 * * *` → `0 10 * * *`
- `dags/adm_dma_export.py`: `0 8 * * *` → `0 10 * * *`

## ⚠️ Deploy coordination
This must land together with the corresponding bigquery-etl `dags.yaml` change (mozilla/bigquery-etl#9767), within the same daily window. Because the sensors match on `execution_date` with a 0 delta, deploying either side alone leaves a one-run desync where the export sensor looks for a search_terms run that doesn't exist (`check_existence=True` → failure). Both should be live before the next scheduled trigger.

Ticket: DENG-11459

### DIFF
--- a/dags/adm_dma_export.py
+++ b/dags/adm_dma_export.py
@@ -45,7 +45,7 @@ adm_sftp_secret = Secret(
 
 with DAG(
     dag_name,
-    schedule_interval="0 8 * * *",
+    schedule_interval="0 10 * * *",
     doc_md=DOCS,
     default_args=default_args,
     tags=tags,
--- a/dags/adm_export.py
+++ b/dags/adm_export.py
@@ -44,7 +44,7 @@ adm_sftp_secret = Secret(
 
 with DAG(
     dag_name,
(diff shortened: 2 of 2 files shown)


csadilek (approved):
Thank you! r+ for change control.

PR #2410, opened 2026-08-10
fix(copy_dedupe): Avoid processing all tables if given empty only list

BenWu:
## Description

follow-up to https://github.com/mozilla/telemetry-airflow/pull/2378. passing an empty list in `only_tables` to the copy_deduplicate function makes it do all tables

### DIFF
--- a/utils/gcp.py
+++ b/utils/gcp.py
@@ -315,6 +315,12 @@ def bigquery_etl_copy_deduplicate(
     :return: GKEPodOperator
     """
     kwargs["name"] = kwargs.get("name", task_id.replace("_", "-"))
+    # An empty only_tables list emits no --only args, which makes bqetl process every live table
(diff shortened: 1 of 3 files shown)


scholtzan (approved):

utils/gcp.py
```diff
@@ -315,6 +315,12 @@ def bigquery_etl_copy_deduplicate(
     :return: GKEPodOperator
     """
     kwargs["name"] = kwargs.get("name", task_id.replace("_", "-"))
+    # An empty only_tables list emits no --only args, which makes bqetl process every live table
+    # instead of none. Probably not intended, so fail.
+    if only_tables is not None and len(only_tables) == 0:
+        raise ValueError(
```
BenWu, line 321: This error is only at parse time so CI would get it

PR #2423, opened 2026-09-03
feat(highwind): Add DAG to schedule the Highwind experiment analysis job

jaredlockhart:
## Description

Schedules the Highwind proof of concept ([mozilla/docker-etl#578](https://github.com/mozilla/docker-etl/pull/578)), which computes CUPED-adjusted sequential confidence intervals for live Firefox Desktop experiments from per-branch sufficient statistics aggregated in BigQuery. It runs alongside Jetstream and replaces nothing.

Because

* The Highwind proof of concept job lands in docker-etl, which publishes its container image, and nothing schedules that image.
* A run has to wait on the specific upstream tables it reads rather than on a fixed offset, since it aggregates `clients_daily`, `clients_last_seen` and `search_clients_daily`.

This commit

* Adds a daily DAG running the highwind docker-etl image for one run date.
* Waits on the three upstream bqetl tasks the job reads, via `ExternalTaskSensor`.
* Caps the DAG at one active run and disables retries, since a run is a full recompute of every live experiment.

Two things worth a reviewer's attention.

**The permissions are not in place yet.** The job writes tables to the `highwind_poc` dataset in `moz-fx-data-experiments`, and one JSON blob per experiment to `gs://mozanalysis/highwind/`. No DAG in this repo sets `service_account_name`, so the pod runs as the default identity on `workloads-prod-v1`, which needs `bigquery.tables.create` plus write on that dataset, and write on that bucket prefix. If someone can point me at the right identity I will request the grant. Until it lands the task fails at the write step, so this should not be unpaused on merge.

**Retries are off deliberately.** A run recomputes every live experiment, so a retry is a second full run's worth of scanning, and the job already isolates per-experiment failures and exits non-zero only when every experiment failed. Happy to add them back if the house style is to keep them regardless.

* [EXP-7482](https://mozilla-hub.atlassian.net/browse/EXP-7482) (this DAG)
* [EXP-7390](https://mozilla-hub.atlassian.net/browse/EXP-7390) (Highwind proof of concept epic)
* [mozilla/docker-etl#578](https://github.com/mozilla/docker-etl/pull/578) (the job this schedules)

[EXP-7482]: https://mozilla-hub.atlassian.net/browse/EXP-7482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/highwind.py
+++ b/dags/highwind.py
@@ -1,0 +1,103 @@
+"""
+See [highwind in the docker-etl repository](https://github.com/mozilla/docker-etl/tree/main/jobs/highwind).
+
+Highwind is a proof of concept for Nimbus experiment analysis. It aggregates per-branch sufficient
+statistics for Firefox Desktop guardrail metrics in BigQuery, then computes CUPED-adjusted
+sequential confidence intervals from them. One run is a full recompute of every live Firefox
+Desktop experiment: it reads the experiment mirror, aggregates each metric over windows anchored to
+each analysis unit's own enrollment, and writes results to the `highwind_poc` dataset in
+`moz-fx-data-experiments` plus one JSON blob per experiment.
+
+It runs alongside Jetstream and replaces nothing.
+"""
+
+from collections import namedtuple
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.sensors.external_task import ExternalTaskSensor
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.constants import ALLOWED_STATES, FAILED_STATES
+from utils.tags import Tag
+
+default_args = {
+    "owner": "jkerim@mozilla.com",
+    "email": ["jkerim@mozilla.com", "telemetry-alerts@mozilla.com"],
+    "depends_on_past": False,
+    "start_date": datetime(2026, 9, 3),
+    "email_on_failure": True,
+    "email_on_retry": False,
+    # A run recomputes every live experiment, so a retry costs a second full run's worth of
+    # scanning. The job already isolates per-experiment failures and exits non-zero only when every
+    # experiment failed, which is not a condition a retry fixes.
+    "retries": 0,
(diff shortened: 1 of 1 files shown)


dags/highwind.py
```diff
@@ -0,0 +1,103 @@
...
+from airflow import DAG
+from airflow.sensors.external_task import ExternalTaskSensor
+
+from operators.gcp_container_operator import GKEPodOperator
+from utils.constants import ALLOWED_STATES, FAILED_STATES
+from utils.tags import Tag
+
+default_args = {
+    "owner": "jkerim@mozilla.com",
+    "email": ["jkerim@mozilla.com", "telemetry-alerts@mozilla.com"],
```
mikewilli: Do you want Yashika on here too?

dags/highwind.py
```diff
@@ -0,0 +1,103 @@
...
+
+Highwind is a proof of concept for Nimbus experiment analysis. It aggregates per-branch sufficient
+statistics for Firefox Desktop guardrail metrics in BigQuery, then computes CUPED-adjusted
+sequential confidence intervals from them. One run is a full recompute of every live Firefox
+Desktop experiment: it reads the experiment mirror, aggregates each metric over windows anchored to
+each analysis unit's own enrollment, and writes results to the `highwind_poc` dataset in
+`moz-fx-data-experiments` plus one JSON blob per experiment.
+
+It runs alongside Jetstream and replaces nothing.
+"""
```
mikewilli, line 12: It would be good to also include a short Triage Notes section so that the engineer on triage knows whether to take action, and what to do. Or if you don't want them to take any action, add the `no_triage` tag instead.

mikewilli (approved):
Couple comments but looks good to me

PR #2424, opened 2026-09-03
chore(stable-write): Use stable-write SA in shredder and copy_dedupe

scholtzan:
## Description

Depends on https://github.com/mozilla/dataservices-infra/pull/2261

This is to update the `copy_dedupe` and `shredder` DAGs to use the SA that has dedicated write access to stable tables.

* proposal: https://docs.google.com/document/d/1KhWaH-o7EDG1PaOR7JKrGWyP8hmdtodCQRino-PKbA0/edit?tab=t.0

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @mozilla/dataplatform-wg
--- a/dags/shredder.py
+++ b/dags/shredder.py
@@ -95,6 +95,7 @@ common_task_args = {
 telemetry_main = GKEPodOperator(
     task_id="telemetry_main",
     name="shredder-telemetry-main",
+    service_account_name="stable-table-editor",
     arguments=[
         *base_command,
         "--parallelism=2",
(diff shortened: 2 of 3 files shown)


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @mozilla/dataplatform-wg
```
scholtzan: Adding dataplatform-wg as codeowner on all files since certain service accounts have access to sensitive data and can write to stable. So a DPE/DE needs to take a look before merging

scholtzan:
Tested this change on the first_shutdown_ping task yesterday and it has been running fine (https://github.com/mozilla/telemetry-airflow/pull/2425). So it's good to get used for other tasks as well

dags/shredder.py
BenWu, line 88: This should work too, although it applies it to the experiments task which doesn't need the account
```suggestion
    "dag": dag,
    "service_account_name": "stable-table-editor",
```

BenWu (approved):

PR #2425, opened 2026-09-08
chore(stable-editor): Use stable-table-editor SA for first_shutdown_ping

scholtzan:
## Description

This is to test the new `stable-table-editor` SA on the `copy_deduplicate_first_shutdown_ping` before using it for other tasks that write to stable tables (https://github.com/mozilla/telemetry-airflow/pull/2424/changes). 

I created a new k8s service account via `kubectl create serviceaccount stable-table-editor -n default` and ran `kubectl annotate serviceaccount --namespace default stable-table-editor iam.gke.io/gcp-service-account=stable-table-editor@moz-fx-data-airflow-gke-prod.iam.gserviceaccount.com` so it can be referenced as `stable-table-editor` in the Airflow task definition.

Currently, both the default SA and `stable-table-editor` still have write access to stable tables.

### DIFF
--- a/dags/copy_deduplicate.py
+++ b/dags/copy_deduplicate.py
@@ -213,6 +213,7 @@ with models.DAG(
         task_id="copy_deduplicate_first_shutdown_ping",
         target_project_id="moz-fx-data-shared-prod",
         billing_projects=("moz-fx-data-shared-prod",),
+        service_account_name="stable-table-editor",
         only_tables=[
             "telemetry_live.first_shutdown_use_counter_v4",
             "telemetry_live.first_shutdown_v5",


BenWu (dismissed):

dags/copy_deduplicate.py
```diff
@@ -221,15 +213,14 @@
         task_id="copy_deduplicate_first_shutdown_ping",
         target_project_id="moz-fx-data-shared-prod",
         billing_projects=("moz-fx-data-shared-prod",),
+        service_account_name="stable-table-editor",
```
BenWu, line 216: Actually shouldn't this be using `gcp_conn_id`?

BenWu (approved):
nvm read the PR description again. I'm not completely sure how it works but worth a try

whd:
I don't think https://github.com/mozilla-services/cloudops-infra/pull/6995 ever landed so I wouldn't expect the new SA to have write access yet.

EDIT: actually the new SA also got added to `workgroup:dataplatform/airflow` so it would get access, but only after https://ops-master.jenkinsv2.prod.mozaws.net/job/gcp-pipelines/job/data-shared/job/access-groups/78012/ finishes propagating.

PR #2427, opened 2026-09-10
Revert "chore(stable-write): Use stable-write SA in shredder and copy_dedupe"

scholtzan:
Reverts mozilla/telemetry-airflow#2424

Issues with acessing accounts_backend_live


whd:
I traced this to two issues:

1. Similar to https://github.com/mozilla-services/cloudops-infra/pull/6995#issuecomment-5533874867, I think I made a mistake when moving workgroups to GCPv2 where the `live` label was dropped from https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/global/access-groups/main.tf#L106. It is present on `live_restricted` however: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/global/access-groups/main.tf#L114. It should be applied on both.

Ordinarily you wouldn't notice this because `live` datasets always also grant reader access to [workgroup:mozilla-confidential/data-viewers](https://dawg.quick.mozilla.cloud/workgroup/mozilla-confidential#data-viewers) which includes Airflow SAs. But it turns out:

2. `mozilla-confidential/data-viewers` also enumerates the airflow subgroups and not `airflow` which is a composite group: https://github.com/mozilla/global-platform-admin/blob/main/google-workspace-management/tf/workgroups/mozilla-confidential.yaml#L13-L15. I believe this has something to do with composite group nesting not working in GCPv2 the way it did in GCPv1. So adding the new SA to https://dawg.quick.mozilla.cloud/workgroup/dataplatform#airflow did not add it to `mozilla-confidential/data-viewers`, you need to add `airflow-stable-table-editor` as well.

Fixing either of these should be sufficient to fix the access issue, but we should probably fix both.

scholtzan:
1. I added the live label as part of the draft PR https://github.com/mozilla-services/cloudops-infra/pull/6995#issuecomment-5533874867
2. Opened https://github.com/mozilla/global-platform-admin/pull/7328 to add the SA/subgroup to `mozilla-confidential/data-viewers`

PR #2428, opened 2026-09-10
chore(stable-write): Use stable-write SA in shredder and copy_dedupe

scholtzan:
Reverts mozilla/telemetry-airflow#2427

The live table access issue has been fixed via https://github.com/mozilla/global-platform-admin/pull/7328


dags/shredder.py
```diff
@@ -86,6 +86,7 @@
     "on_finish_action": OnFinishAction.DELETE_POD.value,
     "reattach_on_restart": True,
     "dag": dag,
+    "service_account_name": "stable-table-editor",
```
sean-rose, line 89: **question:** Since shredder applies to more than just stable tables, I presume the `stable-table-editor` service account has permission to edit all of our BigQuery tables?

sean-rose (approved):
r+wc

dags/shredder.py
scholtzan: Right. The service account has edit permission to any stable and derived table (what the current default Airflow account has atm)

PR #2429, opened 2026-09-10
fix(highwind): move the schedule off peak to 23:00 UTC

jaredlockhart:
## Description

Because

* Every run recomputes the same set of experiments, so its cost is fixed, but its wall time depends on how many slots it can get, and the 11:00 slot sits in the busiest window on the shared analysis-and-etl reservation.
* Wall time grew from about half an hour, to two hours, to over four across the first three runs, and the third exceeded execution_timeout and was killed having written nothing.
* 11:00 is also earlier than the upstreams reliably finish. Over 21 days search_clients_daily has a p95 of 13:19 and has run as late as 19:21, so the sensors can hold for hours.

This commit

* Moves the schedule to 23:00 UTC, the quietest hour the upstream dependencies allow, so a run contends for slots with far less neighbouring work.
* Recomputes each sensor's execution_delta for the new hour, so they still resolve to bqetl_main_summary at 02:00 and bqetl_search at 03:00.

* https://mozilla-hub.atlassian.net/browse/EXP-7539

### DIFF
--- a/dags/highwind.py
+++ b/dags/highwind.py
@@ -51,28 +51,30 @@ UPSTREAM = [
         "clients_daily",
         "bqetl_main_summary",
         "telemetry_derived__clients_daily__v6",
-        timedelta(hours=9),
+        timedelta(hours=21),
     ),
     Upstream(
         "clients_last_seen",
         "bqetl_main_summary",
         "telemetry_derived__clients_last_seen__v2",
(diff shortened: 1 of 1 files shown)


jaredlockhart:
Closing in favour of https://github.com/mozilla/telemetry-airflow/pull/2430

PR #2431, opened 2026-09-17
Bug 2064326 - run the probe_scraper one day later

Archaeopteryx:
Firefox's version gets increased on Thursday UTC every two weeks since the switch to the two-week release cycle.
probe_scraper will run late on Thursday PT (early Friday UTC).

## Description

This PR moves the day the probe_scraper checks for probes which will expires soon by one day to align with the new two-week release cycle of Firefox. 

* https://mozilla-hub.atlassian.net/browse/DENG-11512


badboy:
If I'm not mistaken we actually do need to adjut probe-scraper as well: https://github.com/mozilla/probe-scraper/blob/e0d9da1d5564f5cea166f8c356615d13c21ea5ad/probe_scraper/ping_expiry_alert.py#L316

 cc @chutten

chutten:
That probe-scraper piece is actually the "Glean Pings Expiring for X" email about stable table retention (warning folks that old data will be deleted, the first time that happens for a new ping). It's based entirely on calendars and so can remain untouched.

I _think_ probe-scraper is otherwise uninterested in the day it's run.

